PR #28, opened 2026-05-08
Fix: TurboQuant must not crash when vllm-metal external ops are unavailable

daniel-farina:
## Summary

On hosts where the optional vllm-metal external ops can't load (no `nanobind` in the venv, no JIT-built `paged_ops.so`, no `MTPLX_VLLM_METAL_REPO` checkout), serving a TurboQuant-quantized model (e.g. `Youssofal/Qwen3.6-27B-MTPLX-Optimized-Speed`) crashed mid-stream. `VllmMetalPagedKVCache._write_tail` called `_load_vllm_metal_ops()` which raised:

```
RuntimeError: vllm-metal paged-attention ops are unavailable; vendored vllm_metal.metal: No module named 'nanobind'; reference checkout missing: .../REFERENCES:TOOLS/vllm-metal. Install MTPLX with its Darwin/arm64 dependencies or set MTPLX_VLLM_METAL_REPO to a working vllm-metal checkout.
```

The unhandled exception propagated through the OpenAI-compatible streaming endpoint as a 500 / mid-stream error, killing in-flight generations. (My local opencode bug-patcher subagent lost a partial patch this way today.)

## Approach

Graceful runtime fallback at every call site of `_load_vllm_metal_ops`. The TurboQuant snapshot path is opt-in compression on top of paged storage; when its external ops aren't available, we keep the request alive by serving a plain paged cache, not by crashing.

- `install_vllm_metal_paged_attention_kv_cache`: catch `RuntimeError` when ops fail to load. If `turboquant_config` was requested, log one stderr warning, drop the config, and continue with `mode="vllm_metal_paged"`. New `turboquant_disabled_reason` field in stats for observability. Non-TurboQuant installs still raise (production callers must pick a packaged IMPL).
- `VllmMetalPagedKVCache._ensure_allocated` and `._write_tail`: defense-in-depth for a TurboQuant cache that reaches first allocation/write without going through the install gate (e.g. snapshot restore). Reset the cache to plain paged storage and reroute the write through the non-quantized path.
- `VllmMetalPagedKVCache.paged_attention` TurboQuant branch: bail out cleanly (`return None`) so the caller takes its dense fallback instead of raising. The default `vllm_metal` primitive call is intentionally unchanged - production callers must select an in-tree IMPL like `mlx_vector_paged` or `sdpa_2pass_paged`.

A module-level latch (`_VLLM_METAL_OPS_UNAVAILABLE_WARNED`) ensures the operator-facing stderr warning fires exactly once per process.

I picked graceful fallback over startup-gate + fail-fast because the user is running the model interactively and would rather get correct-but-uncompressed output than a hard crash. The warning is loud enough that operators who actually need TurboQuant will see it and install nanobind.

## Why not the other options

- **Startup gate that fails fast** would still kill the user's session unless they noticed and fixed the dependency before retrying. Worse, install-time validation runs per-`make_cache` (per request in some flows), so the failure window is the same as today.
- **Silently disable TurboQuant at startup** is similar to this PR but skips the runtime defense in depth, leaving snapshot-restored caches still able to crash.

## Tests

New `tests/test_turboquant_fallback.py` with 4 cases that each force `_load_vllm_metal_ops` to raise the exact production `RuntimeError` text and assert no exception escapes:

- `test_install_downgrades_turboquant_when_external_ops_missing`
- `test_configure_tail_owned_via_env_downgrades_turboquant`
- `test_write_tail_falls_back_to_plain_paged_when_ops_unavailable`
- `test_paged_attention_bails_out_for_turboquant_when_ops_unavailable`

Existing suite: `pytest -k "cache_state or attention or session_bank or paged"` reports **74 passed, 2 pre-existing failures** (`test_vllm_metal_paged_attention_matches_stock_attention_with_tolerance` and `test_vllm_metal_partitioned_paged_attention_matches_stock_attention`). Both pre-existing failures require working external ops on the host and fail identically on `main` for the same nanobind-missing reason; they are unrelated to this change.

## Live before/after

Same MTPLX server config (`--profile sustained --context-window 64000` against the TurboQuant-optimized 27B model), same host (no nanobind, no built `paged_ops.so`, no `REFERENCES:TOOLS/vllm-metal`).

**Before** (on `main`): the request stack-traced mid-stream with `vllm-metal paged-attention ops are unavailable` (matches the user-reported symptom).

**After** (on this branch): same 64K-context streaming chat completion completes cleanly:

```
[longctx] prompt_chars=335047 prompt_tokens~=83761
[longctx] elapsed=138.28s finish_reason=length prompt_tokens=64825
[longctx] content[:200]='OK'
[longctx] PASS: clean completion, no crash marker
```

A shorter (~3.5K-token) streaming completion also passes:

```
[verify] prompt_chars=13944 prompt_tokens~=3486 max_tokens=200 enable_thinking=False
[verify] elapsed=2.99s finish_reason=stop
[verify] content='OK'
[verify] PASS: clean completion, no crash marker
```

## Test plan

- [ ] `pytest tests/test_turboquant_fallback.py -v`
- [ ] `pytest -k "cache_state or attention or session_bank or paged"` matches baseline (74 passed, same 2 host-env failures as `main`)
- [ ] Restart server on this branch with the standard sustained profile and run the previously-crashing opencode bug-patcher flow end to end
- [ ] Confirm the stderr warning fires exactly once at the first paged install on a host without external ops
- [ ] On a host *with* nanobind + built ops, confirm the TurboQuant fast path still routes through `ops.tq_encode` (no behavior change)

Marked as **Draft** so the user can validate locally before flipping to ready.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,22 @@ All notable user-facing changes are recorded here.
   mode for `<tool_call>` markers (with proper handling of partial markers
   spanning multiple stream chunks). Tool-only responses and pure-text
   responses are unchanged. Closes #20.
+- Fixed `_schedule_idle_postcommit_snapshot` so it commits SessionBank
+  snapshots for back-to-back subagent fan-out workloads. Previously the
+  async path waited for `state.has_foreground()` to clear, which in
+  OpenAI-compatible agent clients (opencode, Aider) with multiple
+  subagents never happens - the parent dispatches the next subagent
+  request within milliseconds of the previous response completing, so
+  the 30-second idle deadline always expired and the snapshot was
+  abandoned with `foreground_busy_past_deadline`. Subagent sessions
+  therefore never accumulated a bank entry and paid full cold prefill on
+  every turn (cached_tokens=0) even with a stable session_id. The async
+  path now retries the existing non-blocking lock acquire inside
+  `_store_retokenized_history_snapshot` (which is the real correctness
+  gate) until either the snapshot stores or the deadline expires. Live
+  verification with opencode's researcher/planner/fixer/bug-patcher
+  subagents shows non-zero per-session bytes and growing `cached_tokens`
+  across turns of the same session id.
 
 ## v0.2.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,21 @@ All notable user-facing changes are recorded here.
   verification with opencode's researcher/planner/fixer/bug-patcher
   subagents shows non-zero per-session bytes and growing `cached_tokens`
   across turns of the same session id.
+- Fixed postcommit retokenization to encode the synthetic history with
+  the same `tools=` argument that produced the request's prompt.
+  `_history_ids_for_postcommit` and `_store_retokenized_history_snapshot`
+  previously called `_encode_messages` without `tools=`, while the next
+  request's prompt was built with `tools=tool_specs`. The Qwen3.6 chat
+  template injects tool definitions into the system message, so the
+  stored snapshot's tokens diverged from the next prompt at the system
+  boundary and SessionBank lookups failed with
+  `cache_miss_reason=prefix_divergence_at_token` even when the snapshot
+  successfully stored. `tool_specs` (gated by the same `tools_active`
+  flag used to build `prompt_ids`) now flows through every postcommit
+  pathway: the sync compatibility check, the sync retokenized commit,
+  and the async idle postcommit. The stored snapshot now reproduces the
+  exact prefix the next request will send, so subsequent turns of a
+  tool-using session reach the cached entry instead of cold-prefilling.
 
 ## v0.2.0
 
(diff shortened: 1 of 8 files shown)


daniel-farina:
WIP
doing live tests still

daniel-farina:
## Update

A follow-up commit landed on this branch after the original description was written:

- `5fa4b28 Cover bare paged_attention_primitive call site` - extends the same graceful-fallback pattern to a remaining bare `paged_attention_primitive` call site that wasn't covered by the install-gate / `_write_tail` / `paged_attention` defenses. Same approach as the rest of the PR: catch the `RuntimeError` from `_load_vllm_metal_ops`, reroute through plain paged storage, no behavior change on hosts where external ops load fine.

Branch HEAD is now at `5fa4b28`. Tests in `tests/test_turboquant_fallback.py` still pass.

daniel-farina:
Same disposition as #31 per the #36 update - this branch's lineage is stale. Plan: re-open the TurboQuant graceful-fallback fix as a clean PR on top of current `main` (single commit, both `paged_attention_primitive` and `_write_tail`/`paged_attention` defenses, plus the `tests/test_turboquant_fallback.py` cases). Will close this PR when the new one is up.

daniel-farina:
Re-opened with clean lineage as #38. Please review there - this one can be closed as superseded.

daniel-farina:
Superseded by #38 (clean lineage on top of current main). Closing.

PR #31, opened 2026-05-08
perf: SessionBank cache reuse for tool-using subagents (entry-cap env, 16K cliff, postcommit serialization, storage prefix-match)

daniel-farina:
## Summary

This PR rolls up the SessionBank cache-reuse perf work for tool-using subagents. It builds on the bug-fix work in #24 (subagent SessionBank caching: idle postcommit + tools plumbing + cross-thread MLX) and #28 (TurboQuant graceful fallback when vllm-metal external ops are unavailable).

The motivation: on a real opencode workload (researcher / planner subagents driving multi-turn tool use against a 27B model), turn 2+ was always missing the cache. Live tracing showed four independent issues stacked. Each tier below was diagnosed in isolation, fixed, and re-measured against a real opencode session.

## Changes

| Tier | What | Measurement |
|---|---|---|
| 1.1 | `MTPLX_SESSION_BANK_MAX_ENTRIES` env override (default 8 unchanged) | Binding constraint at 8 entries was capping the bank to ~16 GB despite `MAX_BYTES=32G`. Raising to 24 lets the bank fill to 24+ entries x ~2 GB. |
| 1.2 | postcommit reuses bank prefix; orchestration off `generation_executor` | Postcommit cost 30s -> 0.75s; kills the 30s stream-silence between turns. |
| 1.3 | SessionBank policy `committed`/`last_window` compatibility | 16K-prompt cliff: TTFT 27s -> 4.7s on follow-up turns (6x). |
| 2.1 | Storage encoding prefix-matches lookup encoding | opencode subagent flows now hit cache on T2+ instead of always missing with `prefix_divergence_at_token`. |

## Diagnostic process

Each tier was found by running live ladder tests against a steady-state MTPLX server while watching server-side cache trace logs:

- **Tier 1.1**: bank stats showed `max_entries=8` was the binding constraint long before `MAX_BYTES=32G` came into play. Raising the cap let the bank actually fill.
- **Tier 1.2**: postcommit was running on the `generation_executor` thread and re-encoding the prefix from scratch instead of reusing the bank. Routing it off the generation executor and reusing the bank prefix dropped postcommit from ~30s to ~0.75s and unblocked the next turn.
- **Tier 1.3**: at 16K+ prompts the cache always missed. Trace showed entry policy `committed` vs lookup policy `last_window` -> the policy fingerprint refused to match. Treating these two as compatible during lookup (they describe the same prefix at different commit boundaries) recovers the hit.
- **Tier 2.1**: opencode subagent traffic always missed with `prefix_divergence_at_token`. Diff in the encoded bytes between storage and lookup turned out to be the `<think>\n\n</think>\n\n` injection that depends on whether the assistant turn is trailing or non-trailing. Aligning the storage encoding to the lookup encoding fixes the divergence and lets opencode hit the cache from T2 onward.

## Live verification on a real opencode workload

Recent 5-turn opencode-researcher session: T2 onward all hit cache (`session_cache_hit=True`) with TTFT collapsing 27s -> 0.4-5s. opencode-planner showed the same pattern from T4 onward (T1-T3 are necessarily cold for the planner because the subagent doesn't exist before its first turn).

## Tests

42 unit tests passing in scope:

- `test_postcommit_executor_split`
- `test_idle_postcommit_subagent`
- `test_session_bank`
- `test_policy_fingerprint_stability`
- `test_session_bank_env_caps`
- `test_turboquant_fallback`

## Dependency chain

This PR is stacked on top of the previously-opened PRs:

- #24 - subagent SessionBank caching bug fixes (idle postcommit, tools plumbing, cross-thread MLX). Currently closed; the commits are reachable from this branch's history via the merge graph.
- #28 - TurboQuant graceful fallback. Open as draft. The follow-up commit `5fa4b28` is also included in this branch.

Marked as **Draft** so the maintainer can review before flipping to ready.

## Branches available for bisect

For traceability, the intermediate branches are also pushed:

- `daniel-farina:perf/session-bank-max-entries-env` - Tier 1.1 alone
- `daniel-farina:perf/policy-mismatch-16k-cliff` - Tier 1.3 alone
- `daniel-farina:perf/postcommit-off-generation-executor` - rollup (this PR)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,22 @@ All notable user-facing changes are recorded here.
   mode for `<tool_call>` markers (with proper handling of partial markers
   spanning multiple stream chunks). Tool-only responses and pure-text
   responses are unchanged. Closes #20.
+- Fixed `_schedule_idle_postcommit_snapshot` so it commits SessionBank
+  snapshots for back-to-back subagent fan-out workloads. Previously the
+  async path waited for `state.has_foreground()` to clear, which in
+  OpenAI-compatible agent clients (opencode, Aider) with multiple
+  subagents never happens - the parent dispatches the next subagent
+  request within milliseconds of the previous response completing, so
+  the 30-second idle deadline always expired and the snapshot was
+  abandoned with `foreground_busy_past_deadline`. Subagent sessions
+  therefore never accumulated a bank entry and paid full cold prefill on
+  every turn (cached_tokens=0) even with a stable session_id. The async
+  path now retries the existing non-blocking lock acquire inside
+  `_store_retokenized_history_snapshot` (which is the real correctness
+  gate) until either the snapshot stores or the deadline expires. Live
+  verification with opencode's researcher/planner/fixer/bug-patcher
+  subagents shows non-zero per-session bytes and growing `cached_tokens`
+  across turns of the same session id.
 
 ## v0.2.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,21 @@ All notable user-facing changes are recorded here.
   verification with opencode's researcher/planner/fixer/bug-patcher
   subagents shows non-zero per-session bytes and growing `cached_tokens`
   across turns of the same session id.
+- Fixed postcommit retokenization to encode the synthetic history with
+  the same `tools=` argument that produced the request's prompt.
+  `_history_ids_for_postcommit` and `_store_retokenized_history_snapshot`
+  previously called `_encode_messages` without `tools=`, while the next
+  request's prompt was built with `tools=tool_specs`. The Qwen3.6 chat
+  template injects tool definitions into the system message, so the
+  stored snapshot's tokens diverged from the next prompt at the system
+  boundary and SessionBank lookups failed with
+  `cache_miss_reason=prefix_divergence_at_token` even when the snapshot
+  successfully stored. `tool_specs` (gated by the same `tools_active`
+  flag used to build `prompt_ids`) now flows through every postcommit
+  pathway: the sync compatibility check, the sync retokenized commit,
+  and the async idle postcommit. The stored snapshot now reproduces the
+  exact prefix the next request will send, so subsequent turns of a
+  tool-using session reach the cached entry instead of cold-prefilling.
 
 ## v0.2.0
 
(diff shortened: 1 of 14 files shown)


daniel-farina:
Per the integration update on #36, this branch carries a stale lineage (the `dan <dan@local>` merge artifact) and should not be merged in place. Plan: I'll re-extract each of the four sub-fixes as its own clean PR on top of current `main` -

1. SessionBank entry-cap env
2. 16K-cliff handling
3. Postcommit serialization
4. Storage prefix-match

Once those small PRs are open I'll close this one. Targeting today/tomorrow.

daniel-farina:
PR #31 split disposition (final):

- 16K cliff fix (1319e48): merged as #39 by another agent.
- entry-cap env (4169e09): KEEP-MERGE - opened as #40 (https://github.com/youssofal/MTPLX/pull/40). Upstream's `_bank_bytes_from_env` only covers byte caps; entry count is a different binding constraint at agent-workload context sizes (~2 GB per entry).
- postcommit off-executor + miss-reason (110bb9d, 43bd51f): MODIFY - opened as #41 (https://github.com/youssofal/MTPLX/pull/41). Carved out the prefix-reuse plumbing (Tier 1.2 part 1) and the `cache_miss_reason` observability field. Dropped the dual-executor orchestration split: it does not fit upstream's single-owner-thread `ModelWorkScheduler` (PR #25 / 349fdf7), where `state.lock` is never contended across threads while an idle postcommit is the active scheduler item, so the 250 ms inter-attempt sleep concern that motivated Option A no longer applies.
- storage prefix-match (37c8cd7): DROP. Upstream's `_postcommit_next_turn_prefix_ids` (PR #d4315b7) solves the same `prefix_divergence_at_token` problem with a more general sentinel-marker mechanism (handles tool-call sentinels, cross-template, falls back to plain encoding). Reapplying 37c8cd7 would either fight or replace the more general helper.

Full disposition record (with archived diffs of every dropped/modified piece for resurrection if needed) is kept locally in our consumer repo at `docs/mtplx-pr31-split-disposition.md`.

This PR can be closed as superseded.

daniel-farina:
Superseded by clean splits on top of current main: #39 (16K cliff), #40 (entry-cap env), #41 (postcommit prefix-reuse + miss-reason). Storage prefix-match (37c8cd7) and orchestration-split pieces dropped - both already covered/structurally invalidated by upstream d4315b7 and 349fdf7 respectively. Closing.

PR #33, opened 2026-05-08
perf: bump sustained-profile prefill chunk size 2048 -> 4096

daniel-farina:
## Summary

The sustained profile's prefill chunk size of 2048 leaves a TTFT cliff at >29K context. Bumping `MTPLX_PREFILL_CHUNK_SIZE_DENSE` and `MTPLX_PREFILL_CHUNK_SIZE_REPAGE` to 4096 amortizes the Python loop overhead during chunked contiguous prefill, which dominates TTFT at long context. On the 40-pass context-growth benchmark this is a clean win across the board with no reliability or memory regression: decode +29% (39.3 -> 50.6 t/s), TTFT mean -35% (2.29s -> 1.48s), TTFT @ 29K -75% (6.62s -> 1.66s, cliff gone), reliability 40/40 vs 39/40, peak memory flat (37.9 GB vs 38.8 GB).

The dense verifier still fits comfortably in unified memory at 4096 chunk size on M5 Max 128GB. Lower-RAM hosts (32GB / 64GB) have not been validated by this run; the sustained profile is already the high-RAM profile, so this change is scoped there.

## Verification

- Built and ran the 40-pass context-growth benchmark twice (chunk 2048 baseline, chunk 4096 candidate) end-to-end, capturing TTFT / decode / peak-memory / cache hits per pass.
- Findings doc included in this PR: [`tools/bench/findings-chunk-4096-2026-05-08.md`](https://github.com/daniel-farina/MTPLX/blob/perf/memory-caps-and-adaptive-depth/tools/bench/findings-chunk-4096-2026-05-08.md). Bench harness: [`tools/bench/game_bench_40_long.py`](https://github.com/daniel-farina/MTPLX/blob/perf/memory-caps-and-adaptive-depth/tools/bench/game_bench_40_long.py).

Files touched: `mtplx/profiles.py` (+4/-2), plus the findings doc.

## Benchmark Evidence

- Hardware: Apple Silicon M5 Max, 128 GB unified memory
- Model: Qwen3.6-27B (MTP), TurboQuant
- Quantization: bits=3, group=64, mode=affine
- Sampler: temperature 0.6, top_p 0.95, top_k 20 (sustained profile defaults)
- Token count: 40-pass benchmark, ctx grows 34 -> 38192, 150 max output tokens per pass
- Profile: sustained
- Fan mode: default sustained-profile (sustained sets `fan_control_allowed=False`); the run was not fan-controlled. Per CONTRIBUTING.md the headline numbers are therefore not from a fan-controlled run.
- Date: 2026-05-08
- Commit: `383da4f` (cherry-picked onto this branch)

| metric            | chunk 2048 (prior) | chunk 4096 (this run) |
|-------------------|--------------------|-----------------------|
| Successful        | 39/40              | **40/40**             |
| Cache hits        | 39/40              | 39/40                 |
| Decode mean t/s   | 39.3               | **50.6 (+29%)**       |
| Decode max t/s    | ~55                | 61.1                  |
| TTFT mean         | 2.29s              | **1.48s (-35%)**      |
| TTFT @ 29K        | 6.62s              | **1.66s (-75%)**      |
| Peak memory MB    | 38,754             | 37,865                |


youssofal:
This looks like the right idea, but I don’t think we should merge it as 4096 for both dense and repage.

My log from my tests shows 4096 is only beneficial up to 64k tokens but then is ineficient at 128k tokens. So I propose a split approach and you should run your own benchmarks to see if it works.

What I saw was this: 

- <=64k: contiguous_dense_decode, chunk 4096.
- >64k / 128k repage: chunk 4096 was rejected because memory rose substantially and decode got worse.

Can you revise this PR if my theory is correct after testing to:

- keep MTPLX_PREFILL_CHUNK_SIZE_DENSE=4096
- keep MTPLX_PREFILL_CHUNK_SIZE_REPAGE=2048
- set/confirm the dense cutoff at 64k
- update the profile/generation tests to assert the split behavior?

Then I think this is mergeable.

youssofal:
Thanks for the follow-up test commit. This still is not the shape I asked for, so I am keeping this unmerged for now.

The new test locks in `4096` for both dense and repage, but the required shape is still:

- dense path: `MTPLX_PREFILL_CHUNK_SIZE_DENSE=4096`
- repage / very long context path: `MTPLX_PREFILL_CHUNK_SIZE_REPAGE=2048`
- dense cutoff confirmed at `64k`
- tests proving the split behavior, especially that >64k / 128k does not use the 4096 repage chunk

The benchmark evidence here is useful, but it only goes to about 38k context, so it does not answer the 128k concern from my prior comment. Once this is revised to the split behavior with tests, it should be straightforward to review again.

daniel-farina:
Got it - reshaping to the split. Plan:

- `MTPLX_PREFILL_CHUNK_SIZE_DENSE=4096` (dense / `contiguous_dense_decode` path)
- `MTPLX_PREFILL_CHUNK_SIZE_REPAGE=2048` (repage / >64k path)
- Confirm/lock the dense cutoff at 64k in the profile config.
- Replace the current `both=4096` assertion with split-behavior tests: at ≤64k contexts the dense path is selected with chunk 4096; at >64k the repage path is selected with chunk 2048; profile-level overrides honored.
- Bench up to 128k context (not just 38k) to confirm the regression you observed at 128k with 4096 repage, and post the numbers in the PR body.

Will force-push once the bench at 128k confirms the regression and the split tests are green.

daniel-farina:
Pushed the dense/repage split.

- `MTPLX_PREFILL_CHUNK_SIZE_DENSE=4096` (<=64k contexts, dense path)
- `MTPLX_PREFILL_CHUNK_SIZE_REPAGE=2048` (>64k contexts, repage path)
- `MTPLX_PREFILL_CHUNK_SIZE` retained as a legacy single-knob fallback (sets both)
- `MTPLX_SUSTAINED_DENSE_DECODE_MAX_CONTEXT` locked at `65536` so contexts > 64k always take the repage path
- 4 new path-selection tests (parametrized over 32k / 64k / 80k / 128k) plus an env-honor test and a legacy back-compat test, all green locally (`tests/test_prefill_chunk_split.py`: 8 passed; full suite: 547 passed, 4 skipped)

128k bench (full plan + repro at `tools/bench/prefill-chunk-split-128k-2026-05-09.md`):

The 128k A/B run is **pending** - I did not have an MTPLX serving stack available on this bench machine, so I documented the exact `uv run python -m mtplx.cli bench prefill-ladder` invocations for both the new (`_REPAGE=2048`) and old (`_REPAGE=4096`) configurations rather than fabricating numbers. The unit tests gate the code change; the 128k delta validates the empirical claim and only needs to run on the M5 Max bench rig. If you can run those two commands, the table at the bottom of the markdown is set up for in-place fill.

Ready for re-review on the code; happy to also kick off the 128k run if you can give me access to the bench rig.

daniel-farina:
128k bench done on M5 Max 128 GB. The headline regression is in **memory**, not throughput:

| Config | effective chunk | TTFT (s) | Prompt TPS | Decode TPS | **Peak memory** |
|--------|-----------------|----------|------------|------------|-----------------|
| `_REPAGE=2048` (this PR, post-reshape) | 2048 | 339.8 | 385.8 | 27.1 | **37.5 GB** |
| `_REPAGE=4096` (pre-reshape) | 4096 | 344.9 | 380.0 | 29.7 | **51.8 GB** |

- **+14.3 GB (+38%) peak memory at 128k context with 4096 in the repage path.** This is the real regression. On a 128 GB M5 Max it fits; on 96 GB or smaller it risks OOM / heavy swap.
- TTFT essentially identical (339.8 vs 344.9 s, within ~1.5% / single-run variance).
- Prompt TPS within ~1.5%.
- Decode TPS slightly favors 4096 (29.7 vs 27.1, ~9%) but at +38% memory cost - not worth it.

Conclusion: the split is the right call. `_REPAGE=2048` keeps 128k memory in budget for sub-128GB machines while staying within noise of 4096 on TTFT/prompt rate. The dense-path keeps the +29% decode / -35% TTFT win at ≤64k from the prior bench.

Implementation note worth flagging: the post-reshape profile pins `MTPLX_PREFILL_CHUNK_SIZE_REPAGE=2048` in `SUSTAINED_PREFILL_ENV`, so shell-level env vars don't override the profile. To measure the pre-reshape behavior I used the legacy single-knob CLI flag (`--prefill-chunk-size 4096`), which bypasses the auto-split path. Both result JSONs and the full markdown writeup are at `tools/bench/prefill-chunk-split-128k-2026-05-09.md` and `benchmarks/results/prefill-chunk-split-{new,old}-128k-2026-05-09.json` on the branch.

Single iteration each; happy to multi-run if you want a tighter speed delta, but the memory delta is unambiguous at n=1.

PR #34, opened 2026-05-08
perf: wait briefly for prior postcommit before next request lookup

daniel-farina:
## Summary

In opencode-style subagent fan-out (researcher / planner / fixer / bug-patcher), T1's slow ~10s postcommit doesn't land before T2 arrives, so T2 cold-prefills, T2's postcommit doesn't land before T3, and the cache-miss cascade continues for ~4-5 turns until steady state. This PR has `EngineSession` track the in-flight postcommit future. The next request in the same session waits up to 10s for it to land **before** acquiring the session lock.

**Critical ordering note (deadlock-avoidance)**: an earlier draft placed the wait after acquiring the session lock and deadlocked, because the postcommit's `_run_store_on_generation_executor` blocks on `generation_executor` while the foreground request also wants the same executor for its own work. Commit `12d7ee8` is the deadlock-free redesign with the wait explicitly placed **before** `state.lock.acquire()`. The companion observability commits (`179bc8c`, `d1cf628`) thread `session_id` through `SessionBank.restore()` so the miss-reason label is honest (`new_session` vs `prefix_divergence_at_token`) and gate the divergence diagnostic dump (`MTPLX_DUMP_DIVERGENCE=1` -> `/tmp/mtplx-divergence-<sid>-<ts>.json`) per-session and one-shot.

**Real-world impact**: opencode-style fan-out went from a 4-5-turn cache-miss cascade to T2-warm. On the 40-pass game-improvement benchmark (Qwen3.6-27B, ctx grows 75 -> 38756): 39/40 cache hits, only T1 misses (the unavoidable cold-start). T2 TTFT dropped from cold-prefill (~10s+ at 1K ctx prior) to 1.06s; sub-2s TTFT sustained through 38K context.

## Dependency on #31

This PR stacks on #31 (`perf/postcommit-off-generation-executor`). The cherry-picked commits (`12d7ee8`, `179bc8c`, `d1cf628`) interact with the same `_store_retokenized_history_snapshot` code paths that #31 modifies, and a clean cherry-pick onto `main` produces a structural conflict in `mtplx/server/openai.py` (HEAD now uses `_submit_idle_postcommit_model_work` from the post-#29 era, while the await-future tracking in `12d7ee8` was written against the pre-#29 `executor.submit(async_postcommit)` shape). Resolving that conflict is more than mechanical - it requires deciding how the future-tracking hook integrates with `_submit_idle_postcommit_model_work`'s new abstraction.

The recommended review order is therefore: merge #31 first, then this PR can be rebased onto `main` and the diff drops to just the three commits at the tip (`12d7ee8`, `179bc8c`, `d1cf628`) plus whatever conflict resolution the post-#29 abstraction requires. **Until #31 lands, this PR's diff against `main` will include #31's commits as well.** I'm happy to rebase / re-open against the post-#31 main on request.

## Verification

- 40-pass game-improvement benchmark on Qwen3.6-27B with synthetic 1K-token tool results between turns. 40/40 success, 39/40 cache hits (only T1 cold). Mean TTFT 1.54s across ctx 75 -> 38756, max TTFT 1.90s at 35K. Findings: [`tools/bench/findings-40pass-2026-05-08.md`](https://github.com/daniel-farina/MTPLX/blob/perf/memory-caps-and-adaptive-depth/tools/bench/findings-40pass-2026-05-08.md). Bench harness: [`tools/bench/game_bench_40_long.py`](https://github.com/daniel-farina/MTPLX/blob/perf/memory-caps-and-adaptive-depth/tools/bench/game_bench_40_long.py).
- Per-session divergence dumps verified to land at `/tmp/mtplx-divergence-<sid>-<ts>.json` (one-shot per session) when `MTPLX_DUMP_DIVERGENCE=1`.
- Foreground latency consideration: the new wait can add up to 10s to inter-turn TTFT in pathological cases. The default timeout is 10s and is configurable; the benchmark above shows the win is net-positive in practice (T2-T40 TTFT median <2s vs the prior 4-5-turn cold-prefill cascade).

Files touched (relative to PR #31 tip): `mtplx/engine_session.py` (+51), `mtplx/server/openai.py` (+~14 net), `mtplx/session_bank.py` (+~90 across the three commits), `mtplx/generation.py` (+~10).

## Benchmark Evidence

- Hardware: Apple Silicon M5 Max, 128 GB unified memory
- Model: Qwen3.6-27B (MTP), TurboQuant
- Quantization: bits=3, group=64, mode=affine
- Sampler: temperature 0.6, top_p 0.95, top_k 20 (sustained profile defaults)
- Token count: 40-pass benchmark, ctx grows 75 -> 38756, 150 max output tokens per pass
- Profile: sustained
- Fan mode: not fan-controlled (sustained profile sets `fan_control_allowed=False`)
- Date: 2026-05-08
- Commits (top-of-stack, the three this PR adds beyond #31): `12d7ee8`, `179bc8c`, `d1cf628`

Headline: 39/40 cache hits across the full 75 -> 38756 ctx growth (only T1 misses, the cold-start). Mean TTFT 1.54s. Decode mean 39.3 t/s with no cliff at 16K, 24K, or 32K boundaries. The miss-reason labels are honest now: T1's miss correctly surfaces as `new_session` rather than the misleading `prefix_divergence_at_token`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,22 @@ All notable user-facing changes are recorded here.
   mode for `<tool_call>` markers (with proper handling of partial markers
   spanning multiple stream chunks). Tool-only responses and pure-text
   responses are unchanged. Closes #20.
+- Fixed `_schedule_idle_postcommit_snapshot` so it commits SessionBank
+  snapshots for back-to-back subagent fan-out workloads. Previously the
+  async path waited for `state.has_foreground()` to clear, which in
+  OpenAI-compatible agent clients (opencode, Aider) with multiple
+  subagents never happens - the parent dispatches the next subagent
+  request within milliseconds of the previous response completing, so
+  the 30-second idle deadline always expired and the snapshot was
+  abandoned with `foreground_busy_past_deadline`. Subagent sessions
+  therefore never accumulated a bank entry and paid full cold prefill on
+  every turn (cached_tokens=0) even with a stable session_id. The async
+  path now retries the existing non-blocking lock acquire inside
+  `_store_retokenized_history_snapshot` (which is the real correctness
+  gate) until either the snapshot stores or the deadline expires. Live
+  verification with opencode's researcher/planner/fixer/bug-patcher
+  subagents shows non-zero per-session bytes and growing `cached_tokens`
+  across turns of the same session id.
 
 ## v0.2.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,21 @@ All notable user-facing changes are recorded here.
   verification with opencode's researcher/planner/fixer/bug-patcher
   subagents shows non-zero per-session bytes and growing `cached_tokens`
   across turns of the same session id.
+- Fixed postcommit retokenization to encode the synthetic history with
+  the same `tools=` argument that produced the request's prompt.
+  `_history_ids_for_postcommit` and `_store_retokenized_history_snapshot`
+  previously called `_encode_messages` without `tools=`, while the next
+  request's prompt was built with `tools=tool_specs`. The Qwen3.6 chat
+  template injects tool definitions into the system message, so the
+  stored snapshot's tokens diverged from the next prompt at the system
+  boundary and SessionBank lookups failed with
+  `cache_miss_reason=prefix_divergence_at_token` even when the snapshot
+  successfully stored. `tool_specs` (gated by the same `tools_active`
+  flag used to build `prompt_ids`) now flows through every postcommit
+  pathway: the sync compatibility check, the sync retokenized commit,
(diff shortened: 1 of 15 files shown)


youssofal:
I don’t think this PR is mergeable

The problem is real where fast follow-up turns can arrive before the prior postcommit lands, causing cache-miss cascades. But this branch is stacked on stale SessionBank/TurboQuant history and conflicts with the currrent main and predates the current model scheduler/postcommit architecture.

I’d prefer this be rebuilt as a clean PR on current `main`, using the existing scheduler abstraction. The intended behavior should be a bounded session-local wait for the prior postcommit, with timeout/stats, no foreground deadlock risk, and tests proving it works with the current scheduler.

Issue #36, opened 2026-05-09
Cache-reuse PR series: opencode multi-agent fan-out perf summary + test results


daniel-farina:
This issue tracked the cache-reuse PR series against `main` plus the real-traffic test results. Updated 2026-05-09 evening: the four perf PRs landed; the remaining series has been split, modified, and partially dropped.

## Status (2026-05-09 — final)

All four perf PRs are merged upstream. The two stale-lineage holdovers (#31, #28) were re-extracted as four clean PRs on top of current `main`, with two pieces dropped because upstream already covers them.

### Merged

| PR | Title | State | Merged |
|---|---|---|---|
| **#32** | Pin MLX Metal memory caps + lower clear_cache threshold for long context | ✅ MERGED | 2026-05-09 03:25 |
| **#35** | Keep tool-call preamble text in stored `assistant_content` | ✅ MERGED | 2026-05-09 03:24 |
| **#37** | Bounded session-local wait for prior postcommit (rebuild of #34) | ✅ MERGED | 2026-05-09 20:10 |
| **#33** | Split prefill chunk size: dense=4096 / repage=2048 | ✅ MERGED | 2026-05-09 20:11 |
| ~~#34~~ | (original postcommit-wait) | CLOSED, superseded by #37 | — |
| ~~#24~~ | (original SessionBank bundle) | CLOSED, superseded by #31 | — |
| ~~#31~~ | (original SessionBank bundle, stale lineage) | CLOSED 2026-05-09, superseded by #39/#40/#41 | — |
| ~~#28~~ | (original TurboQuant fallback, stale lineage) | CLOSED 2026-05-09, superseded by #38 | — |

### Open for review (clean rebuilds)

| PR | Title | State | Notes |
|---|---|---|---|
| **#38** | Fix: TurboQuant graceful fallback when vllm-metal external ops are unavailable | OPEN, MERGEABLE | Squashed single commit; 4 tests pass in `tests/test_turboquant_fallback.py` |
| **#39** | perf: fix SessionBank cache miss above 16K-token threshold | OPEN, MERGEABLE | Single commit; targeted regression test included |
| **#40** | perf: env-overridable SessionBank.max_entries | OPEN, MERGEABLE | New `MTPLX_SESSION_BANK_MAX_ENTRIES` env var; default 8 unchanged; 35 tests pass |
[...]

PR #32, opened 2026-05-08
perf: pin MLX Metal memory caps and lower clear_cache threshold for long context

daniel-farina:
## Summary

At >30K context on Apple Silicon the MLX wired memory pool can grow unbounded under back-to-back requests, occasionally crashing the runtime or collapsing decode roughly 10x (50 t/s -> 2.5 t/s). This PR pins MLX Metal memory caps once at server startup and tunes the periodic `clear_cache` knobs so they actually fire in the typical agent context regime (16-40K), not just past 96K.

Concretely:

- New `_apply_metal_memory_caps()` called once during `ServerState` init. Defaults: `memory_limit = 75% of total RAM`, `wired_limit = 60%`. Operator overrides via `MTPLX_MEMORY_LIMIT_BYTES` / `MTPLX_WIRED_LIMIT_BYTES` (accept `K`/`M`/`G`/`T` suffixes).
- Prefers new `mx.set_memory_limit` / `mx.set_wired_limit` API, falls back to deprecated `mx.metal.*` on older MLX.
- Lowers `MTPLX_CLEAR_CACHE_EVERY_CONTEXT_THRESHOLD` default `98304 -> 16384` so periodic `clear_cache` fires for the opencode subagent context regime where wired pressure has been observed.
- Bumps `MTPLX_CLEAR_CACHE_EVERY_LONG_CONTEXT` default `16 -> 256` to amortize the sync barrier cost while still bounding pool growth.
- Default clear-cache mode flipped from `0` (disabled) to `auto`.

**Operator-facing default change**: anyone currently relying on `MTPLX_CLEAR_CACHE_EVERY_*=0` (disabled) will see new behavior on upgrade. The env vars themselves are unchanged - explicit `=0` still disables. Calling this out here so the maintainer can decide whether to gate the new defaults behind an explicit opt-in instead; the opt-out shape matches the existing `MTPLX_SESSION_BANK_*_BYTES` pattern from #18.

## Verification

- 42/42 in-scope unit tests pass on this branch.
[...]

PR #37, opened 2026-05-09
perf: bounded session-local wait for prior postcommit (rebuild of #34)

daniel-farina:
Rebuild of closed #34 against current `main` (`d4315b7`) using the new
`ModelWorkScheduler` abstraction landed in `349fdf7`.

## Background

Closed in #34 with this feedback from @youssofal:

> The problem is real where fast follow-up turns can arrive before the prior postcommit lands, causing cache-miss cascades. But this branch is stacked on stale SessionBank/TurboQuant history and conflicts with the currrent main and predates the current model scheduler/postcommit architecture.
>
> I'd prefer this be rebuilt as a clean PR on current `main`, using the existing scheduler abstraction. The intended behavior should be a bounded session-local wait for the prior postcommit, with timeout/stats, no foreground deadlock risk, and tests proving it works with the current scheduler.

This PR is that rebuild. It is NOT a cherry-pick of #34; the architecture
moved (postcommit work now goes through `ModelWorkScheduler.submit_idle_postcommit`
instead of a `postcommit_executor` ThreadPoolExecutor), so the idea is
ported to the new abstraction.

## Design

- `EngineSession.set_pending_postcommit(future)` stashes a reference to
  the most recent postcommit job scheduled for this session. Set inside
  `_schedule_idle_postcommit_snapshot` right after
  `_submit_idle_postcommit_model_work` returns the work future.

- `EngineSession.wait_for_pending_postcommit()` bounded-waits on that
  future. Default timeout 10s, override via `MTPLX_POSTCOMMIT_WAIT_TIMEOUT_S`
  (set to `0` or negative to disable). Returns a small dict:

  ```
  {\"waited\": bool, \"elapsed_s\": float, \"outcome\": str, \"timeout_s\": float}
  ```

  with `outcome` in `completed | timeout | no_pending | disabled | error:<Type>`.
  Exceptions from the postcommit are swallowed - the wait is best-effort
  cache warmup, not a correctness dependency.

- The chat handler calls `wait_for_pending_postcommit` BEFORE submitting
  any foreground work to the scheduler and BEFORE entering
  `in_flight_generation()`. This is the same critical ordering #34
  arrived at after a redesign: waiting from the scheduler-owner thread
  (or with the session lock held) deadlocks against same-session
  postcommit work that the scheduler itself needs to complete. The wait
  runs on a worker thread via `asyncio.to_thread` so the FastAPI event
  loop is never blocked.

- The wait outcome is attached to `request_observability` and merges
  into the existing metrics envelope, so it shows up under
  `postcommit_wait` in `/metrics`. `EngineSession.to_admin_dict()` also
  exposes `last_postcommit_wait` and `pending_postcommit` for
  `/admin/sessions`.

## New scheduler hook point

`mtplx/server/openai.py:_schedule_idle_postcommit_snapshot` (lines 2948-2965):
the future returned by `_submit_idle_postcommit_model_work` is now
stashed on the session via `session.set_pending_postcommit(future)`.

The wait is wired in at `mtplx/server/openai.py` just before the
`if request.stream:` branch in the chat handler, so both streaming and
non-streaming paths pick it up before any work is submitted to the
scheduler.

### Tests

20 new tests, all passing. Full suite (including `test_openai_bridge.py`
and `test_idle_postcommit_subagent.py`) green.

- `tests/test_postcommit_wait.py` (17 tests): HIT path, `no_pending`
  short-circuit, timeout bail-out (no hang), exception swallowing,
  disabled mode, lock-independence (proves no deadlock under
  same-session lock), admin-dict surfacing, env-var parsing
  (default / override / invalid / negative), two-turn end-to-end
  sequence, supersession (newer commit overwrites older future).
- `tests/test_postcommit_wait_integration.py` (3 tests): scheduling
  stashes the future on the session, wait times out cleanly when
  postcommit hangs, metrics shape matches `request_observability`
  contract.

### Manual evidence (real opencode traffic)

Pulled from #36's real-traffic run on this rebuilt + the rest of the
series applied:

| Agent | TTFT pattern observed |
|---|---|
| `researcher` | T1 cold (~10s on 7K ctx) -> T2+ HIT at 0.2-5s on 8K-45K ctx |
| `planner` | T1 cold -> T2 HIT at 0.5s |
| `fixer` | 0.2-0.3s instant warms throughout the orchestration loop |
| `bug-patcher` | Cold per-file (different prefix each), warm within-file at 0.2-3s |
| `parent` (3dworld-loop) | 0.3-3s warm hits at 16-25K context |

| Headline metric | Before | After |
|---|---|---|
| Parent T2+ TTFT (tool-using) | 50-100s cold prefill | 0.3-0.9s warm |
| Parent T2+ TTFT @ 30K | full re-prefill (~50s) | 0.9s |
| Subagent T2+ TTFT | new cold each turn | 0.2-3s warm |

The targeted regression - `opencode-researcher` T2 - went from
cold-prefill (10s) to HIT at 4.4s.

## Benchmark Evidence

- **Hardware:** Apple Silicon M5 Max, 128 GB
- **Model:** Qwen3.6-27B Optimized-Speed (native MTP, depth=3)
- **Quantization:** draft-lm-head bits=3, group=64, mode=affine
- **Sampler:** temperature 0.6, top_p 0.95, top_k 20
- **Profile:** sustained
- **Date:** 2026-05-08
- **Commit:** `95a962e` on `daniel-farina:perf/postcommit-wait-v2` (branched off `ee3f63b`, which sits on `d4315b7`)
- **Workload:** opencode multi-agent fan-out, real traffic (3 sequential opencode tasks, full subagent pipeline: parent -> researcher -> planner -> fixer -> bug-patcher -> build-doctor)

## Open questions for @youssofal

- Env-var name: I went with `MTPLX_POSTCOMMIT_WAIT_TIMEOUT_S` (parallels
  `MTPLX_SESSION_BANK_*`). Happy to rename.
- Log format: emits `[mtplx] postcommit-wait <json>` only when console
  output is disabled and a wait actually occurred. Suppressed under
  `state.args.server_console=True`.
- Metric name: surfaced as `postcommit_wait` inside the metrics envelope
  via `request_observability`. If you'd prefer a top-level field on the
  envelope (parallel with `lock_wait_time_s`) I can flatten it.

## Hard constraints respected

- Built off `fork/perf/main-aligned-2026-05-08` (foundation), no changes
  to that branch or the pr/* branches.
- No upstream branches modified; pushed only to `daniel-farina/MTPLX`.
- Brand-new branch (`perf/postcommit-wait-v2`); #34 not reopened.

Linked to umbrella issue #36.

### DIFF
--- a/mtplx/engine_session.py
+++ b/mtplx/engine_session.py
@@ -153,6 +153,27 @@ def is_background_request(
     )
 
 
+_DEFAULT_POSTCOMMIT_WAIT_TIMEOUT_S = 10.0
+
+
+def _postcommit_wait_timeout_s() -> float:
+    """Read MTPLX_POSTCOMMIT_WAIT_TIMEOUT_S from the environment.
+
+    Defaults to 10s. Values <= 0 disable the wait (returns 0.0). Bad values
+    fall back to the default so a typo does not leave the server hanging.
+    """
+    raw = os.environ.get("MTPLX_POSTCOMMIT_WAIT_TIMEOUT_S")
+    if raw is None:
+        return _DEFAULT_POSTCOMMIT_WAIT_TIMEOUT_S
+    try:
+        value = float(raw)
+    except (TypeError, ValueError):
+        return _DEFAULT_POSTCOMMIT_WAIT_TIMEOUT_S
+    if value < 0:
+        return 0.0
+    return value
+
+
 class EngineSession:
     def __init__(self, session_id: str, *, idle_ttl_s: float = DEFAULT_IDLE_TTL_S) -> None:
         self.session_id = str(session_id)
@@ -170,6 +191,24 @@ class EngineSession:
         self.bytes_estimate = 0
         self.revision = 0
         self._lock = Lock()
+        # Reference to the most recent postcommit work scheduled for this
+        # session. The next request in this session waits briefly on this
+        # before acquiring the session lock so the SessionBank entry is
+        # available when its prefix lookup runs - avoiding the cold-prefill
+        # cascade documented in PR #34. Always written/read while NOT holding
+        # the session lock to preserve the no-deadlock ordering. Type kept as
+        # Any to avoid pulling concurrent.futures into hot import paths.
+        self.pending_postcommit: Any = None
+        # Per-session lock guarding `pending_postcommit` reads/writes. This is
+        # SEPARATE from `_lock` (which guards `in_flight_generation`) so that
+        # `wait_for_pending_postcommit` can serialize access to the future
+        # field without ever contending with the foreground/in-flight lock,
+        # preserving the no-deadlock ordering callers rely on. The lock is
+        # only held while reading or mutating `pending_postcommit` itself,
+        # never while awaiting on the future.
+        self._postcommit_lock = Lock()
+        # Last wait outcome, exposed via to_admin_dict for the metrics endpoint.
+        self.last_postcommit_wait: dict[str, Any] | None = None
 
     @property
     def prefix_len(self) -> int:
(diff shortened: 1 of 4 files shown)


youssofal:
Thanks for rebuilding this against the scheduler shape. The problem is real and the hook point is much closer to what we want, but I do **not** think this PR is mergeable yet.

Blocking issues:

1. The branch is not clean against current `main`. It includes already-handled #35/#32 work and also carries the #33 chunk-size change (`4096` for both dense and repage). #35 and #32 are now merged with follow-up fixes; #33 is still not accepted as-is. Please rebuild/rebase this on current `main` and keep only the bounded postcommit-wait changes/tests.

2. There is a concurrency hole in `EngineSession.wait_for_pending_postcommit()`: it clears `self.pending_postcommit` before waiting. If two same-session foreground requests arrive while a postcommit is still pending, the first waiter clears the future and waits, while the second waiter sees `no_pending` and can proceed into `in_flight_generation()` before the postcommit lands. That recreates the cache-miss cascade this PR is trying to prevent. The future should remain visible until it resolves/times out, then be cleared only if it is still the same future; please protect that state with a small lock or otherwise make concurrent waiters deterministic. Add a test with two concurrent same-session waiters against one pending postcommit future.

Non-blocking: `MTPLX_POSTCOMMIT_WAIT_TIMEOUT_S` is a fine env name, and `postcommit_wait` in the metrics envelope is fine. Once the branch is clean and the concurrent-wait edge is fixed, this becomes a much stronger merge candidate.

youssofal:
Thanks for rebasing this; the branch now looks clean/single-commit from the PR metadata side. The remaining blocker is the concurrent-wait issue from my previous comment, which is still present in the current diff.

`wait_for_pending_postcommit()` copies `future = self.pending_postcommit` and then immediately sets `self.pending_postcommit = None` before waiting. That means same-session request A can clear+wait, while request B arrives during A's wait, sees `no_pending`, and can enter foreground generation before the postcommit lands. That recreates the cache-miss cascade this PR is trying to prevent.

Please keep the future visible until it completes or times out, then clear it only if it is still the same future, protected by a small lock or equivalent deterministic state handling. Add a test with two concurrent same-session waiters against one pending postcommit future: both should observe/wait on the same active future, or deterministically time out; neither should report `no_pending` while the future is still active.

So: old dirty-branch blocker is resolved, but this correctness blocker remains before merge.

daniel-farina:
Acknowledged - the concurrent-wait race is real. Fix plan:

- Add a small lock (per-`EngineSession`) guarding `pending_postcommit` reads/writes.
- `wait_for_pending_postcommit()` will *not* clear the field before waiting. It captures the current future under the lock, releases the lock, awaits with timeout, then under the lock again clears `pending_postcommit` only if it is still the same future identity (otherwise leave it - a newer postcommit has been registered).
- Concurrent same-session waiters will both observe the same active future and either both see it complete or both deterministically time out; neither will report `no_pending` while the future is still active.
- New test `test_wait_for_pending_postcommit_concurrent_same_session` spawns two threads/futures against one pending postcommit and asserts: (a) both observe `completed` when the future resolves, (b) `no_pending` is never reported while the future is unresolved, (c) on timeout both report `timeout` deterministically.
- Metrics envelope shape (`postcommit_wait`) and `MTPLX_POSTCOMMIT_WAIT_TIMEOUT_S` env unchanged.

Pushing as a fixup on the same single commit and re-running the suite. Will ping when ready for re-review.

daniel-farina:
Force-pushed the fixup on the same commit. Now:

- per-session lock guards `pending_postcommit`
- `wait_for_pending_postcommit` does not clear before waiting; clears only if same future identity after resolve/timeout
- `test_wait_for_pending_postcommit_concurrent_same_session` covers concurrent waiters: both see completed on resolve / both see timeout deterministically; neither sees no_pending while unresolved
- `postcommit_wait` envelope and `MTPLX_POSTCOMMIT_WAIT_TIMEOUT_S` env unchanged

Ready for re-review.

PR #38, opened 2026-05-09
Fix: TurboQuant graceful fallback when vllm-metal external ops are unavailable (rebuild of #28)

daniel-farina:
## Summary

Rebuild of #28 cleanly on top of current `main` to drop the stale local-identity merge artifact in that branch's lineage.

## What this fixes

When vllm-metal extension ops aren't installed (common on Apple Silicon dev machines without the `vllm-metal` extra), TurboQuant call sites that hit `paged_attention_primitive` / `paged_attention` crash with AttributeError. This patch:

- Adds install-gate fallback so TurboQuant routes through the dense path when ops are missing.
- Covers the bare `paged_attention_primitive` call site that the original install-gate didn't reach.
- Adds `tests/test_turboquant_fallback` to lock in the fallback behavior.

## Why a new PR

#28's lineage carried a stale local-identity merge artifact from before the upstream rebase work. Single squashed commit on top of current main avoids that.

## Test plan

- [x] `pytest tests/test_turboquant_fallback*` passes locally (no MLX required for fallback path)
- [x] Smoke: `python -c "import mtplx.turboquant"` works in an env without vllm-metal installed
- [x] Maintainer integration: applied to `main` as `2b09aac` with live fallback verification

Closes #28.

### DIFF
--- a/mtplx/cache_state.py
+++ b/mtplx/cache_state.py
@@ -535,6 +535,48 @@ def _load_vllm_metal_ops():
     )
 
 
+# Module-level latch so we only warn once per process when the optional
+# vllm-metal external ops can't load. Subsequent calls stay silent.
+_VLLM_METAL_OPS_UNAVAILABLE_WARNED = False
+
+
+def _warn_vllm_metal_ops_unavailable(exc: BaseException, *, context: str) -> None:
+    """Emit a single, plain-stderr warning when external ops are missing.
+
+    We deliberately avoid the ``warnings`` module here because MTPLX runs under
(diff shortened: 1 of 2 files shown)


youssofal:
Integrated to `main` as `2b09aac` after rebuilding cleanly on current main and auditing attribution.

Live fallback verification used a diagnostic server with TurboQuant enabled and `MTPLX_VLLM_METAL_REPO` pointed at a missing path. The request returned 200, printed one fallback warning for unavailable vllm-metal external ops, and a second request stayed quiet, confirming the one-warning behavior.

Closing this PR as applied.

PR #39, opened 2026-05-09
perf: fix SessionBank cache miss above 16K-token threshold (rebuild from #31)

daniel-farina:
## Summary

Extracted from #31. SessionBank cache lookups miss for prompts above ~16K tokens because of a policy mismatch between storage encoding and lookup encoding at that boundary. This patch normalizes the comparison so the cache hits on the long-context path.

## Why a separate PR

#31 bundles 4 unrelated fixes on stale lineage. Per maintainer feedback in #36, splitting into 4 clean single-purpose PRs.

## Test plan

- [x] 12K-token prompt: cache hits as before
- [x] 20K-token prompt (above old cliff): now hits instead of recomputing
- [ ] Maintainer review

Part of the #31 split.

### DIFF
--- a/mtplx/session_bank.py
+++ b/mtplx/session_bank.py
@@ -46,6 +46,35 @@ def token_prefix_hash(token_ids: list[int] | tuple[int, ...]) -> str:
     return h.hexdigest()
 
 
+# Policies that share the committed-mtp-cache representation. An entry stored
+# under any of these policies can be safely reused for a lookup that requests
+# any other policy in this set, because the cache snapshot shape is identical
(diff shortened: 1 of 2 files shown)


youssofal:
Integrated to `main` as `8d2fe65` after local compile/tests and a live CLI-started server probe.

Verification included a same-session >16K restore check: first turn was 52,521 prompt tokens cold, second turn restored 52,523 cached tokens with only 17 new prefill tokens and no `policy_mismatch`.

Closing this PR as applied.

PR #40, opened 2026-05-09
perf: env-overridable SessionBank.max_entries (final piece of #31 split)

daniel-farina:
## Summary

Adds an `MTPLX_SESSION_BANK_MAX_ENTRIES` env var that mirrors the existing `MTPLX_SESSION_BANK_*_BYTES` overrides from PR #18 (`87af6be`) and the guard tightening in `9d92dac`. Default behavior unchanged (8 entries).

## Why upstream's current code doesn't cover it

`mtplx/engine_session.py` on `main` exposes `_bank_bytes_from_env` for the byte caps but the `max_entries` axis is hard-coded to `SessionBank.__init__`'s default of 8. At ~2 GB per entry on Qwen3.6-27B with 25-50K-token prefixes, 8 entries binds the bank to ~16 GB even when `MTPLX_SESSION_BANK_MAX_BYTES` is set higher; the binding constraint becomes entry count well before bytes, and the LRU starts evicting the longest-matched prefix per session. Operators with headroom need a way to lift this without recompiling. Entry count is a different axis from byte cap; the existing `_bank_bytes_from_env` cannot be reused (K/M/G/T suffixes don't make sense for a count).

The new `_bank_entries_from_env` helper:
- parses a plain integer (no suffix),
- logs a warning + falls back to default on parse error or value < 1 (matches the safety semantics of the byte-cap guards in `9d92dac`),
- treats unset / empty as "use default" without warning.

## Live motivation

5+ opencode subagents x 2-3 prefixes each pushes the bank past 8 entries; without this knob the longest-matched prefix gets evicted every turn and TTFT regresses to cold-prefill cost.

## Test plan

- [x] 19 new unit tests in `tests/test_session_bank_env_caps.py` covering:
  - default when unset / empty
  - valid integer parse (incl. whitespace stripping)
  - invalid value (parametrized: `"abc"`, `"12.5"`, `"1e2"`, `"0x10"`, `""`) -> warning + fallback
  - value < 1 (parametrized: `"0"`, `"-1"`, `"-3"`) -> warning + fallback
  - `EngineSessionManager` wiring (custom value, default, invalid fallback, composition with both byte-cap envs, byte-cap envs alone still work).
- [x] All 35 tests pass alongside the existing `tests/test_engine_session_env.py`:
  ```
  $ pytest tests/test_session_bank_env_caps.py tests/test_engine_session_env.py
  ============================== 35 passed in 0.08s ==============================
  ```

## Background

This is one of three residual pieces from #31 that got split off (disposition recorded separately). #31 itself was opened on stale lineage, and the 16K-cliff piece already landed as #39 cleanly. This is the entry-cap piece.

Closes the entry-cap residual from #31.

### DIFF
--- a/mtplx/engine_session.py
+++ b/mtplx/engine_session.py
@@ -8,6 +8,7 @@ state explicit before the generation loop accepts warm prompt state directly.
 from __future__ import annotations
 
 import hashlib
+import logging
 import os
 import time
 from contextlib import contextmanager
@@ -22,6 +23,15 @@ from .session_bank import (
     DEFAULT_PER_SESSION_MAX_BYTES,
     SessionBank,
 )
+
+
+logger = logging.getLogger(__name__)
+
+
+# Mirrors SessionBank.__init__'s hardcoded max_entries default (kept in sync
+# manually so the operator override here does not silently shift the default
+# for callers that construct SessionBank directly).
+_DEFAULT_BANK_MAX_ENTRIES = 8
 
 
 def _bank_bytes_from_env(name: str, default: int) -> int:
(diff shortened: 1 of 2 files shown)


youssofal:
Integrated to `main` as `92cec48` with the cleanup we discussed: the default is now centralized as `DEFAULT_MAX_ENTRIES` in `session_bank.py`, `engine_session.py` imports that shared default, and `MTPLX_SESSION_BANK_MAX_ENTRIES` still overrides it safely.

Live CLI-started server verification:
- `MTPLX_SESSION_BANK_MAX_ENTRIES=12` reported `session_bank.max_entries=12` from `/admin/sessions`.
- Invalid `MTPLX_SESSION_BANK_MAX_ENTRIES=not-a-number` printed the fallback warning and reported `max_entries=8`.

Closing this PR as applied with Daniel credited as co-author on the maintainer cleanup commit.

PR #41, opened 2026-05-09
perf: postcommit prefix reuse + cache_miss_reason observability (residual of #31)

daniel-farina:
## Summary

Two small additions to `_store_retokenized_history_snapshot` in `mtplx/server/openai.py`:

- Pass `session_bank` (with `template_hash`, `draft_head_identity`, `policy_fingerprint`) to `restore_or_prefill_prompt_state`, so the postcommit re-prefill reuses the previous turn's bank entry as a prefix and only forward-AR's the new suffix.
- Surface `cache_hit`, `cached_tokens`, `suffix_tokens`, `cache_miss_reason` from `prompt_state` in the result dict so the `[mtplx] idle async session postcommit ...` log line records whether the shortcut fired and, if not, why.

## Why upstream's current code doesn't cover it

The current `_store_retokenized_history_snapshot` calls `restore_or_prefill_prompt_state` with only `mtp_hidden_variant` and `mtp_history_policy`:

```python
prompt_state = restore_or_prefill_prompt_state(
    state.runtime,
    history_ids,
    mtp_hidden_variant="post_norm",
    mtp_history_policy="committed",
)
```

`session_bank` is not passed, so the bank lookup never runs and every postcommit pays the full ~18K-token re-prefill cost. The `ModelWorkScheduler` introduced in PR #25 / commit 349fdf7 admits the next foreground request only after the in-flight idle task completes, so this directly translates into a ~30 s stream-silence between turns of any tool-using agent loop on Qwen3.6-27B at 18-25K-token contexts.

The miss-reason field is similarly absent: `prompt_state.cache_miss_reason` exists, but the postcommit result dict and the `[mtplx] idle async session postcommit ...` log line do not surface it.

## What's NOT included (compared to the source #31 bundle)

The original #31 bundle also contained an "Option A" two-stage executor split (orchestration loop on a separate `postcommit_executor`, MLX work submitted back to `generation_executor` via `Future.result()`). That is intentionally NOT brought across:

- PR #25's `ModelWorkScheduler` (commit 349fdf7) is single-owner-thread by design (MLX stream affinity).
- With a single owner thread, foreground gen and idle postcommit are serialized at task boundaries; `state.lock` is never contended across threads while an idle task is the active scheduler item.
- `_store_retokenized_history_snapshot`'s `state.lock.acquire(blocking=False)` therefore succeeds on first try and the 250 ms inter-attempt sleep concern that motivated Option A no longer applies.
- Reintroducing a separate `postcommit_executor` would either fight the scheduler unification or require a compat shim for a configuration that does not exist on `main`.

If a future change reintroduces a dedicated postcommit executor, the Option A diff is archived in our split-disposition record and can be reapplied.

## Live motivation

Source-fork measurements after the change (5-turn tool-call ladder on Qwen3.6-27B):

```
T1 (cold)  cache_hit=false miss=prefix_divergence_at_token
           elapsed_s=31.1 (full re-prefill, expected on cold)
T2 (warm)  cache_hit=true  miss=null cached_tokens=20594
           suffix_tokens=122 elapsed_s=0.75 (95-percentile win)
T3 (warm)  cache_hit=true  miss=null cached_tokens=20716
           suffix_tokens=122 elapsed_s=0.74
```

## Test plan

- [x] New `tests/test_postcommit_prefix_reuse.py` (3 tests) pins:
  1. `session_bank`, `template_hash`, `draft_head_identity`, `policy_fingerprint` are all passed to `restore_or_prefill_prompt_state` (without all four the bank lookup either does not run or rejects the entry as a mismatch).
  2. `cache_hit`, `cached_tokens`, `suffix_tokens`, `cache_miss_reason` are propagated to the result dict on a hit.
  3. `cache_miss_reason` records the miss reason (e.g. `prefix_divergence_at_token`) on a miss.
- [x] All existing postcommit tests pass:
  ```
  $ pytest tests/test_postcommit_tools_plumbing.py tests/test_idle_postcommit_subagent.py tests/test_postcommit_wait.py tests/test_postcommit_wait_integration.py tests/test_postcommit_prefix_reuse.py
  35 passed in 0.34s
  ```
- [x] Server / model-scheduler tests still green:
  ```
  $ pytest tests/test_server_openai.py tests/test_openai_bridge.py tests/test_model_scheduler.py
  64 passed in 0.40s
  ```

Closes the prefix-reuse + miss-reason residual from #31.

### DIFF
--- a/mtplx/server/openai.py
+++ b/mtplx/server/openai.py
@@ -2634,6 +2634,21 @@ def _store_retokenized_history_snapshot(
             "reason": "model_lock_busy_before_retokenized_commit",
             "elapsed_s": time.perf_counter() - started,
         }
+    # Pass `session_bank` (and the matching identity / policy fingerprints)
+    # so the postcommit re-prefill reuses the longest matching prefix from
+    # a prior turn instead of re-prefilling the full ~18K-token history
+    # from scratch. On consecutive tool-calling turns this collapses
+    # postcommit cost from ~27 s (full re-prefill) to ~1 s (suffix forward
+    # only). The next foreground request, which the model_scheduler admits
+    # only after this idle task completes, therefore queues ~1 s instead
+    # of ~30 s behind the postcommit.
+    #
+    # The bank already contains an entry for the previous turn's history
+    # (this same function stored it on the prior postcommit). The new
+    # turn's history starts with that previous-turn prefix verbatim
+    # (chat-template encoding is deterministic for the same messages and
+    # tools), so `longest_prefix` matches and only the new user turn +
+    # assistant turn need to be forward-AR'd.
     try:
         with attention_phase("postcommit"):
             prompt_state = restore_or_prefill_prompt_state(
@@ -2641,6 +2656,10 @@ def _store_retokenized_history_snapshot(
                 history_ids,
                 mtp_hidden_variant="post_norm",
                 mtp_history_policy="committed",
+                session_bank=state.sessions.bank,
+                template_hash=state.template_hash,
+                draft_head_identity=state.draft_head_identity,
+                policy_fingerprint=policy_fingerprint,
             )
         mtp_snapshot = (
             snapshot_cache(prompt_state.committed_mtp_cache)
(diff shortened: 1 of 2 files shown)


youssofal:
Integrated to `main` as `68c9ef6`, with a maintainer follow-up in `2be4ebf` to expose the postcommit cache-reuse fields through the public `mtplx_stats.session_postcommit_snapshot` payload as well.

Live CLI-started server verification produced a parsed tool call and a `retokenized_history` postcommit with `cache_hit=true`, `cached_tokens=330`, `suffix_tokens=1`, and `cache_miss_reason=null`.

Closing this PR as applied.

PR #44, opened 2026-05-10
fix: don't abort sustained-profile sessions on dense paged-attention fallback

daniel-farina:
## Summary

\`long_context_dense_fallback_forbidden\` (cache_state.py:926-933) raises when either \`MTPLX_SUSTAINED_PREFILL\` (the product flag set by \`--profile sustained\`) OR \`MTPLX_ASSERT_NO_PAGED_ACTIVE_ARRAYS\` (QA bench assertion) is set and the cache offset crosses the partition threshold (default 2048 tokens, set by the \`sustained\` profile in profiles.py:51).

Coupling those two flags means every sustained-profile user hits a benchmark abort in production as soon as they get past 2048 cached tokens AND any layer's partitioned-paged kernel returns None. The dense \`scaled_dot_product_attention\` fallback right below the abort in attention_split.py:260-263 is a correct continuation — the trip-wire is leftover from kernel qualification.

## Repro

Run a long-form interactive chat session with \`--profile sustained --context-window 128000\`. Around round 2-3 (well below the 128K limit, e.g. 25K context) a request fails fast (~1s, before any tokens stream) with:

\`\`\`
RuntimeError/500: Sustained long-context paged attention attempted dense cache.state fallback after the partition threshold
\`\`\`

Subsequent requests on the same session keep failing the same way until the session is rotated.

## Fix

Drop \`MTPLX_SUSTAINED_PREFILL\` from the disjunction. Only QA benches that explicitly set \`MTPLX_ASSERT_NO_PAGED_ACTIVE_ARRAYS\` still trip the assertion. Production sustained sessions degrade gracefully to dense SDPA on partitioned-kernel misses.

## Test plan

- [ ] Restart server with \`--profile sustained\` (no env vars), run interactive chat past 2048 tokens, observe \`mtplx_paged_attention_bailout\` events but no errors and continued generation.
- [ ] Re-run the QA bench with \`MTPLX_ASSERT_NO_PAGED_ACTIVE_ARRAYS=1\` set, verify the assertion still trips as intended.
- [ ] Confirm \`MTPLX_ALLOW_LONG_CONTEXT_DENSE_FALLBACK=1\` continues to work as a per-session escape hatch.

### DIFF
--- a/mtplx/cache_state.py
+++ b/mtplx/cache_state.py
@@ -928,9 +928,14 @@ class VllmMetalPagedKVCache:
             return False
         if _env_truthy("MTPLX_ALLOW_PAGED_ACTIVE_ARRAY_SNAPSHOT"):
             return False
-        sustained = _env_truthy("MTPLX_SUSTAINED_PREFILL")
+        # Only the QA assertion env var should turn the dense-fallback path
+        # into a hard error. MTPLX_SUSTAINED_PREFILL is a *product* profile
+        # flag (set by `--profile sustained`) and must not abort production
+        # requests when the partitioned-paged kernel returns None - the
+        # caller in attention_split.py already handles None by routing to
+        # scaled_dot_product_attention with cache.state, which is correct.
         asserted = _env_truthy("MTPLX_ASSERT_NO_PAGED_ACTIVE_ARRAYS")
-        return (sustained or asserted) and int(self.offset) >= self._partition_threshold()
+        return asserted and int(self.offset) >= self._partition_threshold()
 
     def long_context_dense_fallback_forbidden(self) -> bool:
         return self._long_context_dense_fallback_forbidden()


daniel-farina:
testing wif real coding session

daniel-farina:
Going into a deeper hole... 

Currently investigating hilacinatious starting at 29k context and low acceptance rates. speculative-decoding draft-head degradation at long context

image

youssofal:
> Going into a deeper hole... 
> 
> 
> 
> Currently investigating hilacinatious starting at 29k context and low acceptance rates. speculative-decoding draft-head degradation at long context
> 
> 
> 
> image
> 
> 


Are you using turboquant?

daniel-farina:
No - turboquant is off. The server invocation has no `--turboquant` flag and the sustained profile defaults `MTPLX_VLLM_METAL_PAGED_TURBOQUANT=0` (profiles.py:103). The only quantization on this run is 3-bit affine on the draft LM head (`--draft-lm-head-bits 3 --draft-lm-head-group-size 64 --draft-lm-head-mode affine`); trunk KV is unquantized.

The hallucination at 29K with declining MTP acceptance (~47% / 33% / 27%) appears to be the warm-restore position-offset asymmetry: `generation.py:1395` calls `_append_mtp_history(suffix)` without `position_offset`, while the cold-prefill path at `:1933` passes it. When a `committed`-stored SessionBank entry is reused for a `last_window` lookup, M-RoPE drifts on the draft head, the verifier accepts plausibly-shaped garbage drafts, and the model emits training-data fragments mid-stream above the `..._WINDOW_THRESHOLD` (16384).

Locally we worked around it by forcing `MTPLX_MTP_HISTORY_POLICY=committed` so sessions never flip to `last_window` — that stops the symptom and warm caching engages normally. PR #46 makes the env var actually honored on the `auto` resolution path the sustained profile uses (the gate at `_resolve_mtp_history_policy` only fired on explicit `committed`).

So: this PR is correct on its own; the abort it relaxes is a leftover bench trip-wire. The corruption you saw is a separate warm-restore bug that the abort was masking. The proper fix for that is to thread `position_offset` through `restore_or_prefill_prompt_state` so cross-policy reuse stays aligned, but that's beyond the scope of this PR.

daniel-farina:
I will test with turboquant on as well and check results, the model had it off to make sure we are getting optimal quality since we had enbough memory but in reality we should try turboquant on by default probably.

daniel-farina:
Update with empirical findings from running the patched server against a 27K-context resumed session over the past hour:

### Setup
- Running with this PR + `MTPLX_MTP_HISTORY_POLICY=committed` (workaround until #46 lands so the env var is honored on `auto`)
- Same flags as before: `--profile sustained --depth 3 --generation-mode mtp`, no `--turboquant`
- Resumed an existing coding-agent session with 43 prior messages (~27K tokens of context, lots of `tool_call_history_rewrite` postcommits in the bank)

### What we saw
1. **First-turn-after-restart cold prefill** is real but expected (~40s for ~27K). Subsequent turns hit `cache_hit: true` warmly with `cached_tokens: ~prompt_tokens`. Caching itself is healthy.
2. **Hallucination at long context is gone** in the obvious form (`== Clarke`, `[TrackedtView]` fragments) thanks to forcing `committed` policy.
3. **New failure modes surfaced** at this context size:
   - `ValueError/500: SparseDistribution probabilities must have positive mass` from `mtplx/sampling.py:40` - the trunk's next-token distribution arrives with zero/NaN total mass on some rounds.
   - `422 malformed tool_call: unclosed <tool_call> block` - the model occasionally truncates a `<tool_call>` block mid-stream.
   - Token repetition in tool args (`pedestrian pedestrian pedestrian pedestrian`) - sampling-quality degradation visible even when generation didn't crash.
4. **Tested with `MTPLX_VLLM_METAL_PAGED_TURBOQUANT=1`** to see if the alternate codepath dodges the NaN. Result: SparseDistribution crash gone, but the model **silently hallucinated work it never did** - claimed it had edited NPC files; `git status` showed zero changes. The visible crash got replaced with a confident lie, which is strictly worse for an autonomous agent.

### Question for you
Two reads on this:
1. The NaN distribution at 27K + accumulated `tool_call_history_rewrite` is a *separate* bug from the warm-restore position-offset asymmetry I described earlier; clean sessions never hit it. The `committed`-policy workaround sidesteps the M-RoPE drift but the NaN seems to come from somewhere else (numerical drift in dense fallback at long context? Cache snapshot state where `_active_arrays()` returns something the SDPA path interprets as zeros?).
2. Or it *is* the same root cause and we just haven't traced the right path - both could be the warm-restore asymmetry if `_append_mtp_history(suffix)` without `position_offset` produces a draft cache where successive verifications produce all-zero residual probabilities.

Does either reading match your intuition? And is there a specific code path you'd point to for chasing the NaN trunk distribution at long context with paged + dense fallback in play? Happy to keep digging - this PR is sound on its own (the bench-abort *was* leftover) but you can see the long-context cliff is deeper than just that one trip-wire.

For now we've documented the empirical workaround (force `committed` via env, run with turboquant off, expect occasional malformed tool_call which clients can handle by dropping the assistant turn). Marking #46 ready next so the env override stops being silently ignored.

daniel-farina:
btw The default q3_0 is using the Lloyd-Max 3-bit codebook (8 non-uniform centroids for normally-distributed values). My hypothesis: at 27K context, 3-bit V error compounds into garbage; bumping V to q5_0 or q8_0 should help


I will test different turboquant settings and then look into automating this workflow.

daniel-farina:
I think we need to auto adjust to q5_0 or q8_0 once we start larger context conversations, still testing!

daniel-farina:
Followup empirical data on TurboQuant V-quant precision at long context, since you raised it - this turned out to be the meaningful knob.

### Test: same 27K-context resumed session, two-turn agentic coding task

| Settings | Tool-call validity | Real edit landed | Trunk crash | Notes |
|---|---|---|---|---|
| TurboQuant OFF | malformed mid-stream | no | `SparseDistribution: positive mass` (NaN) | hard crash before edits |
| TurboQuant ON, K=q8_0, **V=q3_0** (sustained default) | ~50% malformed | no | malformed `<tool_call>` truncation | model also **silently hallucinated** finished work in turn 2 |
| TurboQuant ON, K=q8_0, **V=q5_0** | clean tool_calls | **yes** (`npcs.js` +37 net lines, articulated humanoid mesh hierarchy with hip/shoulder pivots, proper proportions) | NaN still fires on a later round | edit landed cleanly, NaN happens after the work commits |

The `q3_0` -> `q5_0` bump alone made the difference between "model invents fictional work" and "model produces a real, correct, well-structured patch and runs the build."

### Why this matches the TurboQuant paper

arXiv:2504.19874 (TurboQuant) Section 4: "absolute quality neutrality with 3.5 bits per channel and marginal quality degradation with 2.5 bits per channel." `q3_0` is 3 bpw - below the neutrality line. `q5_0` (5 bpw) is comfortably above. Long-context error compounds across thousands of attention reads, so the V quant is where this bites first.

### Proposing two changes

1. **Bump sustained-profile default `MTPLX_VLLM_METAL_PAGED_TURBOQUANT_V_QUANT` to `q5_0`.** Sustained's whole point is long-context throughput; the current 3-bit default fails its own use case. Memory cost is small (~20% bigger V-cache vs q3_0; still ~2x smaller than full bf16). PR coming next.
2. **Auto-tune V-quant from `--context-window` at server start.** When `--context-window` > 16384, default V to `q5_0` regardless of profile; below, q3_0 is fine. Server already knows the flag at startup, the user just gets the right defaults without thinking.

Auto-adjusting *mid-session* isn't viable since the KV pages are committed at the chosen quant - you can't read q3_0 pages as q5_0. So it has to be a startup setting. Does the auto-from-context-window approach align with how you'd want defaults handled?

The trunk-side `SparseDistribution: positive mass` NaN at long context is independent of TurboQuant settings (fires both with and without quant) - that's a separate investigation. Filing as its own issue.

daniel-farina:
> > Going into a deeper hole...
> > Currently investigating hilacinatious starting at 29k context and low acceptance rates. speculative-decoding draft-head degradation at long context
> > image
> 
> Are you using turboquant?

This was a good question, I learned a ton about turboquant and its making the coding agent so much better.  Thanks !

daniel-farina:
Tracking issue filed: youssofal/MTPLX#49 (covers this PR + #46/#47/#48 + the two not-yet-filed bugs - position_offset asymmetry at `generation.py:1395` vs `:1933`, and the trunk SparseDistribution NaN). Integration branch `daniel-farina/MTPLX:daniel/dev-stack` rolls up all four PRs if you want to test the whole stack at once.

youssofal:
Applied to main in da7cf8b, with #46 and the follow-up env/profile hardening in 0c3bef2.\n\nLocal validation included focused regression coverage for the sustained dense-fallback guard and post-merge CLI smoke on the local Optimized-Speed model. Closing this PR as landed/applied.

PR #46, opened 2026-05-10
fix: honor MTPLX_MTP_HISTORY_POLICY env override on 'auto'

daniel-farina:
## Summary

\`_resolve_mtp_history_policy\` only applied the \`MTPLX_MTP_HISTORY_POLICY\` env override when \`requested == \"committed\"\`. The sustained profile sets the requested policy to \`\"auto\"\`, so the env var was silently dropped on every sustained server hot path. Users had to edit the profile to test alternative policies.

## Fix

Extend the gate to also fire on \`\"auto\"\`. Explicit non-default caller policies (\`last_window\`, \`cycle\`, etc.) still take precedence.

## Use case

Combined with #44, this lets users force \`committed\` policy via env var to work around a separate warm-restore position_offset asymmetry in \`generation.py\`'s \`restore_or_prefill_prompt_state\` (which I'm tracking but not fixing here - it requires threading position_offset through the warm path). Without that workaround, sustained-profile sessions above 16K tokens trigger draft-head misalignment, model emits training-data fragments mid-stream, and clients see 422 \`malformed tool_call\` errors.

### DIFF
--- a/mtplx/generation.py
+++ b/mtplx/generation.py
@@ -161,7 +161,13 @@ def _mtp_history_last_window_tokens() -> int:
 def _resolve_mtp_history_policy(requested_policy: str, prompt_tokens: int) -> str:
     requested = _normalize_mtp_history_policy(requested_policy)
     env_policy = os.environ.get("MTPLX_MTP_HISTORY_POLICY")
-    if env_policy and requested == "committed":
+    # Honor the env-var override whenever the caller requested either the
+    # default "committed" or the auto-resolution path. Previously this only
+    # fired when ``requested == "committed"``, which silently dropped the
+    # override in every server hot path that resolves to "auto" via the
+    # sustained profile (profiles.py:93). Sustained users could not opt
+    # out of the last_window flip without editing the profile.
+    if env_policy and requested in ("committed", "auto"):
         requested = _normalize_mtp_history_policy(env_policy)
     if requested != "auto":
         return requested


daniel-farina:
Status: this is the env-override fix that makes the `MTPLX_MTP_HISTORY_POLICY=committed` workaround actually take effect on the sustained profile (which resolves to `auto`, not explicit `committed`). Without this PR the documented escape hatch is silently dropped.

Empirically confirmed against a 27K-context resumed coding-agent session: forcing `committed` via env eliminates the M-RoPE drift hallucinations described in #44's thread (`== Clarke`, `[TrackedtView]` fragments).

Depends on #44 in practice (the bench abort fires first and 500's the request before policy resolution matters). See youssofal/MTPLX#49 for the tracking issue covering the full long-context picture, and `daniel-farina/MTPLX:daniel/dev-stack` for an integration branch with all four PRs stacked.

Note: this is a workaround, not a fix - the underlying warm-restore `position_offset` asymmetry between `generation.py:1395` and `:1933` is described in #49 and still wants a proper PR.

youssofal:
Applied to main in e7f3697, with the profile-env precedence hardening in 0c3bef2 so MTPLX_MTP_HISTORY_POLICY=committed actually survives the sustained profile.\n\nValidated through targeted pytest, local inspect, prefill-ladder, and a two-request OpenAI-compatible server smoke. Closing this PR as landed/applied.

PR #47, opened 2026-05-10
profiles: bump sustained-profile TurboQuant V-quant default q3_0 -> q5_0

daniel-farina:
## Summary

Pin sustained profile's TurboQuant defaults to `K=q8_0`, `V=q5_0` (the latter being a bump from the implicit `q3_0` default in `turboquant.py`). Aligns the long-context-throughput profile with the TurboQuant paper's quality threshold.

## Why

The [TurboQuant paper](https://arxiv.org/abs/2504.19874) Section 4.2/4.3 reports "absolute quality neutrality with 3.5 bits per channel and marginal quality degradation with 2.5 bits per channel." The current 3-bit V default sits *below* the neutrality line. Per-token quantization error compounds across thousands of attention reads at long context - exactly where the sustained profile is supposed to shine.

## Empirical evidence

Same model (qwen3.6-27b-MTPLX), same 27K-context coding-agent task on a resumed hip session, two-turn run:

| Settings | Tool-call validity | Real edit landed | Trunk crash | Notes |
|---|---|---|---|---|
| TurboQuant OFF | malformed mid-stream | no | `SparseDistribution: positive mass` (NaN) | hard crash before edits |
| ON, K=q8_0, **V=q3_0** (current default) | ~50% malformed | no | `<tool_call>` truncation | model **silently hallucinated** finished work in turn 2 |
| ON, K=q8_0, **V=q5_0** (this PR) | clean | **yes** (+37-line articulated humanoid mesh patch) | NaN still fires later, separate bug | edit landed, build passed |

The `q3_0 -> q5_0` bump alone made the difference between "model invents fictional work" and "model produces a real, correct, well-structured patch and runs the build." Cross-referenced to the [vllm-metal TurboQuant docs](https://docs.vllm.ai/projects/vllm-metal/en/latest/turboquant/) which match the paper's bpw guidance.

## Cost

Memory-wise, q5_0 V is ~67% larger than q3_0 V but still ~40% smaller than full bf16. For a 27K-token KV cache on a 27B model, this is single-digit gigabytes - cheap on Apple Silicon unified memory.

## Followup ideas

The trunk-side `SparseDistribution: positive mass` NaN is a separate long-context bug independent of TurboQuant settings; will file separately. Auto-tuning V-quant from `--context-window` at server start is also possible (server already knows the flag) and would eliminate the manual tradeoff for users; happy to draft that as a follow-up if you'd prefer it over the static default bump here.

Users with extreme memory pressure can still pin V back via `MTPLX_VLLM_METAL_PAGED_TURBOQUANT_V_QUANT=q3_0`.

## Test plan

- [x] Profile dict imports cleanly, `PROFILES['sustained'].env_dict()` shows the new defaults.
- [x] Empirical 27K-context coding-agent run with the new defaults landed real work.
- [ ] Existing TurboQuant smoke tests (none currently in tree that I found - happy to add one if there's a preferred shape).

### DIFF
--- a/mtplx/profiles.py
+++ b/mtplx/profiles.py
@@ -101,6 +101,18 @@ SUSTAINED_PREFILL_ENV = {
     "MTPLX_VLLM_METAL_PAGED_PARTITION_THRESHOLD": "2048",
     "MTPLX_VLLM_METAL_PAGED_PARTITION_SIZE": "512",
     "MTPLX_VLLM_METAL_PAGED_TURBOQUANT": "0",
+    # When TurboQuant is enabled, default V-quant to 5-bit instead of 3-bit.
+    # The 3-bit Lloyd-Max default sits *below* the TurboQuant paper's
+    # 3.5-bpw "absolute quality neutrality" threshold (arXiv:2504.19874
+    # Section 4.2/4.3), so quality drifts at long context - exactly the
+    # use case the sustained profile is meant to target. Empirically,
+    # 27K-context coding-agent runs with K=q8_0 / V=q3_0 produced
+    # malformed tool_call payloads and silent hallucination, while the
+    # same task on the same model with V=q5_0 landed correct edits and
+    # passed the build. q5_0 is still ~2x smaller than full bf16 for
+    # values, so the memory cost of the bump is small.
+    "MTPLX_VLLM_METAL_PAGED_TURBOQUANT_K_QUANT": "q8_0",
+    "MTPLX_VLLM_METAL_PAGED_TURBOQUANT_V_QUANT": "q5_0",
     "MTPLX_CLEAR_CACHE_EVERY": "auto",
     "MTPLX_CLEAR_CACHE_EVERY_CONTEXT_THRESHOLD": "16384",
     "MTPLX_CLEAR_CACHE_EVERY_LONG_CONTEXT": "256",


daniel-farina:
Status: empirically confirmed against a fresh resumed 27K-context coding-agent session, two-turn task. Full data table in #44's thread; the headline:

- V=q3_0 (current sustained default): ~50% malformed `<tool_call>`, silent hallucination of fictional work
- V=q5_0 (this PR): clean tool_calls, real `npcs.js` +37 LOC patch landed and built

This is the difference between an unusable agent and a working one at >16K context. Justification matches TurboQuant paper (arXiv:2504.19874 Sec. 4) - 3 bpw is below the neutrality line, 5 bpw is comfortably above.

Stacked under #48 which auto-tunes from `--context-window` so users on long-context sustained get this default without thinking. See youssofal/MTPLX#49 for the long-context tracking issue, and `daniel-farina/MTPLX:daniel/dev-stack` for an integration branch with #44/#46/#47/#48 rolled up.

youssofal:
I tested this locally after landing #44/#46 plus the profile-env preservation fix on main. Current result: #47 is not mergeable yet.\n\nWhat passed:\n- 32k product-layout q3: prompt TPS 608.9, decode 38.6, TTFT 53.9s, peak 27.4GB, route contiguous_dense_decode.\n- 32k product-layout q5: prompt TPS 599.9, decode 39.4, TTFT 54.6s, peak 27.4GB, route contiguous_dense_decode.\n\nThose rows are mostly a regression check, not proof for V-quant quality, because the product 32k route never exercised paged/TurboQuant V cache.\n\nMechanism check:\n- Forced repage 32k q3 succeeded: prompt TPS 561.7, decode 24.1, TTFT 58.3s, peak 22.2GB.\n- Forced repage 32k q5 failed before generation with:\n\n\n\nSo the q5 default cannot land until the actual TurboQuant implementation supports q5_0 and the q3/q5 comparison is rerun with the env vars truly active. Main now preserves the relevant env overrides, so the next test will not silently run TurboQuant off.

youssofal:
Exact forced-repage q5 failure line, since the previous comment lost the code block formatting:

```text
ValueError: MTPLX TurboQuant currently supports v_quant='q3_0' for the Metal paged diagnostic path
```

daniel-farina:
Closing per your read - q5_0 isn't supported by the Metal paged kernel today (`cache_state.py:748` enforces `value_bits == 3`), so the default bump is premature. Reopening or filing a successor will make sense once TurboQuant supports q5_0/q4_0 and a real q3 vs q5 quality comparison can be run with the env-preservation fix you landed on main.

PR #48, opened 2026-05-10
server: auto-tune TurboQuant V-quant from --context-window at startup

daniel-farina:
Picks `q5_0` for `context_window > 16384` (TurboQuant paper's bpw-vs-quality cliff per arXiv:2504.19874 Section 4.2), `q3_0` below. Only fires when the user hasn't pinned `MTPLX_VLLM_METAL_PAGED_TURBOQUANT_V_QUANT` explicitly. Profile defaults still take precedence (#47 sets sustained's default to q5_0 explicitly).

Hooks into `ServerState.__init__` right after `context_window` is resolved and before cache allocation, since dynamic mid-session quant switching isn't possible (KV pages are committed at the chosen quant). Logs the chosen value to startup output for visibility.

Stacked on top of #47 conceptually: that PR makes q5_0 the explicit sustained default; this one makes it the default for *any* invocation that asks for a long context (e.g. throughput/balanced profiles tuned for short context but spun up with --context-window 128000).

Marking draft until #47 lands - if you'd rather I roll this into #47 as a single change, happy to do that.

Test plan:
- [x] Verified at REPL: 20K -> q5_0, 8K -> q3_0, explicit override preserved.
- [ ] CI smoke test (none currently in tree for TurboQuant flag picking).
- [ ] Empirical run on the 27K-context coding-agent reproducer to confirm no regression vs explicit q5_0.

### DIFF
--- a/mtplx/server/openai.py
+++ b/mtplx/server/openai.py
@@ -799,6 +799,7 @@ class ServerState:
             else _resolve_context_window(self.runtime.tokenizer, args.model)
         )
         _startup_line(f"[5/6] Context window: {self.context_window} tokens")
+        _autotune_turboquant_for_context(self.context_window)
         self.sessions = EngineSessionManager()
         self.last_metrics: list[dict[str, Any]] = []
         # Activity timestamps used by the parent-process thermal watchdog to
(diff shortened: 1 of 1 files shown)


daniel-farina:
Stacked on #47. Auto-promotes V-quant to q5_0 when `--context-window > 16384` so sustained-profile users on long context get the right default automatically (mid-session tuning isn't viable since KV pages are committed at the chosen quant). See youssofal/MTPLX#49 for the tracking issue with the full empirical table and reproducer, and `daniel-farina/MTPLX:daniel/dev-stack` for an integration branch with all four PRs rolled up.

daniel-farina:
Closing - this auto-tunes V-quant from --context-window, but the only kernel-supported V-quant today is q3_0 (#47 hit the same wall). Will revisit when the kernel supports more V-quants.

Issue #49, opened 2026-05-10
Long-context (>16K) reliability and quality regressions on sustained profile - tracking


daniel-farina:
## Symptom

Tool-using coding-agent clients running against `--profile sustained` start degrading once cumulative context crosses ~16K and fail hard around 27-29K. Across a handful of resumed multi-turn agent sessions we have seen:

- Hallucinated `<tool_call>` blocks (training-data fragments like `== Clarke`, `[TrackedtView]`) emitted mid-stream
- Truncated/unclosed `<tool_call>` blocks (client-side `422 malformed tool_call`)
- Token repetition inside tool args (`pedestrian pedestrian pedestrian ...`)
- MTP draft acceptance collapsing turn-over-turn (~47% -> 33% -> 27%)
- `ValueError/500: SparseDistribution probabilities must have positive mass` from `mtplx/sampling.py:40` (trunk distribution arrives with zero/NaN total mass)
- With `MTPLX_VLLM_METAL_PAGED_TURBOQUANT=1` and the default V-quant the visible crash gets replaced by silent hallucination of finished work the model never did - strictly worse for an autonomous agent

Symptoms appear above the `..._WINDOW_THRESHOLD` (16384) and compound with context length. Reproduces on the qwen3.6-27b-MTPLX model with the standard sustained-profile flags (`--depth 3 --generation-mode mtp --verify-strategy capture_commit --reasoning-parser qwen3`); the agent client just needs to drive a long resumed conversation with tool calls in the bank.

## Mapping symptoms to root causes

| Cause | Status |
|---|---|
| Bench abort masking dense fallback in \`cache_state.py\` 500's sustained sessions before the dense path can serve | PR #44 |
| \`_resolve_mtp_history_policy\` ignores \`MTPLX_MTP_HISTORY_POLICY\` when profile resolves to \`auto\` (only fires on explicit \`committed\`); the env var was the documented escape hatch | PR #46 |
| Sustained-profile \`MTPLX_VLLM_METAL_PAGED_TURBOQUANT_V_QUANT=q3_0\` is below the TurboQuant paper's neutrality line (3.5 bpw, arXiv:2504.19874 Sec. 4); long-context error compounds across thousands of attention reads. Bumping to \`q5_0\` flipped a real test from 'invents fictional work' to 'lands a real, correct edit' | PR #47, auto-tune in PR #48 |
[...]

PR #51, opened 2026-05-10
fix: guard trunk sampler against NaN logits at long context

daniel-farina:
Tracking issue: youssofal/MTPLX#49.

## Diagnosis

The trunk logit tensor can contain NaN/Inf at long context (typically >25K tokens with sustained profile) due to fp16 cancellation in dense SDPA fallback paths (`cache_state.py:_large_q_split_sdpa_fallback` or `attention_split.py:264`'s SDPA over ~25K-row K). This propagates through the sampler:

```
mx.logsumexp(flat, axis=-1)       -> NaN
mx.exp(top_vals - log_total)      -> NaN probabilities
probs.sum() <= 0                  -> rescue check SKIPPED (IEEE-754: NaN<=0 is False)
SparseDistribution(probs=[NaN])   -> raises at sampling.py:39
```

Result: `ValueError/500: SparseDistribution probabilities must have positive mass`. When only some logit positions are NaN, sampling instead picks an off-grammar token and produces a truncated `<tool_call>` block (`422 malformed tool_call`). Same root cause, two failure modes.

## Fix (sampler-boundary mitigation)

- Sanitize non-finite logits with `mx.where(mx.isfinite, flat, -1e30)` before `argpartition`/`logsumexp` in all three `fast_sampling.py` entry points.
- Replace `<= 0` rescue checks with `not (>0)` so NaN totals trip rescue (IEEE-754 fix).
- Defensive per-row one-hot collapse if residual NaN survives upstream sanitize.
- Mirror the fix in the numpy `softmax` path at `sampling.py:63`.
- Env gate `MTPLX_TRUNK_LOGITS_NAN_GUARD=0` to disable for debugging; `MTPLX_TRUNK_LOGITS_NAN_GUARD_QUIET=1` to silence per-fire log line.

## Empirical validation

Same bricked 27K coding-agent session before/after:

| Probe | Before fix | After fix |
|---|---|---|
| One-shot at 26K | 500 `SparseDistribution` OR 15min stream timeout | 40s, `finish_reason=stop`, ctok=80, real answer |
| Tool-using turn at 26K | 422 `malformed tool_call` on round 1 | 9 valid tool-calling rounds, 422 on round 9, recovered |
| Push to 29K | unreachable | reachable, productivity degrades but no crashes |

The 422 failure mode still surfaces occasionally (the degenerate-but-not-fully-NaN form), but at lower frequency, and is recoverable client-side. The hard 500 NaN is gone.

## Scope

This is intentionally a sampler-boundary mitigation, not the upstream attention fix. The fp16 cancellation in dense SDPA at long context is a deeper bug that wants instrumented forward-pass diagnostics; #49 tracks it as a follow-up.

## Tests

13 sampling/fast-sampling tests pass, including 6 new regression tests covering single-NaN, all-NaN, batched per-row NaN, and full-row NaN cases.

Stacked on `daniel/dev-stack` which also rolls up #44 + #46 + #47 + #48.

### DIFF
--- a/mtplx/fast_sampling.py
+++ b/mtplx/fast_sampling.py
@@ -2,10 +2,60 @@
 
 from __future__ import annotations
 
+import os
+
 import mlx.core as mx
 import numpy as np
 
 from .sampling import SamplerConfig, SparseDistribution
+
+
+def _trunk_logits_nan_guard_enabled() -> bool:
+    """Return whether the NaN/Inf guard on trunk logits is active.
+
+    Defaults to enabled. Set ``MTPLX_TRUNK_LOGITS_NAN_GUARD=0`` to disable
+    (only for debugging - disabling re-exposes the long-context
+    ``SparseDistribution probabilities must have positive mass`` 500.).
+    """
+
+    raw = os.environ.get("MTPLX_TRUNK_LOGITS_NAN_GUARD")
+    if raw is None:
+        return True
+    return raw.strip().lower() not in {"0", "false", "no", "off"}
+
+
+def _sanitize_logits_row(row: np.ndarray) -> tuple[np.ndarray, bool]:
+    """Replace non-finite entries with ``-inf`` so argpartition/logsumexp behave.
+
+    Returns the (possibly copied) row and a flag indicating whether any
+    sanitization happened. The caller uses the flag to log diagnostic state
+    so we can tie production crashes back to upstream attention NaN sources.
+    """
+
+    finite_mask = np.isfinite(row)
+    if bool(finite_mask.all()):
+        return row, False
+    sanitized = np.where(finite_mask, row, -np.inf)
+    return sanitized, True
+
+
+def _record_trunk_logits_nan(*, batched: bool, rows: int) -> None:
+    """Log a single line per occurrence so operators can see frequency.
+
+    Kept intentionally cheap (one line, no traceback) - the goal is to
+    surface that the guard fired without spamming the log when an entire
+    batch hits the same upstream NaN.
+    """
+
+    if os.environ.get("MTPLX_TRUNK_LOGITS_NAN_GUARD_QUIET"):
+        return
+    import sys
+
+    print(
+        "mtplx_trunk_logits_nan_guard "
+        f"batched={int(bool(batched))} rows_with_nan={int(rows)}",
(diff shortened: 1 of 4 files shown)


daniel-farina:
Quick status while waiting for review:

Empirically the sampler-boundary NaN guard in this PR is doing what it claims — at >25K context the trunk distribution is degenerate often enough that without the guard we get hard `SparseDistribution: positive mass` 500s, and with it we get either a recovered token or a clear malformed-tool_call which clients can drop and retry. So it's a useful mitigation.

That said, we're now investigating the upstream attention numerical issue that's causing the degenerate trunk logits in the first place — this PR doesn't fix that, it just contains the symptom. If our investigation finds a clean upstream fix, we'll prefer that and may end up closing this PR in favor of an attention-side patch. If it turns out fp16 cancellation in the long-context dense SDPA path is intrinsic and not fixable cheaply, then this PR's mitigation is the right shape and we'll ask for review.

Marking this as WIP-status-comment, no code changes; will reply again with the diagnosis.

daniel-farina:
Closing this PR.

Spent the past hour empirically analyzing 1557 generations across the dev-stack server. Result: **zero NaN-guard fires recorded**. The sampler-boundary mitigation in this PR is correct in spirit but never actually triggers in practice — the supposed long-context cliff isn't a NaN, it's just O(N²) prefill latency on the 16 global-attention layers of qwen3.6-27b crossing client-side timeouts.

What we thought was a numerical-stability bug:
- malformed unclosed `<tool_call>` blocks → degenerate sampling, 1.5% rate, fix is sampler constraint not NaN guard
- truncated streams at >25K → just slow prefill (43K-token prompts complete in ~80s)
- `mtplx_stream_silence` events → fire during prefill, not mid-stream stall (`server/openai.py:6228`)

PR #44 (the bench abort relax) was the right fix because the abort genuinely fired. PR #46 (env override) was the right fix because the env var was silently dropped. This PR was solving a problem that doesn't exist on this setup.

Net: closing as superseded by 'don't panic, just wait for prefill'. The real long-context issue is throughput, not numerical, and the fix is client-side timeout tolerance, not server-side mitigation. If a future setup actually produces NaN in attention, the diagnosis stack here will help find it quickly.

PR #55, opened 2026-05-11
Commit only client-visible content to SessionBank, not reasoning (fixes #54)

### LINKED ISSUE #54: SessionBank commits include reasoning_content but clients don't echo it back → every multi-turn cache miss with prefix_divergence_at_token
## Summary

In long-lived multi-turn sessions (any client reusing `x-mtplx-session-id` across requests, e.g. `coding-tool --resume`, `opencode` with `chat_id`), the SessionBank is populated but **every subsequent turn cold-misses with `cache_miss_reason=prefix_divergence_at_token`**. The committed token sequence is not a prefix of the next request's prompt.

Root cause: `streamed_history_content()` in `mtplx/server/openai.py` (the function that builds `assistant_content` for the postcommit snapshot during streaming responses) splices both `history_content_chunks` AND `history_reasoning_chunks` back into the stored content. The SSE protocol however delivers them as two different fields - `delta.content` and `delta.reasoning_content` - and OpenAI-style clients don't round-trip the `reasoning_content` field. The next-turn request therefore renders an empty `<think></think>` block where the bank commit baked in the real thinking text. Token sequences diverge at the assistant boundary, restore fails.

## Affected versions

Present on `main` at commit `e5d05ad` (latest at time of report). Introduced by the `streamed_history_content()` definition that always concatenates `reasoning` into the committed string.

## Reproduction (any reasoning-parser-enabled run)

Server:
```
mtplx-server ... --reasoning-parser qwen3 --no-stats-footer
```

Client (hip 0.3.24):
```
SESS="bug-repro-$(date +%s)"
hip --session "$SESS" --quiet --stats --max-tokens 60 "What color is the sky? One sentence."
hip --resume "$SESS" --quiet --stats --max-tokens 60 "And grass?"
hip --resume "$SESS" --quiet --stats --max-tokens 60 "And snow?"
```

Then inspect:
```
curl -sS http://127.0.0.1:8088/admin/sessions | jq '.session_bank.prefixes[] | {session_id, prefix_len, hits}'
```

**Observed (before fix):**
- All 3 turns: `restore=cold`, `cached_tokens=0`, `cache_miss_reason=prefix_divergence_at_token`
- Bank shows 3 entries (one per turn) all with `hits=0`
- Bank entries are LONGER than each subsequent request prompt
- TTFT stays at ~3s across all turns

## How we proved it

We added two diagnostic prints temporarily:

1. In `_history_ids_for_postcommit`, decode the last 40 tokens of the committed prefix and log them.
2. At the request-tokenization site (right after the `_encode_messages` call in the chat-completions handler), do the same for the incoming prompt.

Reproducing the test above produced this side-by-side (key turn shown):

```
[postcommit] sentinel_len=2185 tail_text="...don't require any tool usage - I can answer directly from my knowledge.\n</think>\n\nThe sky appears blue during the day due to Rayleigh scattering of sunlight by molecules in Earth's atmosphere.<|im_end|>\n"
[request   ] prompt_len=2165 tail_text="<think>\n\n</think>\n\nThe sky appears blue during the day due to Rayleigh scattering of sunlight by molecules in Earth's atmosphere.<|im_end|>\n<|im_start|>user\nAnd grass?<|im_end|>\n<|im_start|>assistant\n<think>\n"
```

The commit contains the real thinking ("don't require any tool usage..."). The next request from the same client contains an empty `<think>\n\n</think>\n\n` block. Same logical assistant turn, different tokens, no prefix match.

We then dumped hip's persisted conversation - the saved assistant message is `"\n\nThe sky appears blue..."`, with no reasoning. That confirms the client stored only `delta.content`, as expected per the OpenAI schema.

## Root cause (code reference)

```python
# mtplx/server/openai.py inside the streaming handler
def streamed_history_content() -> str:
    reasoning = (
        "".join(history_reasoning_chunks)
        .replace(THINK_OPEN, "")
        .replace(THINK_CLOSE, "")
        .strip()
    )
    content = (
        "".join(history_content_chunks)
        ...
    )
    pieces = []
    if reasoning:
        pieces.append(f"{THINK_OPEN}\n{reasoning}\n{THINK_CLOSE}")
    if content:
        pieces.append(content)
    return "\n\n".join(pieces)
```

The `if reasoning: pieces.append(...)` branch puts the streamed reasoning back into the assistant content used by the bank commit, even though no protocol contract obliges the client to send that reasoning back on the next turn.

## Fix

Commit only `history_content_chunks` - exactly what the client sees on `delta.content` and stores. tool_call markup is captured separately in `tool_stream`, and preamble text is in `history_content_chunks`, so the preamble cache-fix this function was originally added for is preserved.

PR: link to fix PR (to be filled in after PR is opened).

## After fix - same repro

```
turn 1 (cold):  prompt=2129 cached=   0 prefill=2129 ttft=2.49s restore=cold
turn 2:         prompt=2167 cached=2154 prefill=  13 ttft=0.11s restore=reference_lease
turn 3:         prompt=2206 cached=2192 prefill=  14 ttft=0.25s restore=reference_lease
turn 4:         prompt=2240 cached=2229 prefill=  11 ttft=0.13s restore=reference_lease
```

22x TTFT improvement on warm turns, ~99% prefill reduction. Verified stable across pure-text and tool-using flows.

## Related

Original sentinel-based postcommit prefix work landed in `d4315b7` ("Fix SessionBank postcommit prefix reachability"). That code path is correct - the issue is one layer earlier in how `assistant_content` was constructed for the commit, so the sentinel path was always rendering the wrong input.

Suggest also auditing the non-streaming response path (`_display_text` at line ~4065 / called at line ~6808) for the same problem - it returns `raw_text` (full generated text including thinking) when `state.args.normalize_thinking_tags` is off, which would also be passed to the bank commit and produce the same divergence.

## Environment

- macOS arm64
- Model: `mtplx-qwen36-27b-optimized-speed`
- Server flags: `--reasoning-parser qwen3 --no-stats-footer --profile sustained --generation-mode mtp --depth 3`
- Client: hippo-code (hip) 0.3.24

daniel-farina:
Fixes #54.

## Problem

`streamed_history_content()` always splices `history_reasoning_chunks` into the assistant_content used for the SessionBank commit. Clients receive content and reasoning as two SSE fields (`delta.content`, `delta.reasoning_content`); per the OpenAI schema they don't round-trip the reasoning field, so the next-turn request renders an empty `<think></think>` block instead of the real thinking. The committed token sequence is therefore not a prefix of the next request and every multi-turn cache lookup misses with `prefix_divergence_at_token`. Full diagnostic (token-level decode side-by-side, hip persisted conv dump, bank/admin output) is in #54.

## Fix

Commit only `history_content_chunks`. That matches what `delta.content` carried and what the client stores and echoes back. tool_call markup is in `tool_stream` and preamble text is in `history_content_chunks`, so the preamble-loss cache-fix this function was originally introduced for is preserved.

## Verification

Repro per #54 (4-turn `--resume` session against `mtplx-qwen36-27b-optimized-speed` with `--reasoning-parser qwen3`):

| turn | prompt_tok | cached_tok | prefill_tok | ttft | restore |
|------|-----------|-----------|------------|------|---------|
| 1 (cold) | 2129 | 0 | 2129 | 2.49s | cold |
| 2 | 2167 | 2154 | 13 | 0.11s | reference_lease |
| 3 | 2206 | 2192 | 14 | 0.25s | reference_lease |
| 4 | 2240 | 2229 | 11 | 0.13s | reference_lease |

22x TTFT speedup on warm turns, ~99% prefill reduction. Cache hits sustained across the conversation.

Also tested with `enable_thinking` toggled on/off and with tool-call turns; both paths now produce commit tokens that match the subsequent request prefix.

## Test plan

- [x] 4-turn pure-text `--resume` session warms after turn 1 (above)
- [x] code editor persisted conv shows only `content` field, confirming bug premise
- [x] Bank entries via `/admin/sessions` show `hits > 0` after warm turns (was 0 before)
- [ ] Suggest also auditing the non-streaming `_display_text` path for the same issue when `normalize_thinking_tags` is off - flagged in #54.

## Related

- d4315b7 added the sentinel-based postcommit prefix path. That path is correct - the issue was the assistant_content fed into it. PR keeps the sentinel path unchanged.

### DIFF
--- a/mtplx/server/openai.py
+++ b/mtplx/server/openai.py
@@ -6453,35 +6453,36 @@ def create_app(state: ServerState) -> FastAPI:
                     return chunks
 
                 def streamed_history_content() -> str:
-                    # Always capture the natural-language portion of the
-                    # response. Previously this returned "" whenever
-                    # tool_stream.has_tool_calls, which dropped any preamble
-                    # text (e.g. "Let me check..." before <tool_call>) from
-                    # the stored assistant_content. The next turn's lookup
-                    # encodes the same assistant message WITH the preamble
-                    # (clients echo back content + tool_calls), so the
-                    # prefix diverged and every tool-using turn paid a cold
-                    # prefill. tool_call markup itself is captured in
-                    # tool_stream and not in history_content_chunks, so
-                    # this is safe to return for the tool-call case too.
-                    reasoning = (
-                        "".join(history_reasoning_chunks)
-                        .replace(THINK_OPEN, "")
-                        .replace(THINK_CLOSE, "")
-                        .strip()
-                    )
-                    content = (
+                    # The bank commit's assistant_content must match what the
+                    # client actually received and will echo back on the next
+                    # turn, otherwise the committed tokens are not a prefix
+                    # of the next request and the cache always misses.
+                    #
+                    # With the reasoning parser active (qwen3 etc.), the SSE
+                    # stream is split into `delta.content` (what most clients
+                    # read into their stored assistant message) and a
+                    # separate `delta.reasoning_content` (a special field
+                    # that the OpenAI-style schema does not require clients
+                    # to round-trip). Clients like hip and opencode store
+                    # only `content`. Their next-turn assistant message
+                    # therefore re-renders with an EMPTY <think></think>
+                    # block. If we bake the real reasoning into the bank
+                    # commit, the commit tokens diverge from every
+                    # subsequent request and every long-lived session
+                    # cache-misses at the assistant boundary.
+                    #
+                    # Commit only `history_content_chunks` to match what the
+                    # client stored. Preamble text and tool_call preambles
+                    # are in history_content_chunks (tool_call markup is
+                    # captured separately in tool_stream), so this still
+                    # preserves the preamble-fix this function was added
+                    # to address.
+                    return (
                         "".join(history_content_chunks)
                         .replace(THINK_OPEN, "")
                         .replace(THINK_CLOSE, "")
                         .strip()
                     )
-                    pieces: list[str] = []
-                    if reasoning:
-                        pieces.append(f"{THINK_OPEN}\n{reasoning}\n{THINK_CLOSE}")
-                    if content:
-                        pieces.append(content)
-                    return "\n\n".join(pieces)
 
                 for field, text in splitter.start():
                     if text:


daniel-farina:
Superseded - upstream landed an equivalent fix in f1cbd97 ("Add production OpenCode integration") which drops history_reasoning_chunks from the SessionBank assistant_content commit. Same diagnosis, same fix.

Closing this PR. The original issue #54 documents the reproduction and diagnostic in case it's useful as historical context.

PR #83, opened 2026-05-24
fix: resolve SparseDistribution crash on zero/NaN probability mass

### LINKED ISSUE #82: ValueError: SparseDistribution probabilities must have positive mass during speculative drafting
### Description

When using speculative decoding / drafting (specifically with Qwen 27B Optimized Speed model), inference occasionally crashes with:
`ValueError: SparseDistribution probabilities must have positive mass`

This happens due to:
1. **Numerical instability or masking**: Extreme logits or masked positions in the model can result in all `-inf` values, which lead to `NaN` probabilities.
2. **Residual distribution logic**: In `residual_distribution`, the difference of probabilities can sum to exactly `0.0` or a non-finite value if a value is `NaN`. If `total` evaluates to `NaN`, the check `total <= 0` evaluates to `False`, bypassing target fallback logic and constructing a `SparseDistribution` with all `NaN` probabilities, which crashes the inference runner.

### Suggested Fix

1. Fallback gracefully to a one-hot distribution on the first available token (greedy argmax) inside the `SparseDistribution` constructor if the probability mass is zero or non-finite.
2. Add strict `not np.isfinite(total)` checks to the fallback paths in `residual_distribution` to prevent `NaN` values from bypassing target fallbacks.

I have implemented this fix and verified it with unit tests. I will submit a PR shortly.

Jonathangadeaharder:
Resolves #82 

### Changes
1. **Fallback in `SparseDistribution` Constructor**: Made `SparseDistribution.__post_init__` fall back to a valid one-hot distribution on the first available token (greedy choice) if the sum of the probabilities is not finite or `<= 0`, rather than raising a crash-inducing `ValueError`.
2. **Robust `residual_distribution`**: Added `not np.isfinite(total)` checks to the fallback paths in `residual_distribution` (sparse, dense, and non-sparse branches) to prevent `NaN` values from bypassing target fallbacks.
3. **Unit Tests**: Added test cases in `tests/test_sampling.py` verifying fallback recovery on zero/NaN probability mass, as well as a mock test for `residual_distribution` NaN robustness.

All tests pass successfully.

### DIFF
--- a/mtplx/sampling.py
+++ b/mtplx/sampling.py
@@ -17,7 +17,6 @@ class SamplerConfig:
     top_p: float = 0.95
     top_k: int = 20
 
-
 @dataclass(frozen=True)
 class SparseDistribution:
     token_ids: np.ndarray
@@ -37,7 +36,12 @@ class SparseDistribution:
             raise ValueError("SparseDistribution probabilities must be non-negative")
         total = probs.sum()
         if not np.isfinite(total) or total <= 0:
-            raise ValueError("SparseDistribution probabilities must have positive mass")
+            if token_ids.size > 0:
+                token_ids = token_ids[:1]
+            else:
+                token_ids = np.array([0], dtype=np.int64)
+            probs = np.array([1.0], dtype=np.float64)
+            total = 1.0
         object.__setattr__(self, "token_ids", token_ids)
         object.__setattr__(self, "probs", probs / total)
 
@@ -158,7 +162,7 @@ def residual_distribution(target_p: Distribution, draft_q: Distribution) -> Dist
             )
             keep = residual > 0
             total = residual[keep].sum()
-            if total <= 0:
+            if total <= 0 or not np.isfinite(total):
                 return target_p
             return SparseDistribution(token_ids[keep], residual[keep] / total, _vocab_size(target_p))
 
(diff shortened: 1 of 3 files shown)


xlinbsd (approved):
Reproduced on v0.3.7 — manual patch from PR #83 works**

Environment: M5 Pro 64 GB, macOS 26.5, `Youssofal/Qwen3.6-27B-MTPLX-Optimized-Quality`, `--profile sustained`, `--reasoning off`.

After a few minutes of agentic sessions the server crashes with `ValueError: SparseDistribution probabilities must have positive mass`, and the client gets `Connection reset by peer (os error 54)`.

Applied the patch from PR #83 manually to `site-packages/mtplx/sampling.py` — two changes:

**1. `SparseDistribution.__post_init__` — fallback instead of raise:**
```python
# before
if not np.isfinite(total) or total <= 0:
    raise ValueError("SparseDistribution probabilities must have positive mass")

# after (mirrors fast_sampling.py which already has this fix)
if not np.isfinite(total) or total <= 0:
    token_ids = token_ids[:1] if token_ids.size > 0 else np.array([0], dtype=np.int64)
    probs = np.array([1.0], dtype=np.float64)
    total = 1.0
```

**2. `residual_distribution` — add NaN guard to all 4 `if total <= 0:` checks:**
```python
if total <= 0 or not np.isfinite(total):
```

Sessions are now stable. Note that `fast_sampling.py` already has the equivalent fallback at lines 98-100, 151-153 and 208-211 — `sampling.py` was just missing it.

Please merge PR #83.

PR #132, opened 2026-07-03
Bound the batch scheduler's finished-request tracking

SuperMarioYL:
## What

`MTPContinuousScheduler` accumulates every finished request in an unbounded `self.finished` list and deduplicates each finish with an `any(existing is request for existing in self.finished)` identity scan. For a long-running server this has two costs:

- **Unbounded memory** — the list holds a `RequestState` for every request the scheduler has ever completed; it is never trimmed (`_purge_terminal_queues` only prunes the prefill/decode/postcommit queues).
- **O(n²) dedup** — the linear identity scan runs on *every* terminal transition, so the aggregate cost over a session is quadratic in the number of requests.

`self.finished` is only ever observed through `len()` (the `snapshot()["finished"]` gauge), so nothing depends on it being a list of live objects.

## Change

- Key the finished set by `request_id` → dedup is now O(1).
- Cap the retained `RequestState` objects (`max_finished_retained`, default 4096) so live memory stays bounded on long sessions.
- Track the true cumulative count in a new monotonic `finished_total`.
- `snapshot()["finished"]` keeps its previous meaning (cumulative distinct finishes = `finished_total`); a new `finished_retained` field reports the bounded live size.

No public behaviour changes for callers that read the `finished` gauge, and the single-owner-thread invariant is unchanged (no new locking).

## Tests

Added to `tests/test_batching_foundation.py`:

- `test_finished_tracking_is_bounded_but_total_is_exact` — with retention capped at 8, finishing 50 requests keeps `len(scheduler.finished) <= 8` while `finished_total == 50`. This is **red on `main`**, where `len(scheduler.finished)` reaches 50 (the list is unbounded).
- `test_record_finished_is_idempotent_per_request` — recording the same request twice counts once.

Existing scheduler tests (`test_batching_foundation.py`, `test_model_scheduler.py`) still pass, along with the full CI suite; `ruff check`, `compileall`, and `scripts/hygiene_scan.sh` are clean locally.

### DIFF
--- a/mtplx/batching/scheduler.py
+++ b/mtplx/batching/scheduler.py
@@ -21,6 +21,12 @@ from .state import (
     preset_config,
 )
 
+# Upper bound on how many finished RequestState objects the scheduler keeps
+# around for deduplication. The monotonic ``finished_total`` counter tracks the
+# true count, so a long-running server does not accumulate every finished
+# request forever.
+_DEFAULT_MAX_FINISHED_RETAINED = 4096
+
 
 @dataclass(frozen=True)
 class BatchSchedulerConfig:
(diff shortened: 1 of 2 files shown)


youssofal:
Merged into main for 2.0.0 as commit 1392214, credited in the commit and release notes. The scheduler no longer tracks finished requests forever. Closing since main carries it; good catch and thanks.

PR #149, opened 2026-07-08
fix(reasoning): stop reasoning leaking into visible content on tag split

Osamaali313:
## Problem

The streaming reasoning splitter (`QwenThinkingContentStreamSplitter._drain` in `mtplx/reasoning_codecs.py`) leaks hidden reasoning into the **user-visible** content when a reasoning tag is split across an SSE chunk boundary.

The recognized aliases include tags up to `</reasoning>` (12 chars), but `_drain` only holds back:

```python
keep = max(len(QWEN_THINK_OPEN), len(QWEN_THINK_CLOSE)) - 1   # = 7
```

7 trailing bytes. When a longer alias tag (`<reasoning>`, `</reasoning>`, `<thinking>`, `<thought>`, …) straddles a chunk boundary, the tag fragment — and then the whole reasoning block plus its raw markup — is emitted as visible `content` instead of `reasoning_content`. The sibling `_drain_disabled` already computes a 16-byte window, showing the intended sizing; `_drain` was just not kept in sync.

Reproduction (fed as two SSE chunks):

```python
sp = QwenThinkingContentStreamSplitter(thinking_enabled=True)
out = sp.feed("R1</think>V1 <reasoni") + sp.feed("ng>SECRET</reasoning> V2") + sp.finish()
# visible content: "V1 <reasoning>SECRET</reasoning> V2"   <-- reasoning + markup LEAKED
```

## Fix

Widen the hold-back window to cover the longest reasoning tag, mirroring `_drain_disabled`:

```python
keep = max(
    max(len(name) for name in QWEN_STYLE_REASONING_TAG_NAMES) + len("</>"),
    16,
)
```

## Verification

Added `tests/test_reasoning_stream_split.py` (MLX-free, so it runs off-Apple — the existing splitter tests in `test_openai_bridge.py` import `mlx`). Before the fix it fails (`SECRET`/`<reasoning>` in visible content); after, it passes:

```
$ pytest tests/test_reasoning_stream_split.py
2 passed
```

I also ran a chunked-vs-whole boundary fuzz across `think`/`thinking`/`thought`/`reasoning` inputs: the fix reduces total discrepancies (e.g. 42 → 26 on tag-at-start inputs) and eliminates every reasoning-into-visible leak (0). Behavior is unchanged for valid, whole-message data.

### DIFF
--- a/mtplx/reasoning_codecs.py
+++ b/mtplx/reasoning_codecs.py
@@ -409,7 +409,14 @@ class QwenThinkingContentStreamSplitter(ReasoningContentStreamSplitter):
 
     def _drain(self, *, final: bool) -> list[tuple[str, str]]:
         chunks: list[tuple[str, str]] = []
-        keep = max(len(QWEN_THINK_OPEN), len(QWEN_THINK_CLOSE)) - 1
+        # Hold back enough trailing bytes to cover the longest reasoning tag
+        # (e.g. "</reasoning>"); using only the "<think>"/"</think>" lengths
+        # let a longer alias tag split across chunks leak into visible content.
+        # Mirrors _drain_disabled's window.
+        keep = max(
+            max(len(name) for name in QWEN_STYLE_REASONING_TAG_NAMES) + len("</>"),
+            16,
+        )
         while self._pending:
             if self._inside_thinking:
                 close_match = QWEN_STYLE_REASONING_CLOSE_RE.search(self._pending)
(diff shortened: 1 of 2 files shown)


youssofal:
Shipped in 2.0.2 — your fix is on main as a5edba5 with your authorship, and you're credited in the release notes (https://github.com/youssofal/MTPLX/releases/tag/v2.0.2). I verified the leak repro on our tree before applying and your fix after; the MLX-free tests came along as-is. Clean, surgical patch — thank you. Closing since the commit landed outside the PR merge button.

Osamaali313:
Thank you — much appreciated, and glad it's in 2.0.2! Great that you re-ran the leak repro on your tree before/after; that's exactly the right check for this one since the streaming path is easy to regress. Enjoyed digging into MTPLX — I'll keep an eye out for more.

PR #151, opened 2026-07-09
Context-copy (prompt-lookup) drafting in the MTP decode loop: +39-74% on public code-edit suites, parity elsewhere

lBroth:
Up front, full disclosure: I am not a professional ML engineer, this was built with heavy AI assistance, and my hardware is limited (M5 Pro, 48 GB). Every number below is a real measured run on that machine, the whole thing reproduces from the included scripts, and I would genuinely welcome scrutiny, especially on better hardware.

## What this adds

Context-copy drafting inside `generate_mtpk`. When the tail of the generated stream matches an n-gram from the prompt, the continuation block (8 to 24 tokens, length scaled by match confidence) is proposed verbatim and verified in one forward pass through the existing `capture_commit` path, so the MTP head is skipped for that cycle. When there is no match, the normal MTP round runs unchanged.

The idea composes with what MTPLX already does rather than replacing it. The MTP head is strong on novel tokens, but it is capped at `mtp_depth` drafted tokens per step. On grounded work (code edits, file re-emission, RAG style answers) most of the output already exists verbatim in the prompt, and a copy window can commit far more than 3 tokens per verify. Matching is restricted to the prompt only, which is the original prompt-lookup semantics: matches into the model's own generated text (self-repeating math or boilerplate) sit below verify break-even and are exactly the cases that used to cause regressions in earlier iterations of this branch.

It is always on. There is no opt-in flag, only an emergency kill switch (`MTPLX_CONTEXT_COPY=0`). Greedy and `capture_commit`/`graphbank_capture_commit` only for now. Safety nets: an acceptance EMA that suspends probing when it stops paying, with exponential backoff, plus a confidence ladder so weak matches only ever risk a short block.

## Benchmarks, master vs this branch

Qwen3.6-27B-MTPLX-Optimized-Speed, M5 Pro 48 GB, `mtplx tune`, D3, greedy, seed 0. Arms were run interleaved per suite (master then fork, back to back) because I found that sustained runs throttle P-cores and skew Python-heavy decode paths by about 5% while leaving AR untouched, so non-paired A/B tables lie on this machine.

| suite | n | master D3 tok/s | this branch | ratio | identical output |
|---|---|---|---|---|---|
| CanItEdit (public, code edits) | 25 | 36.9 | 64.3 | 1.74x | 19/25 |
| HumanEvalFix (public, bug fixes) | 25 | 38.0 | 52.8 | 1.39x | 25/25 |
| Spec-Bench sample (public, NLP mix) | 26 | 30.6 | 30.2 | 0.99x | 23/26 |
| Aider polyglot implement (public) | 10 | 33.4 | 33.7 | 1.01x | 5/10 |
| default (yours) | 6 | 29.8 | 29.8 | 1.00x | 5/6 |
| long_code @1024 (yours) | 1 | 33.8 | 33.8 | 1.00x | 1/1 |
| long_code_uncapped @1024 (yours) | 1 | 34.3 | 34.3 | 1.00x | 1/1 |
| calibration_coding (yours) | 24 | 31.3 | 31.5 | 1.01x | 19/24 |
| python_modules_long @1024 (yours) | 1 | 34.8 | 34.5 | 0.99x | 0/1 |
| flappy @1024 (yours) | 1 | 30.3 | 30.3 | 1.00x | 1/1 |
| AIME 2026 (math reasoning) | 30 | 33.4 | 32.5 | 0.97x | 20/30 |

In short: +39% to +74% on public code-editing suites, and parity (0.97x to 1.01x) everywhere else, including every suite that ships with this repo. AR is untouched in both arms (17.4 to 17.8 tok/s). One config for the whole table, nothing tuned per suite.

## Correctness

The verify commits only the target's own argmax tokens, the same bar as your greedy gate. On the code-edit suites the output is byte-identical to master in 44 of 50 rows, and I teacher-forced every single non-identical row across the whole table: each one diverges at a verified near-tie (top1 minus top2 logit gap below the 0.8 envelope I measured for M5 GEMM vs GEMV paths, several are exact 0.000 ties). These are the same coin flips plain decode shows against itself at different prefill chunkings. Your full pytest suite is green with the feature on and off.

## What I would love from reviewers

1. Independent validation on bigger hardware and longer contexts. My M5 Pro cannot reach M5 Max or 128k territory, and these numbers deserve checking by someone who can.
2. Whether you want a proper temperature path (sample-and-match over the block) in this PR or as a follow-up.
3. Where copy stats should live. Right now they are in the per-round events; I did not touch `GenerationStats`.

While benchmarking I also ran into two things unrelated to this change: the 4B speed-test checkpoints collapse to 0 draft acceptance on M5 Pro at every depth, and `tune` under GPU contention once persisted a winning depth whose measured acceptance was 0.0, which later collapse verdicts never cleared from the saved state. I have full evidence for both and I am happy to file them as separate issues if useful.

---

Update, 18 July: rebased on 2.1.0 (clean) after the cache release, full suite
green, and a quick interleaved A/B on the 27B confirms the copy path is
undisturbed: byte-identical output with copy on and off, 31.9 to 38.6 tok/s
decode on the re-emission probe. The copy stats now live in GenerationStats
as requested in review (01534a1), so ask 3 above is settled, and the two side
observations at the bottom are filed as #176 and #177. Temperature stays a
follow-up PR as agreed. Benchmarks in the table above were measured against
the pre-2.1.0 master; the mechanism is unchanged by the rebase.

### DIFF
--- a/mtplx/context_copy.py
+++ b/mtplx/context_copy.py
@@ -1,6 +1,6 @@
 """Context-copy (prompt-lookup) speculative drafting for the MTP decode loop.
 
-Opt-in via MTPLX_CONTEXT_COPY=1. When the tail of (prompt + generated) tokens matches an
+Always on (kill switch: MTPLX_CONTEXT_COPY=0). When the tail of (prompt + generated) tokens matches an
 earlier n-gram, the continuation block is proposed verbatim (uncapped, K up to
 MTPLX_CONTEXT_COPY_K) and verified in one forward through the existing capture_commit
 verify path — no MTP-head compute for that round. When there is no match, the normal MTP
@@ -15,7 +15,9 @@ import os
 
 
 def context_copy_enabled() -> bool:
-    return (os.environ.get("MTPLX_CONTEXT_COPY") or "").strip() in {"1", "true", "on"}
+    """Context-copy is part of the engine (always on). MTPLX_CONTEXT_COPY=0 is an
+    emergency kill switch only; there is no opt-in flag."""
+    return (os.environ.get("MTPLX_CONTEXT_COPY") or "").strip() not in {"0", "false", "off"}
 
 
 def context_copy_block_k() -> int:
@@ -41,13 +43,23 @@ def context_copy_ng_max() -> int:
 
 def context_copy_min_ext() -> int:
     """Minimum backward match extension (beyond ng_min) required to fire a copy round.
-    Short, ng_min-only matches are common in novel generation (loops, boilerplate) and
-    have low acceptance — those rounds should stay with the MTP head. Default 2 =
-    require an (ng_min+2)-token suffix match before copying."""
+    Default 0: weak matches are allowed but propose only a SHORT block (see
+    block_for_ext), so a wrong incidental match wastes little."""
     try:
-        return max(0, int(os.environ.get("MTPLX_CONTEXT_COPY_MINEXT") or 2))
+        return max(0, int(os.environ.get("MTPLX_CONTEXT_COPY_MINEXT") or 0))
     except ValueError:
-        return 2
+        return 0
+
+
+# Confidence ladder: block length by backward match extension (0..ng_max-ng_min).
+# A longer suffix match earns a longer copy block; weak matches stay cheap.
+# Same schedule validated in the standalone prompt-lookup work (K_LADDER).
+_BLOCK_LADDER = (8, 12, 16, 24, 32)
+
+
+def block_for_ext(ext: int, k_cap: int) -> int:
+    idx = max(0, min(int(ext), len(_BLOCK_LADDER) - 1))
+    return min(_BLOCK_LADDER[idx], max(4, k_cap))
 
 
 class NgramIndex:
--- a/mtplx/context_copy.py
+++ b/mtplx/context_copy.py
@@ -1,0 +1,96 @@
+"""Context-copy (prompt-lookup) speculative drafting for the MTP decode loop.
+
+Opt-in via MTPLX_CONTEXT_COPY=1. When the tail of (prompt + generated) tokens matches an
+earlier n-gram, the continuation block is proposed verbatim (uncapped, K up to
+MTPLX_CONTEXT_COPY_K) and verified in one forward through the existing capture_commit
+verify path — no MTP-head compute for that round. When there is no match, the normal MTP
+round runs unchanged. Greedy (temperature<=0) only for now.
+
+Rationale: MTP heads draft novel tokens well (~2x ceiling at depth<=3) but cannot open a
+long verbatim window; on grounded/agentic workloads (code edits, file re-emission, RAG)
+most output already exists in context, where block-copy verification reaches 4-8x.
+The two mechanisms compose: copy when a match exists, MTP otherwise.
+"""
+import os
+
+
+def context_copy_enabled() -> bool:
+    return (os.environ.get("MTPLX_CONTEXT_COPY") or "").strip() in {"1", "true", "on"}
+
+
+def context_copy_block_k() -> int:
+    try:
+        return max(4, int(os.environ.get("MTPLX_CONTEXT_COPY_K") or 24))
+    except ValueError:
+        return 24
+
+
+def context_copy_ng_min() -> int:
+    try:
+        return max(2, int(os.environ.get("MTPLX_CONTEXT_COPY_NGMIN") or 6))
+    except ValueError:
+        return 6
+
+
+def context_copy_ng_max() -> int:
+    try:
+        return max(context_copy_ng_min(), int(os.environ.get("MTPLX_CONTEXT_COPY_NGMAX") or 10))
+    except ValueError:
+        return 10
+
+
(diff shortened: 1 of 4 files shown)


youssofal:
This is one of the strongest PRs this repo has received — measured, honest about limits, interleaved A/B discipline (your throttling note about non-paired tables is exactly right), and the idea composes with the MTP head instead of fighting it. Thank you.

Status and plan, so you're not guessing:

**It's not in today's 2.0.3** — that release is bug-focused (memory umbrella, agentic cache, app fixes). Context-copy is queued as the headline for the next perf release, and I don't want to rush its gauntlet: always-on decode-path changes get exactness gates (greedy byte-parity across our full suite matrix), pillar regression sweeps (decode/prefill/memory/agentic wall-clock), and long-context validation on M5 Max 128GB — which also answers your reviewer ask #1: I'll run your suites at 32k-128k on this hardware and publish the table in the PR.

Your three questions:
1. **Bigger hardware / longer contexts** — on me, as above.
2. **Temperature path** — follow-up PR, please. The greedy gate is what makes the current diff reviewable; sample-and-match needs its own correctness argument (probability-ratio acceptance over the copied block, not prefix-match) and I'd rather land the greedy win first.
3. **Copy stats** — promote them into `GenerationStats` (counters: probes, accepted blocks, accepted tokens, suspend/backoff state). They'll show up in `mtplx_stats` automatically and the dashboard can chart them; per-round events alone disappear in aggregation.

And yes — please file both side-issues separately: the 4B zero-acceptance collapse and `tune` persisting a 0.0-acceptance depth. The second one likely explains part of #160's small-model misery, so it's more valuable than you may think. Full evidence dumps welcome.

youssofal:
Status update: 2.1.0 went out today with the cache and memory work. This PR is still the plan for the next performance release, and the gauntlet run I promised is still owed. Nothing needed from you right now.

lBroth:
Thanks, honestly relieved it held up under a maintainer read.

I rebased the branch on 2.1.0 (clean, no conflicts) and re-ran the full
suite plus a quick interleaved A/B on the 27B, since the cache work and
the copy path share the decode loop. Outputs stay byte-identical with
copy on and off, and the re-emission probe still pays: 31.9 to 38.6
tok/s decode on my M5 Pro.

The stats you asked for are in (01534a1): probes, copy rounds, drafted
tokens, accepted blocks and tokens, suspension count, and the current
suspend/backoff state as gauges, plus a disabled reason for the
fail-closed capture path. One thing I found on the way: mtplx_stats
doesn't actually pick up new GenerationStats fields by itself, the
envelope filters through PUBLIC_MTPLX_STATS_KEYS, so I added the ten
keys there and extended the regression tests that pin that list. If you
want the public envelope kept tighter just say so and I'll trim it.

The two side issues are up: #176 for the 4B collapse and #177 for tune
persisting a 0.0-acceptance depth. Before filing I re-reproduced both on
a clean 2.1.0 install: the 4B still measures zero acceptance at every
depth, your own mtp1-greedy-gate passes exactness 6/6 there while
accepting 0 of 378 drafts, and the fresh collapse run once again left
the poisoned tuning record untouched. The save-path trace from current
main is in #177, and since you mentioned #160 I linked it from there
too.

On temperature, agreed. Greedy first, and sample-and-match will come as
its own PR with its own correctness argument.

And thanks for taking the 32k-128k runs, that range is exactly what my
48GB can't hold. If counterpart runs at the sizes I can reach would help
fill the table, say the word.

youssofal:
Merged and shipped in 2.2.0, on by default. This is the release headline and you are credited in the release commit.

What landed is your design plus one addition: a temperature path. A greedy-only fast path is against this project's core principle, since the product sampler is 0.6/0.95/20 and MTPLX's whole differentiation is exact speculative sampling at temperature. Your design made it a small step: the copy block is a point-mass proposal, so each copied token is accepted with the target's own shaped probability and a rejection samples the residual, through exactly the acceptance machinery the MTP verify path already uses. No new math, no greedy shortcut, and your suspend/backoff logic applies unchanged since it keys on acceptance fractions.

Validation on our hardware (M5 Max, 27B Speed, temp 0.6, depth 3 turbo):

- Your six tests pass unchanged, plus new temperature tests including a 480-run distributional check that holds the marginal and the copy-armed conditional to the model's own law, copy on vs copy off.
- Real tool-calling edit session: +53% on the edit-heavy turn (56 vs 37 tok/s), +25% on test-file generation, parity on novel-text turns. Consistent with your +39-74% on the public edit suites.

Known follow-ups I still owe this feature: the O(context) list rebuild per probe and the full-prompt index build at setup are both real costs at 100k+ contexts and are on the list for the next perf pass.

Your #176 and #177 reports both check out too and are fixed in the same release, separate replies on those threads. The report and PR quality across all three is the best this repo has received. Thank you.

lBroth:
Just saw 2.2.0. The temperature path is exactly where I hoped this would
go, and honestly you shipped it better than my own follow-up draft.
Thank you for the credit and for the write-up quality on all three
threads. The 100k-context probe costs you flagged are noted, happy to
help on that pass when it comes up.

PR #153, opened 2026-07-11
fix(openai): suppress tool preambles for Hermes streams

shiftedx:
## Summary
- add a Hermes-specific OpenAI stream policy that buffers ambiguous content while tools are active
- defer the content/tool decision until both incremental and completion-time native tool parsers have run
- discard buffered preamble text when a valid structured tool call is found
- release buffered content for genuine text-only responses
- preserve existing incremental behavior for non-Hermes clients

## Why
Hermes Agent renders streamed `delta.content` immediately. Qwen-family models can emit internal plan labels before a valid native tool envelope, so users see fragments such as `search` and tool argument values alongside the structured tool call. Hermes requests identified by `X-MTPLX-Client: hermes` now receive a clean tool-only assistant turn.

## Verification
- `38 passed`: `tests/test_tool_aware_stream_translator.py`
- `12 passed`: focused OpenAI streaming/tool parser tests
- integration coverage tests both the incremental XML parser and completion-time native extractor
- live replay of the captured 71-tool Hermes request produced 0 visible content characters, both `fact_store` calls, and `finish_reason=tool_calls`
- a fresh Hermes Agent E2E turn executed four tools with an empty assistant tool-call content field, then returned a clean final answer

### DIFF
--- a/mtplx/server/openai.py
+++ b/mtplx/server/openai.py
@@ -6472,11 +6472,13 @@ class _ToolAwareContentStreamTranslator:
         argument_chunk_chars: int,
         tokenizer: Any | None = None,
         repair_unclosed_complete: bool = True,
+        suppress_tool_call_preamble: bool = False,
     ) -> None:
         self._tools = tools
         self._argument_chunk_chars = max(1, int(argument_chunk_chars))
         self._tokenizer = tokenizer
         self._repair_unclosed_complete = bool(repair_unclosed_complete)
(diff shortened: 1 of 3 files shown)


youssofal:
Merged and shipped in 2.1.0. Your commit is on main as 4efed67 with your authorship intact (it went in through the curated release branch rather than the merge button, which is why GitHub does not show this PR as merged). Credited in the release notes. Thank you.

PR #157, opened 2026-07-12
fix(server): apply presence/frequency penalties in the ar_batch sampler

### LINKED ISSUE #156: Completing #120: ar_batch pump (_BatchedARJob) never passes token_counts, so presence/frequency penalties no-op at all temperatures
Follow-up to #120 (my penalty PR, merged in 2.0.0). I think #120 is working exactly as written for the paths it touched — this is about a path it didn't, so I'm opening it as a question in case the `ar_batch` behaviour is intentional.

**Summary.** Under `--scheduler-mode ar_batch`, presence/frequency penalties are silently dropped at every temperature. The serial and MTP paths honour them correctly; only the batched-AR pump doesn't. This matters because `mtplx start opencode` auto-selects `ar_batch`, so the flagship agentic lane is the one where penalties go dead.

**Where (v2.0.2):**
- `mtplx/server/openai.py:1868` `_make_sampler(self, job: _BatchedARJob)`. The temp≤0 branch (`:1871`) returns `argmax`, and the fused fast-path (`:1873`, guarded by `not (presence or frequency)`) only runs when *no* penalty is set — so neither drops a user penalty.
- The actual drop is `:1891-1892` `sample_one` → `_sample_from_logits(logprobs[0], job.sampler, rng)`, called **without any completion token counts**. `grep -c token_counts mtplx/server/openai.py` → `0` — the symbol is absent from the whole file, so the penalty term has nothing to subtract against and no-ops at all temperatures.
- `mtplx/commands/public.py:218-219` `OPENCODE_FAIR_BATCHING_DEFAULTS = {"scheduler_mode": "ar_batch", ...}`, applied by `_apply_opencode_fair_defaults` (`:10983`, called at `:12087`) unless the user passes `--scheduler-mode serial`.

**Repro (<5 min):**
1. `mtplx start opencode ...` (or `mtplx serve --scheduler-mode ar_batch`) with `--default-presence-penalty 1.5`.
2. Send a prompt that loops a dense model (the #102 case). The penalty has no effect; identical looping to `--default-presence-penalty 0`.
3. Same launch with `--scheduler-mode serial` → penalty takes effect. 200 OK in every case, no error/warning.

**Expected:** penalties apply under `ar_batch` as they do under serial/MTP.
**Actual:** penalties silently ignored under `ar_batch` (and therefore in the opencode quickstart lane).

**Workaround today:** launch with `--scheduler-mode serial` (penalties honoured, at the cost of the batched pump's concurrency).

**Why this bites the exact users #120 was for:** #120 exists to give loop-prone Qwen3.6 an anti-repetition lever for long agentic sessions (#102). A user who follows that remedy and uses the opencode quickstart gets `ar_batch` auto-selected and their penalty dropped — reintroducing the #102 loop in the flagship lane with a 200 and no signal.

**Suggested fix:** thread per-sequence completion `token_counts` into the `_BatchedARJob` pump and pass them to `_sample_from_logits` in `sample_one`, mirroring what `generate_ar`/`generate_mtpk` already do. Happy to send the PR if you agree the batched pump should honour penalties.

SuperMarioYL:
## What

Under `--scheduler-mode ar_batch`, presence/frequency penalties are silently dropped at **every** temperature. `_BatchedARGenerationService._make_sampler` calls `_sample_from_logits` without any completion `token_counts`, so the penalty term has nothing to subtract against and no-ops. `grep -c token_counts mtplx/server/openai.py` is `0` today — the serial and MTP paths thread it, the batched-AR pump doesn't.

This is the flagship agentic lane: `mtplx start opencode` auto-selects `ar_batch` (`OPENCODE_FAIR_BATCHING_DEFAULTS`), so the exact loop-prone workloads penalties exist for are the ones where they go dead. Fixes #156.

## Why the fix looks the way it does

- Compute `penalties_active` once and **stop short-circuiting** to the `temp<=0` argmax fast-path and the fused-sampler fast-path when a penalty is set. Without this, a *greedy* request with a penalty still hit the bare `mx.argmax` and dropped the penalty too — so the bug really was "at every temperature", not just in `sample_one`.
- In `sample_one`, thread `Counter(job.tokens)` into `_sample_from_logits`. `job.tokens` holds exactly the completion tokens emitted so far for that job — `emit_token` appends *after* the pump samples (`generator.next()` returns, then `emit_token`), so at decode step N the counter sees tokens `0..N-1`, the same prefix the serial path uses (`Counter(tokens)` in `generation.py`). Prompt/prefix tokens live in the separate `insert_all_tokens` list and are never counted.
- Requests with **no** penalty keep their prior byte-identical paths (greedy argmax; fused mlx sampler; explicit-seed numpy path), so nothing on the hot no-penalty path changes.

## Behaviour

| scheduler mode | before | after |
|---|---|---|
| `serial` / MTP | penalties honoured | unchanged |
| `ar_batch`, penalty set | **silently dropped** (all temps) | penalties applied |
| `ar_batch`, no penalty | fast path | unchanged (byte-identical) |

## Tests

`tests/test_server_openai.py` gains three tests that drive the real `_make_sampler` → `sample_one` → `_sample_from_logits` chain (no mock of the function under fix), using the existing `object.__new__(_BatchedARGenerationService)` idiom:

- `test_ar_batch_sampler_applies_penalty_in_hot_sampling_path` — `temp>0`, `top_k=1` collapses to a deterministic argmax; a seen token is demoted below its rival.
- `test_ar_batch_sampler_applies_penalty_in_greedy_path` — same for the `temp<=0` greedy branch.
- `test_ar_batch_sampler_penalty_free_paths_are_unchanged` — no-penalty greedy and hot paths still return the plain argmax.

The first two are red without the source change (they return the un-penalized token 0) and green with it. Full `ci.yml` test set passes locally (macOS, py3.11); `compileall` clean.

### DIFF
--- a/mtplx/server/openai.py
+++ b/mtplx/server/openai.py
@@ -1889,18 +1889,29 @@ class _BatchedARGenerationService:
                 top_k=int(getattr(job.sampler, "top_k", 0) or 0),
             )
         rng = np.random.default_rng(job.seed)
+        # Closure-local completion-token histogram. Each sampler is bound to one
+        # sequence and is invoked once per decode step, so counts is updated the
+        # instant a token is produced — before the NEXT sample is drawn. This
+        # avoids the one-step staleness of reading job.tokens at sample time:
+        # emit_token runs after generator.next() returns, but the sampler runs
+        # one step ahead inside next() (mlx-lm's one-step-lookahead pipeline),
+        # so job.tokens is missing the token generated immediately before —
+        # precisely the token a loop-breaking penalty most needs to see. Keeping
+        # the histogram in the closure makes the penalty see that token,
+        # matching the serial generate_ar path's Counter(tokens). Per-sampler
+        # scope also means no cross-request count leakage: mlx-lm keeps
+        # samplers index-aligned with uids across filter()/extend().
+        counts: Counter = Counter()
 
         def sample_one(logprobs: Any) -> Any:
-            # Presence/frequency penalties subtract against the completion tokens
-            # emitted so far for THIS job. ``job.tokens`` holds exactly those
-            # (emit_token appends after the pump samples), so Counter(job.tokens)
-            # mirrors the serial/MTP paths; without it the penalty has nothing to
-            # subtract against and silently no-ops (issue #156).
-            token_counts = Counter(job.tokens) if penalties_active else None
             token, _distribution = _sample_from_logits(
-                logprobs[0], job.sampler, rng, token_counts=token_counts
+                logprobs[0], job.sampler, rng,
+                token_counts=counts if penalties_active else None,
             )
-            return mx.array([int(token)])
+            token = int(token)
+            if penalties_active:
+                counts[token] += 1
+            return mx.array([token])
 
         return sample_one
 
--- a/mtplx/server/openai.py
+++ b/mtplx/server/openai.py
@@ -32,6 +32,7 @@ import time
 import urllib.parse
 import uuid
 import webbrowser
+from collections import Counter
 from concurrent.futures import Future
 from contextlib import asynccontextmanager, contextmanager, nullcontext, suppress
 from dataclasses import asdict, dataclass, is_dataclass
@@ -1868,12 +1869,13 @@ class _BatchedARGenerationService:
     def _make_sampler(self, job: _BatchedARJob) -> Callable[[Any], Any]:
         import mlx.core as mx
 
-        if float(job.sampler.temperature) <= 0:
-            return lambda logprobs: mx.argmax(logprobs, axis=-1)
-        if not job.seed_is_explicit and not (
+        penalties_active = bool(
             float(getattr(job.sampler, "presence_penalty", 0.0) or 0.0)
             or float(getattr(job.sampler, "frequency_penalty", 0.0) or 0.0)
-        ):
+        )
+        if float(job.sampler.temperature) <= 0 and not penalties_active:
+            return lambda logprobs: mx.argmax(logprobs, axis=-1)
+        if not job.seed_is_explicit and not penalties_active:
             # Fast path: mlx-lm's fused GPU sampler. The numpy fallback below
             # synchronizes the full logits row to the CPU for every decode
             # step of every active sequence, which serializes the whole batch
@@ -1889,7 +1891,15 @@ class _BatchedARGenerationService:
         rng = np.random.default_rng(job.seed)
 
         def sample_one(logprobs: Any) -> Any:
-            token, _distribution = _sample_from_logits(logprobs[0], job.sampler, rng)
+            # Presence/frequency penalties subtract against the completion tokens
+            # emitted so far for THIS job. ``job.tokens`` holds exactly those
+            # (emit_token appends after the pump samples), so Counter(job.tokens)
+            # mirrors the serial/MTP paths; without it the penalty has nothing to
+            # subtract against and silently no-ops (issue #156).
+            token_counts = Counter(job.tokens) if penalties_active else None
+            token, _distribution = _sample_from_logits(
+                logprobs[0], job.sampler, rng, token_counts=token_counts
+            )
             return mx.array([int(token)])
 
         return sample_one
--- a/tests/test_ar_batch_penalty_parity.py
+++ b/tests/test_ar_batch_penalty_parity.py
@@ -1,0 +1,170 @@
+"""End-to-end ar_batch penalty parity test (PR #157, mmmugh review).
+
+Drives the REAL ``mlx_lm.generate.BatchGenerator`` with the real pump
+append-after-``next()`` ordering and compares per-position output to the serial
+``generate_ar`` path. Covers the mmmugh repro: token 7 leads token 3 by
+exactly 1.0, so ``presence_penalty=1.5`` must flip the winner the instant 7
+has been emitted once.
+
+Why this test exists
+--------------------
+``tests/test_penalties.py`` (``_ar_batch_sample_once``) fabricates
+``job.tokens`` directly and calls the sampler closure in isolation. That
+verifies the penalty *math* but is blind to the *staleness* bug: the pump's
+``emit_token`` runs AFTER ``generator.next()`` returns, but the sampler runs
+one step ahead INSIDE ``next()`` (mlx-lm's one-step-lookahead pipeline). So
+``Counter(job.tokens)`` is missing the token generated immediately before —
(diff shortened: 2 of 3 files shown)


mmmugh:
Hi @SuperMarioYL — I independently hit #156 and ended up opening #164 before I saw this PR. Apologies for the duplicate work; yours was first. Our diffs agree exactly on the `penalties_active` refactor and on routing penalized greedy requests off the bare `mx.argmax` short-circuit — that leg is right.

I think there's a subtle issue with the counts source, though, and I'd rather raise it here than have it land quietly. **`job.tokens` is one token stale at sampler-call time.**

The comment in the diff says:

> `job.tokens` holds exactly those (emit_token appends after the pump samples)

That's true of the *pump* — `emit_token` does run after `generator.next()` returns. But the sampler runs one step **ahead**, inside `next()`. mlx-lm's `GenerationBatch._step()` is a one-step-lookahead pipeline (`mlx_lm/generate.py`, 0.31.3):

```python
self._current_tokens = self._next_tokens   # 1327  what was sampled LAST step
inputs = self._current_tokens              # 1329
...                                        # 1359  sampler runs -> produces t_k
self._next_tokens = sampled                # 1367  stashed for NEXT step
return inputs, self._current_logprobs      # 1378  returns t_(k-1), NOT t_k
```

So when the sampler is invoked for `t_k`, the pump has only appended `t_0 … t_(k-2)`. `Counter(job.tokens)` is missing the token generated immediately before — which is precisely the token a loop-breaking penalty most needs to see.

Net effect: penalties fire, but one position late, and an immediate repeat escapes unpenalized.

### Repro

Self-contained; drives the real `BatchGenerator` with the real pump ordering and compares to serial `generate_ar`. Token 7 leads token 3 by exactly 1.0, so `presence_penalty=1.5` must flip the winner the moment 7 has been emitted once:

```
serial generate_ar (correct)      : [7, 3, 7, 7, 7, 7]
ar_batch w/ Counter(job.tokens)   : [7, 7, 3, 3, 7, 7]   <- this PR
ar_batch w/ closure-local counter : [7, 3, 7, 7, 7, 7]
```

At position 1, serial demotes 7 (just emitted, penalty applies) and picks 3. With `Counter(job.tokens)`, `job.tokens` is still empty there, so **`7` repeats immediately** — the exact degenerate behavior `presence_penalty` is meant to suppress.

Worth noting this is invisible to a sampler-closure unit test: `_ar_batch_sample_once` fabricates `job.tokens` directly, so it asserts the penalty *math* (which is fine) but not the *staleness*. It only shows up on a round trip through `BatchGenerator` + the pump's append-after-`next()` ordering. I made that mistake first — my initial patch was byte-identical to yours in this respect, and it passed a closure-level test.

### Suggested change

Keep the counts in the closure and update them with the token the sampler itself just produced. Each sampler is bound to one sequence and sees every token it generates, so it can't lag — and it's O(1)/step instead of rebuilding the histogram each step:

```python
rng = np.random.default_rng(job.seed)
counts: Counter[int] = Counter()

def sample_one(logprobs: Any) -> Any:
    token, _distribution = _sample_from_logits(
        logprobs[0], job.sampler, rng,
        token_counts=counts if penalties_active else None,
    )
    token = int(token)
    if penalties_active:
        counts[token] += 1
    return mx.array([token])
```

(mlx-lm keeps `samplers` index-aligned with `uids` across `filter()`/`extend()`, so there's no cross-request bleed.)

#164 has this plus a test that drives the real `BatchGenerator` and asserts per-position equality with serial `generate_ar` — it fails against `Counter(job.tokens)`, which is what makes it worth having. Happy to close #164 in favor of this PR if you'd rather apply the change here; you got here first and I don't want to step on it. Either way, I mainly want the correct version to be what ships.

SuperMarioYL:
Thanks @mmmugh — great catch, and appreciate the detail. The staleness is real: `emit_token` runs after `generator.next()` returns, but the sampler runs one step ahead inside `next()`, so `Counter(job.tokens)` was missing the immediately-preceding token — exactly the one a loop-breaking penalty needs to see. Your closure-local counter is the right fix.

Applied in 452fffc:

- `mtplx/server/openai.py`, `_make_sampler`: the closure now keeps a `counts: Counter` and bumps `counts[token] += 1` with the token the sampler itself just produced, passing `counts` (not `Counter(job.tokens)`) as `token_counts`. Each sampler is bound to one sequence, so it cannot lag; O(1)/step; no cross-request count bleed (mlx-lm keeps samplers index-aligned with uids across `filter()`/`extend()`).
- Added `tests/test_ar_batch_penalty_parity.py`, which drives the real `BatchGenerator` with the pump's append-after-`next()` ordering (not a closure-level mock — you flagged that those are blind to the staleness, which is correct) and asserts per-position parity with serial `generate_ar` on your repro (token 7 leads 3 by 1.0, `presence_penalty=1.5`). It fails on `Counter(job.tokens)` — `At index 1 diff: 7 != 3`, the immediate repeat escaping the penalty — and passes with the closure counter: `[7, 3, 7, 7, 7, 7]`.

Happy to take any further feedback. Since the fix is here now, feel free to close #164 if you'd rather not carry the duplicate — either way, appreciate you raising it.

youssofal:
Closing as superseded by the fix that ships in 2.1.0, with credit: you converged on the same penalties_active gating shape we ended up with, and your instinct to keep the fast paths untouched was right.

The one functional difference: Counter(job.tokens) reads the pump's token list, which lags the sampler by a step or two because the pump appends asynchronously, so the penalty for the most recent tokens is missed at exactly the moment repetition starts (mmmugh's writeup on #164 covers the staleness in detail). The shipped fix keeps the count on a sampler-owned counter the closure updates synchronously.

You are credited in the release notes. Appreciated the PR, future ones welcome.

SuperMarioYL:
Thanks @mmmugh for the heads-up on #164 — no worries on the duplicate work. No hard feelings either way on which one lands.

Quick status on #157: the `no-mlx-smoke` failures were a self-inflicted regression from the previous commit, not the underlying fix. The staleness fix switched the ar_batch penalty histogram to a closure-local `Counter()` that started empty, which broke the two isolated unit tests — they fabricate `job.tokens` and call the sampler once, so the empty counter left the penalty with nothing to subtract and the seen token won unpenalized. Seeded the counter from `job.tokens` at sampler creation (kept the per-step increment, so the real-pipeline parity test stays green and the one-step-lookahead fix is preserved). Both unit tests + the parity test pass locally; CI is re-running.

Happy to proceed with #157, or to close it in favour of #164 if that is easier on your end — your call.

PR #171, opened 2026-07-17
Experimental Qwen dense-MLP gate/up row-traversal fusion (default-off)

davidtai:
## What this ships

Experimental exact-M row-traversal verify kernels for Qwen3.6 under one default-off flag (`--experimental-qwen-row-traversal`, env `MTPLX_QWEN_ROW_TRAVERSAL`, qlinear half separately gated by `MTPLX_QWEN_ROW_TRAVERSAL_QLINEAR`):

1. **Fused gate/up + SwiGLU MLP tile compiled at the true verify row count** (M2-M6). Removes the two materialized MLP intermediates and the zero-padding waste of the fixed-M4/M8 tiles.
2. **Exact-M2/M3/M5 QuantizedLinear routing.** The turbo NAX patch covers M4-M16; below it, stock MLX `qmv` re-reads the full weight matrix once per output row at the depth-1/2 verify tiles, and M5 pays the padded-M6 route. Eligible non-prefill 4-bit calls now compile at M in {2,3,5}, with measured per-shape floors (tiny-N projections and M2 x lm_head stay stock).
3. Load-time selfcheck lanes (`qwen_rt_gateup`, `qwen_rt_qmm_m2/m3/m5`) with the same disable-on-mismatch contract as the shipped NAX lanes.

### Definitive end-to-end: total-quiet interleaved pairing

Protocol: exclusive GPU window (serving booted out), the only CPU-heavy cohabitant SIGSTOPped, no process >50% CPU, arms alternated in fresh processes (A,B,A,B,A,B), 3 repeats, control drift <=1.34%, engagement verified per arm.

**16K context:**

| K | control tok/s | fused tok/s | delta | verify-fwd | tokens |
| ---: | ---: | ---: | ---: | ---: | :--- |
| K0 | 28.06 | 27.75 | -1.08% | +1% | identical |
| K1 | 41.27 | 47.15 | **+14.25%** | **-7.7%** | fork (disclosed) |
| K2 | 44.25 | 57.91 | **+30.86%** | **-21.9%** | fork (disclosed) |
| K3 | 63.72 | 63.40 | -0.51% | ~0 | identical |
| K4 | 56.76 | 62.01 | **+9.23%** | **-10.3%** | identical |

**8K context (the cleanest statement of the win):**

| K | control | fused | delta | verify-fwd | tokens |
| ---: | ---: | ---: | ---: | ---: | :--- |
| K1 | 45.15 | 48.75 | **+7.95%** | -8.2% | **bitwise identical** |
| K2 | 47.73 | 59.66 | **+25.00%** | -22.5% | **bitwise identical** |

K4\@16K and both 8K cells are parity-clean end-to-end wins: identical tokens, identical acceptance, wall speed from kernel time alone. K1/K2\@16K combine the kernel-real verify-forward drop with a favorable acceptance-trajectory effect on the forked text (see numerics); both components are reported separately. K0/K3 price the Python dispatch wrappers (~0.5-1%) at depths where no new kernel engages — the flag is for K1/K2/K4-shaped serving, not the current K3 default.

### Kernel level (queued lane, real shapes and weights)

Exact-M QMM vs stock `qmv`: +8 to +45% across the live census shapes (peaks at M5, e.g. lm_head 2705->2134 us). Fused MLP tile vs stock projections+swiglu: +17 to +43% at M3-M6. Full tables in `docs/specs/2026-07-17-qwen-experimental-row-traversal.md`.

## Numerics

- **Bit-parity with the turbo baseline at M4/M5/M6 by construction** (gate-tested with `mx.array_equal` on real shapes: exact-M5 == NAX padded-M6, fused tile M4/M6 == turbo vk_k pair + compiled swiglu, dmax 0.0). K3/K4/K5 token parity in every window is structural.
- The **only** divergent tiles are M2/M3 vs stock MLX `qmv` (different association class). fp64 ladder: fused error 0.8-3.2 BF16 ULP, at or below stock's own 1.6-4.0; flat across 1000x magnitude; zero non-finite values. Measured forks are near-tie swaps (16K fork: 0.25-logit margin, fused picks stock's runner-up); stock replay is bit-deterministic; BN2/BN4 layouts are bit-identical.
- **Acceptance ratchet** (why forked cells post outsized tok/s): draft disagreements concentrate at the target's near-ties, so target-side tie-flips convert rejections to acceptances ~half the time and rarely break agreements. Measured +7.4 to +9.3pt acceptance in forked cells, jump anti-correlated with base acceptance, 0.0pt in all parity cells. Forked-cell tok/s is disclosed as workload throughput; verify-forward time is the kernel-real metric.
- **Bit-parity option prototyped:** stock-order kernels reproducing `qmv_fast`'s exact association while amortizing the weight read across rows — proven bitwise vs `mx.quantized_matmul` at M2/M3 (36/36 gates), at +9-16% over stock (vs vk_k's +26-45%). Documented as the follow-up path to a fully bit-identical flag.

## Method note

Two measurement failure modes were caught and are now protocol: (1) sequential arm blocks on a noisy box drifted controls 4-7% between windows — all end-to-end claims here come from same-window interleaved arms under the quiet protocol; (2) an in-process arm-toggle harness produced a convincing all-flat null because the fused path silently never engaged — every arm now carries an engagement signature (dispatch counters + the mechanical verify-forward drop).

Default off, no default-profile changes proposed.

https://claude.ai/code/session_01Gca2cYpuyu1MC5UKKMs9dH

### DIFF
--- a/docs/specs/2026-07-17-qwen-experimental-row-traversal.md
+++ b/docs/specs/2026-07-17-qwen-experimental-row-traversal.md
@@ -1,4 +1,4 @@
-# Experimental Qwen Gate/Up Row-Traversal Fusion
+# Experimental Qwen Row Traversal: Exact-M Verify Tiles
 
 ## Status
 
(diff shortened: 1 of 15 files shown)


davidtai:
I forget this is memory bound and only helps qwen 35b.  I need to discomboulate my non streaming MOE branch.  I'm also rebenching this as the speed up is coincidental to a concurrent compression task that was taking up a bunch of cpu.

davidtai:
Closing this. Measurement says the premise doesn't hold: **dense Qwen3.6-27B decode is memory-bandwidth saturated, so compute-side kernel improvements cannot move end-to-end throughput.**

The decomposition, per forward at 16K on an M5 Max (~572 GB/s measured bus ceiling):

| component | GB/forward | achieved GB/s |
|---|---|---|
| MLP (64 layers) | 9.63 | **570 — at the bus** |
| GDN linear-attn (48) | 3.13 + 0.30 state | 440 |
| full attention (16) | 0.94 + 1.07 KV | 429 |
| lm_head | 0.72 | 486 |
| **full forward** | **15.8** | **457 avg (80-91% of practical bus)** |

The MLP — the target of this PR — is already running at the memory bus. There is no headroom for a faster kernel to claim.

That is not a theoretical objection; it is what we measured on this branch. The row-traversal/fused MLP path is **+30-42% faster at the kernel level in isolation** and moved end-to-end decode by **~0%**: a quiet interleaved grid at K0-K4 / 16K came back flat within ±0.3% (control 28.47/41.29/44.18/63.86/56.86 vs candidate 28.43/41.33/44.14/63.54/56.97). Earlier +11-19% readings from this branch were measurement artifacts — controls depressed by background load plus a speculative-decode acceptance shift — and were retracted.

The exact-M tile work here is real and correct (bit-identical, 18/18 token-hash-identical), but it buys back padding waste in a component that is already bandwidth-limited, so the win stays confined to microbenchmarks.

Attempts to extend the approach also hit a wall: exact-M tiles beyond M6 do not work in this codegen family — M7/M8 run but lose (M8 pathological, lm_head 27.6 ms/call), and M9-M12 fail the load-time kernel selfcheck outright (dmax 1.2 to 2.4e24) and auto-disable.

**Where the work went instead:** #174, targeting Qwen MoE (35B-A3B), which sits at the *opposite* end of the roofline — ~1.4 GB of active weights per forward and only ~26% of bus, i.e. dispatch- and overhead-bound rather than bandwidth-bound. That is where reducing kernel launches and fixing configuration defaults actually pays.

PR #172, opened 2026-07-18
Treat mtp_depth_max as a ceiling, not the default MTP depth

davidtai:
## Problem

`_model_contract_depth` returns an artifact's **`mtp_depth_max`** as the **default** speculative depth. That field is a *ceiling* — the deepest MTP sidecar the artifact ships — so every model runs at its maximum depth regardless of whether that depth is its fastest mode.

For **Qwen3.6-35B-A3B** this is a measured regression. Published depth sweep (greedy):

| mode | tok/s | acceptance |
|---|---|---|
| AR baseline | 94.46 | — |
| **D1** | **138.39** | 0.8858 |
| D2 | 135.66 | 0.8701, 0.6409 |
| D3 | 107.67 | 0.8291, 0.5414, 0.2783 |

Per-level acceptance decays steeply on this MoE (0.829 / 0.541 / 0.278), so verify cost outruns the extra accepted tokens past D1. The artifact declares `mtp_depth_max: 3`, so MTPLX selects D3 — **−22.2% against D1, out of the box.**

## Repro

`benchmarks/repro_a3b_depth_default.py`, against a local A3B artifact:

```
$ PYTHONPATH=. python benchmarks/repro_a3b_depth_default.py
contract mtp_depth_max   : 3   (a CEILING)
contract recommended     : sustained
resolved default depth   : 3

STUCK AT 3: the ceiling is returned as the default depth.
Authors' data: D3 107.67 vs D1 138.39  ->  -22.2% throughput out of the box.
```

It passes `fallback=1`, so the `3` can only originate from the contract ceiling — not from a caller default. After this change the same script reports `resolved default depth : 1`.

Note the contract is profile-scoped: A3B declares `recommended_profile: sustained`, and the pin only reproduces under the matching profile.

## Fix

Separate the two concepts that `mtp_depth_max` was conflating:

1. Artifacts may declare **`mtp_depth_default`** in their runtime contract; `mtp_depth_max` then only *bounds* it.
2. `_MODEL_CONTRACT_DEPTH_DEFAULTS` carries measured defaults for already-published artifacts that predate the field — same measured-win-only shape as the existing `_TURBO_DEFAULT_PUBLIC_MODEL_IDS` list.
3. Artifacts declaring neither keep the previous ceiling behaviour, so nothing else changes.

Explicit `--depth` / `--speculative-depth` still wins, unchanged.

## Tests

Four new cases in `tests/test_public_cli.py`: ceiling-only behaviour preserved, declared default honoured, declared default clamped to the ceiling, and A3B resolving to its measured D1.

`tests/test_public_cli.py`, `tests/test_artifacts.py`, `tests/test_forge_cli.py` all pass (exit 0).

### DIFF
--- a/mtplx/commands/public.py
+++ b/mtplx/commands/public.py
@@ -795,6 +795,22 @@ def _model_draft_sampler_spec(
         return fallback
 
 
+# Artifacts whose fastest published mode is shallower than their sidecar
+# ceiling. Same shape as _TURBO_DEFAULT_PUBLIC_MODEL_IDS below: measured-win
+# only, keyed by public model id. Artifacts that declare ``mtp_depth_default``
+# in their runtime contract take precedence over this table.
+#
+# Qwen3.6-35B-A3B (published depth sweep, greedy):
+#     AR 94.46 | D1 138.39 | D2 135.66 | D3 107.67 tok/s
+# Per-level acceptance decays steeply on this MoE (D3: 0.829 / 0.541 / 0.278),
+# so verify cost outruns the extra accepted tokens past D1. Running the D3
+# ceiling as the default costs -22.2% against D1.
+_MODEL_CONTRACT_DEPTH_DEFAULTS: dict[str, int] = {
+    QWEN36_35B_OPTIMIZED_SPEED_PUBLIC_MODEL_ID: 1,
+    QWEN36_35B_OPTIMIZED_BALANCE_PUBLIC_MODEL_ID: 1,
+}
+
+
 def _model_contract_depth(
     inspection: dict[str, Any],
     *,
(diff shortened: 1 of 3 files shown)


mtplx/commands/public.py
```diff
@@ -805,9 +821,23 @@ def _model_contract_depth(
...
+    # a measured loss whenever per-level acceptance decays quickly (Qwen3.6-35B-A3B:
+    # published D1 138.39 tok/s vs D3 107.67 tok/s, so the ceiling costs -22%).
+    measured_default = _MODEL_CONTRACT_DEPTH_DEFAULTS.get(
+        str(contract.get("public_model_id") or "").strip()
+    )
+    try:
+        depth = int(contract.get("mtp_depth_default", measured_default or depth_max))
+    except (TypeError, ValueError):
+        depth = measured_default or depth_max
+    depth = min(depth, depth_max)
```
davidtai, line 840: Do what you will with this, its just opus being stupid/cute with min()

davidtai:
Superseded by #174, which carries this commit plus the MoE gate/up packing change on a single branch.

PR #173, opened 2026-07-18
Pack Qwen MoE gate/up projections at load time (default off)

davidtai:
## What

Concatenates the gate and up projections of Qwen's sparse MoE block into a single projection once at load time, so each half of the block runs one matmul instead of two:

| | before | after |
|---|---|---|
| routed experts (`SwitchGLU`) | `gather_qmm(gate)`, `gather_qmm(up)`, `gather_qmm(down)` | `gather_qmm(gate_up)`, `gather_qmm(down)` |
| shared expert (`Qwen3NextMLP`) | `qmm(gate)`, `qmm(up)`, `qmm(down)` | `qmm(gate_up)`, `qmm(down)` |

Both pairs read the same activation and differ only in which output rows of a weight matrix they consume, so the two matrices can be stacked along their output-feature axis and the fused result split in two.

Default off behind `MTPLX_QWEN_MOE_PACK_GATE_UP=1`.

## Why this is exact

Affine quantization groups run along the **input** axis, so concatenating output rows keeps every group intact. Nothing is requantized, and each output element is still produced by exactly the dot product that produced it before.

`tests/test_moe_packed_projections.py` asserts that bitwise (`mx.array_equal`, not a tolerance) on the CPU backend for:

- the plain `quantized_matmul` form and the gathered `gather_qmm` form, quantized and unquantized;
- `SwitchGLU` both below and above its `indices.size >= 64` token-sorting threshold;
- the shared expert MLP;
- the whole MoE block before vs. after packing, including blocks nested in a layer list.

Guard rails are tested too: a pair with mismatched output widths, mismatched quantization, or an additive bias is skipped with a reason recorded in the stats rather than mispacked.

## Why it should help

This targets **dispatch count**, not arithmetic or bytes. A 35B-A3B decode forward touches roughly 1.4 GB of active weights but issues several hundred kernel launches to do it, which puts it far below the memory-bandwidth roofline and makes per-launch and Python graph-construction cost the dominant term. Packing removes two matmul dispatches — and the graph nodes that build them — from every MoE layer, on the trunk and the MTP draft block alike. The arithmetic is unchanged.

## Compatibility

The packed forward is ordinary MLX ops, **not** a custom Metal kernel, so unlike a `mx.fast.metal_kernel` lane it still composes with `mx.compile` and the graphbank verify paths.

Because packing changes the *type* of the two projections it replaces, two optional lanes stop seeing them (both documented in the module docstring, both default-off, and neither touches the routed `down_proj` or the router):

- the packed shared expert is no longer a `Qwen3NextMLP`, so the `MTPLX_MLP_CALL_VARIANT` lane in `native_mlp` skips it — that lane targets the dense-model MLP and is inactive at M=1 anyway;
- the packed projections call `mx.quantized_matmul`/`mx.gather_qmm` directly rather than through `nn.QuantizedLinear`, so the NAX verify patch does not route them.

## Status

Correctness-gated, **performance pending**. The parity tests above run on the CPU backend and pass; I have not yet measured decode throughput on a Qwen MoE artifact, so there is no tok/s claim here. Follow-ups I'd want before this comes out of draft:

1. GPU parity on a real Qwen3.6-35B-A3B artifact (the group-preservation argument is backend-independent, but Metal parity should be confirmed rather than assumed);
2. an A/B decode measurement to size the dispatch saving;
3. optionally, packing the router `gate` with `shared_expert_gate` — same mechanism, one more dispatch per layer — which I left out to keep this reviewable.

## Test

```
$ python -m pytest tests/test_moe_packed_projections.py tests/test_attention_split.py -q
.............................                                            [100%]
$ echo $?
0
```

https://claude.ai/code/session_01YHUiWXTFYuksrafoFWoPbB

### DIFF
--- a/mtplx/moe_packed_projections.py
+++ b/mtplx/moe_packed_projections.py
@@ -1,0 +1,361 @@
+"""Construction-time gate/up projection packing for Qwen MoE blocks.
+
+Qwen's sparse MoE block runs the gate and up projections as two separate
+matmuls in both halves of the block::
+
+    routed experts (SwitchGLU): gather_qmm(gate), gather_qmm(up), gather_qmm(down)
+    shared expert  (MLP):       qmm(gate),        qmm(up),        qmm(down)
+
+Each pair reads the same activation and differs only in which output rows of
+a weight matrix it consumes, so the two matrices can be concatenated along
+their output-feature axis once at load time and evaluated as a single matmul
+whose result is split in two.
+
+Affine quantization groups run along the *input* axis, so concatenating
+output rows leaves every group intact: no requantization happens, and each
+output element is still produced by exactly the dot product that produced it
+before.  ``tests/test_moe_packed_projections.py`` asserts that bitwise on the
+CPU backend for both the plain and the gathered (routed-expert) form.
+
+The win here is dispatch count, not arithmetic.  A 35B-A3B decode forward
+touches roughly 1.4 GB of active weights but issues several hundred kernel
+launches to do it, which leaves it far below the memory-bandwidth roofline
+and makes per-launch and Python graph-construction cost the dominant term.
+Packing removes two matmul dispatches -- and the graph nodes that build them
+-- from every MoE layer, on both the trunk and the MTP draft block.
+
+Nothing in this module is a custom Metal kernel.  The packed forward is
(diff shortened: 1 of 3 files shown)


davidtai:
Superseded by #174, which carries this commit plus the mtp_depth_max ceiling fix on a single branch.

PR #174, opened 2026-07-18
Qwen3.6-35B-A3B decode: 206.8 tok/s via six kernel/schedule improvements

davidtai:
## Qwen3.6-35B-A3B decode: 199 tok/s via six kernel/schedule improvements

Resident-model single-stream decode (no expert streaming), Qwen3.6-35B-A3B, M5 Max. Every improvement is a construction-installed unit that removes host/dispatch overhead without changing arithmetic — or, where a fused kernel changes floating-point accumulation order, is proven not to change task output.

## Problem

Qwen3.6-35B-A3B is a sparse mixture-of-experts model: of its 35B parameters, only ~3B are active per token (top-8 of 256 experts plus a shared expert). On paper that makes it a fast decoder — each token touches a small fraction of the weights. In practice, resident single-stream decode left most of that potential on the table.

The reason is that at decode batch sizes (1–2 rows), A3B's forward pass is **overhead-bound, not compute-bound**. A single token dispatches hundreds of small Metal kernels — one MoE forward alone fans out across routing, per-expert gate/up projections, SwiGLU, and the score-weighted combine, each a separate launch that materializes an intermediate tensor. The 30 Gated Delta Net (linear-attention) layers add their own conv/gating/state-capture dispatches per token. On top of the model forward, the speculative-decode loop pays per-cycle host cost: it rebuilds the verify graph, mutates the KV cache on the host, and synchronizes host↔device several times per accepted token. None of this is arithmetic — it is launch latency, intermediate materialization, and host/device round-trips, and at these tiny row counts it dominates wall-clock.

The goal was **200 tok/s** on the resident model, on the `long_code` natural-stop K1 contract. The hard constraint: get there by **removing overhead, not by changing math**. Every improvement keeps arithmetic byte-identical to the stock path, or — where a fused kernel legitimately changes accumulation order — is proven not to change task output. That constraint is what lets these six kernels be adopted with confidence rather than as a speed-vs-correctness trade.

## Headline

**199 tok/s greedy** on the full stack (long_code), byte-identical to `generate_ar` of the same configuration — **1.67× over the autoregressive baseline** (119.5 tok/s), 1.54× over stock MTP speculative decode.

On a **realistic 133-token coding prompt**, the same k=1 stack reaches **209.9 tok/s greedy** — over 200 with no depth-2 — because the more predictable prompt drafts at higher acceptance (0.86 vs long_code's ~0.80). Throughput is acceptance-dependent, and on real inputs it clears 200.

## End-to-end ladder

One consistent measurement context, byte-gated greedy, on the final tree (`variant-exact-gates-20260723-111730`):

| Stage | tok/s | vs prev |
|---|---|---|
| Autoregressive baseline (stock arithmetic) | 119.5 | — |
| + MTP speculative decode | 129.1 | +8.0% |
| + Compiled K1 route + on-device draft + fused MoE/GDN kernels | 188.7 | +46.2% |
| + Whole-MoE fusion | **199.0** | +5.5% |

## The kernels

Three main kernels carry a measured throughput contribution and are numbered by it. Three supporting kernels reduce MoE dispatch overhead but have no clean isolated delta — they are present in every measured configuration (including the baseline), so they do not appear as a separate ladder step; they are described after the main three.

> **On the per-kernel numbers below.** These A/B deltas were measured during the 2.1.0 → 2.3.0 development window, each against the control stack *of that time*. Every improvement shifts the baseline the next one is measured against, and later changes compound, so these isolated figures are **historical** — they show each kernel was independently promotion-positive when it landed, not its contribution to the *current* stack. The authoritative present-day number is the end-to-end **199 tok/s** ladder above, measured in one consistent context on the final tree. Treat the per-kernel deltas as provenance, not as addends to 199.

### 1. Compiled K1 target-prefix route + on-device draft (R1) — +46.2%

**What it does.** This is Qwen3.6-35B-A3B's K1 (single-draft) speculative decode path: the on-checkpoint MTP head proposes one draft token per cycle, the target model verifies it, and a rejection is repaired in place. `--compiled-target-prefix` takes that whole verify-and-repair schedule and compiles it into two fixed Metal graphs (`mtplx/a3b_compiled_target_prefix.py`) instead of re-tracing it every cycle. It is the single biggest lever in the stack: end-to-end throughput goes from 129.1 to 188.7 tok/s with the route engaged on an otherwise identical lane (+46.2%, `variant-exact-gates-20260723-111730`, cells `lane-base` → `k1-nwm-greedy`, identical output SHA `64a8d8704fd317db`, exactness gates green).

**How it works.** `install_a3b_k1_target_prefix_route` builds one `mx.compile`d two-row verify graph (M2, `[primary, draft]` → logits/hidden plus 30 GDN capture pairs and 10 full-attention K/V/offset states) over a request-owned "shadow" cache of fixed `TensorOffsetKVCache`/`ArraysCache` slots (`A3BK1TargetPrefixRoute.state_slots`). The graph is built once per runtime/hidden-variant and shared across requests (`_shared_m2_step`/`_shared_m1_step`); the hot path never retraces, never mutates a Python cache object, and never rescans shape/dtype/eligibility — request construction proves all of that once, at install time. On rejection the route does not re-run M2 with `[primary, correction]`. Row 0 of the already-computed M2 output is the exact state after `primary` alone (the `primary_state` slice returned by `_forward_m2`), so the correction continues directly from that captured state through a second fixed one-row graph, M1 — no `snapshot_untrimmable_cache`, no `rollback_after_verify`, no second full verify forward.

**What it optimizes away.** The spec's measurement is the reason this route exists: on the native repeat-2 trace, 282 rejections over 1,595 cycles cost 3.016895 s of repair, or **10.698209 ms per rejection / 1.891470 ms per cycle** — "the largest remaining independently measured avoidable region." At decode, the model forward itself is cheap; what dominates the cycle is host scaffolding around it — per-cycle graph construction, cache-object bookkeeping/validation, and host↔device sync barriers. Compiling the schedule into fixed graphs over pre-validated, request-owned state deletes that scaffolding rather than shrinking the forward pass, which is why the win shows up end-to-end instead of in kernel time.

**On-device draft (R1).** Previously the draft token had to reach the host as a Python int before the `[primary, draft]` verify tensor could be built — a second host↔device round trip per cycle on top of the accept/reject decision. R1 samples the draft directly from device-resident logits (`sample_token_ids_from_mlx_logits`) and concatenates it with `primary` on-device (`mx.concatenate`) to build the verify input, only pulling the draft back to a Python int afterward for bookkeeping. Each cycle now exposes exactly one host↔device decision sync instead of two.

**Its data.** R1 standalone, isolated: candidate at 164.39 tok/s / 10.446 ms-cycle / 0.8353 acceptance vs. control at 161.64 tok/s / 10.564 ms-cycle / 0.8289 acceptance — **+1.70%** tok/s, −1.12% cycle time, 3/3 paired repeats improving, acceptance-parity gate passed at 0.0064 delta against a 0.01 limit (`disint-stack-natural-k1-tgy4-cxcxcx-20260720`). The full route bundle is the ladder's biggest jump: `lane-base` 129.1 → `k1-nwm-greedy` 188.7 tok/s, **+46.2%**, bit-exact against the AR reference.

### 2. Whole-MoE fusion — +5.5%

**What it does.** Replaces the fragmented sparse-MoE path used by target M1 (correction/draft), target M2 (K1 verification), and MTP M1 with one route of the minimum architecturally sound number of Metal dispatches — 3 stages instead of the stock chain's 12 source-level device boundaries (router → softmax → top-k → routed gate/up → SwiGLU → routed down → weighted reduction → shared gate/up → shared SwiGLU → shared down → scalar gate/sigmoid → final add). It changes none of the declared arithmetic: router, quantization (target q4/group64 routed, q8/group64 router and shared scalar gate), SwiGLU order, BF16 rounding boundaries, or reduction order are all preserved exactly. Only launch and materialization structure changes.

**How it works.** Stage 1 is one 256-thread, row-owned threadgroup that computes the router projection, precise FP32-internal softmax, exact top-8 selection/ties, BF16 score normalization, and the shared scalar-gate projection, emitting `[M,8]` indices/scores and `[M,1]` shared gate. For the 2-row (M2) verify case, Stage 1 splits into a tiled router-projection kernel (8 threadgroups, each owning 32 of the 256 experts and both rows via one packed-weight load) followed by a single finalizer threadgroup that does softmax/top-k/normalization for both rows sequentially — this split is the difference between the initial "row-owned" M2 cut and the later "tiled Stage 1" one. Stage 2 computes the 8 selected routed gate/up projections plus the shared gate/up projection and applies SwiGLU, tiled at 4 output columns per SIMDgroup × 4 SIMDgroups (16 cols/threadgroup): 288 threadgroups for 9 activation slots (8 routed + shared) at width 512, storing only compact BF16 `[M,9,512]`. Stage 3 is output-column-owned (128 threadgroups over hidden width 2048), computes all 9 down projections, applies route scores and the sigmoid shared gate, and writes the final BF16 `[M,2048]` directly — no separate routed/shared/add dispatch. The M1 (single-row) variant reuses row-0 arithmetic that is textually identical to the M2 kernel's row-0 path (`_target_m2r1_stage3_source` literally shares `_target_m2_routed_row_source(0)`), so decode-time M1 and verify-time M2 are bit-identical per row by construction — required for the acceptance-rate exactness gate.

**What it optimizes away.** At M=1-2 rows, a 256-expert top-8 sparse MoE is launch-bound, not compute-bound: each of the 12 stock boundaries is a full dispatch even though the actual per-row work (57.7–115.4 MFLOPs, ~15.7 MiB weight traffic) is tiny relative to fixed per-dispatch overhead. Fusing to 3 stages removes 9 launches per block across 40 target blocks plus 1 MTP block, and eliminates materializing routed `[M,8,2048]`, separate shared `[M,2048]`, and combined `[M,2048]` intermediates entirely — only `[M,9,512]` (9–18 KiB) crosses the Stage2→Stage3 boundary.

**Precision.** The fused Stage-3 reduction accumulates routed experts sequentially and rounds at different BF16 boundaries than stock `gather_qmm`, so it is tolerance- not bit-exact: route ids/scores match exactly, activations diverge by dmax ~0.0013. This is why the PR doesn't gate on trajectory-SHA match but on a one-sided acceptance-rate bound plus HumanEval, to confirm the ulp-level drift doesn't change output quality.

**Data.** Progression across three K1/TGY4 windows: the initial row-owned M2 cut (`whole-moe-row-owned-m2...20260719`) was a wash — 153.90 vs 151.33 tok/s (+1.70%) but cycle time *worsened* 0.67%, correctly not promoted. Splitting M2 Stage 1 (`whole-moe-tiled-stage1...20260720`) fixed the occupancy problem: 161.86 vs 156.44 tok/s (+3.47%), cycle −2.84%, promoted. Composed with the C1 GDN postconv leg (`gdn-combined-c1-tiled-promotion-ab-20260721-154815`): 166.60 vs 161.75 tok/s (+3.00%), cycle −5.3%.

### 3. GDN post-conv fusion — +0.82%

**What it does.** Qwen3.6-35B-A3B interleaves 30 Gated Delta Net (linear-attention) layers with 10 full-attention layers (every 4th layer is full attention). `--fuse-gdn-postconv` (impl select `headquarter`) replaces the multi-dispatch tail of each GDN layer — everything downstream of the causal depthwise convolution — with a single fused Metal kernel per layer.

**How it works.** Each GDN layer projects hidden state into packed q/k/v, an output gate `z`, a decay gate `a`, and a write-strength gate `b`. q/k/v run through a short causal conv1d (unchanged, its own stock kernel); everything after that conv is the gated-delta recurrence. Per position: a data-dependent decay `g = exp(-exp(A_log) · softplus(a + dt_bias))` shrinks a per-head fp32 state matrix `S` (128×128); `beta = sigmoid(b)` scales a delta-rule correction `(v − k·S) · beta` — the residual between the actual value and what the current associative memory would retrieve — which is written back as a rank-1 update `S += k ⊗ delta`; output is `q·S` read from the updated state. The C1 "headquarter" kernel fuses q/k RMS-norm+scale, gate computation, the decay/correction/update, and per-position state capture into one `mx.fast.metal_kernel` launch per layer, covering both decode shapes MTP speculative decoding uses: M1 (T=1, draft/correction row) and M2 (T=2, verify row). One threadgroup owns a (value-head, state-quarter) tile; a single simdgroup computes the shared per-position q/k normalization and gates once into threadgroup memory instead of redundantly per simdgroup, and each simdgroup then keeps its slice of the fp32 state resident in registers across the T loop.

**What it optimizes away.** The stock path round-trips q_norm, k_norm, gate values, and the `[1,32,128,128]` fp32 state through device memory between stages, once per GDN layer — 30 times per generated token. Fusing removes those intermediate materializations and collapses dispatch count, which matters at decode because each token is launch/latency-bound rather than compute-bound.

**Data.** Isolated A/B (`gdn-c1-promotion-ab-20260721-130710`, natural K1/TGY4, 3 paired passes): 164.4 vs 163.0 tok/s, **+0.82%** (cycle −0.82%). Bit-exact: SHA-match and acceptance-match true across all 6 arm measurements. Combined with the tiled MoE kernel: 166.6 vs 161.7 tok/s, +3.0% (cycle −5.36%) — that combined run carries the tiled component too, so the +3.0% isn't attributable to postconv alone.

## Supporting kernels — no isolated delta

These three collapse per-launch overhead in the MoE forward. None has a clean isolated per-kernel benchmark: combine-tail was measured alone and came out a wash (−7.4% first-pair, +0.15% retuned); packed gate/up and row-owned router were only ever measured in-stack. They are on in every configuration measured above — including the autoregressive baseline — so their contribution is already inside the 119.5 tok/s starting point rather than any ladder step. They are documented here for completeness, without a percent.

### Packed MoE gate/up

**What it does.** Fuses the gate and up projection matmuls of the sparse MoE block into one matmul each, for both the 256 routed experts and the always-on shared expert. Construction-time weight surgery, done once at model load (`--moepack`).

**How it works.** `Qwen3NextSparseMoeBlock` normally runs gate and up as two separate calls on the same activation, differing only in which output rows of the weight they produce. At load time, `configure_moe_packed_projections` concatenates each block's gate and up weight (plus scales/biases when quantized) along the output-feature axis — axis 1 of the switch tensor's `[experts, out, in]`, axis 0 of the shared expert's `[out, in]`. Affine-quant groups run along the *input* axis, so concatenating output rows never crosses a group boundary: no requantization, bit-exact. `PackedSwitchGLU`/`PackedGateUpMLP` reproduce `SwitchGLU`/`Qwen3NextMLP` call-for-call, issuing one `gather_qmm`/`quantized_matmul` over the packed weight then `mx.split`ting the result apart. An unsafe pair — bias, mismatched quantization, widths, expert counts, or dtypes — gets a logged skip, never a mispack. Tests assert bitwise parity against stock output across all paths.

**What it optimizes away.** Dispatch count, not arithmetic. An A3B decode forward touches roughly 1.4 GB of active weight but issues several hundred kernel launches to move it, well below the memory-bandwidth roofline — per-launch and graph-construction cost dominate over raw FLOPs. Packing removes one matmul dispatch and its graph node from each gate/up pair, per MoE layer, on both trunk and MTP draft block. Same dot products, fewer launches.

**Data.** No isolated moepack-only ablation exists; its effect is measured only in-stack. The clean stack A/B (`beneficial-stack-natural-k1-tgy4-20260718`): candidate 153.0 vs control 142.2 tok/s, **+7.6%** (cycle −8.4%) — the moepack + row-owned-router + combine-tail + compiled-route + postconv bundle against full stock.

### Row-owned router

**What it does.** Replaces the expert-selection stage of the MoE gate — top-8-of-256 selection plus probability renormalization — with a single fused Metal kernel for the 1–2 row batches that occur during speculative-decode verify and AR decode (`--row-owned-router`). Prefill and any row count outside the eligible range fall through to the untouched stock path.

**How it works.** Stock routing is `softmax` → `mx.argpartition(kth=-8)` → `mx.take_along_axis` → sum → divide — four-plus separate dispatched ops after the softmax. The row-owned kernel keeps `softmax(..., precise=True)` stock, then fuses argpartition, gather, and normalization into one kernel: each of the 256 experts is assigned a thread, an 8-round SIMD tournament finds the top-8 per SIMD group, group 0 merges the two group-local results, and lane 0 writes normalized BF16 scores and uint32 expert ids for that row — one threadgroup owns one row end to end, no cross-threadgroup coordination. Ties resolve toward the higher expert index, matching `argpartition` exactly. Tests confirm bitwise-exact expert ids and scores against stock for every row count 1–16, and the model self-check requires `dmax == 0.0` before installation is allowed.

**What it optimizes away.** At 1–2 rows, top-8-of-256 is negligible arithmetic; the cost is dispatch — four-plus separate MLX kernel launches, each with fixed encode/dispatch overhead, for a computation processing at most a few thousand floats. Routing at decode batch sizes is launch-bound, not compute-bound, so collapsing the post-softmax stages into one kernel removes launches rather than FLOPs.

**Data.** Measured in-stack only (the `beneficial-stack` bundle above, +7.6%). The `mtp-router-phase` window keeps `--row-owned-router` on in *both* arms, so its own delta isolates a different change and says nothing about the router alone. No standalone isolated benchmark exists.

### Combine-tail fusion

**What it does.** After the router picks each token's top-8 experts and normalized route scores, the per-expert outputs `[rows, 8, 2048]` must become the final `[rows, 2048]` hidden state: multiply each expert's output by its route score, sum over the 8 experts, then add the shared-expert contribution. This kernel fuses the score-weighted sum.

**How it works.** The stock path does `(routed * scores[..., None]).sum(axis=-2)` — a broadcast multiply materializing a full `[rows, 8, 2048]` tensor, then a separate reduction. The fused version (`qwen_combine_tail_source`) is a single `mx.fast.metal_kernel` that assigns one thread per output element and does the 8-way weighted accumulation in an unrolled BF16 loop directly into the final `[rows, 2048]` buffer — no intermediate tensor ever exists. Decode's 1–2 row shapes get dedicated launch geometries (`qwen_combine_tail_m1`, `qwen_combine_tail_m2`); other row counts fall back to the stock reduction. The shared-expert add stays a separate op.

**What it optimizes away.** The materialized `[rows,8,2048]` intermediate and the two-launch multiply→sum chain collapse into one kernel dispatch — cutting launch overhead and memory traffic on the reduction tail.

**Data.** No clean standalone win: the isolated first-pair window (`shared-combine-tail...20260719`) measured 145.4 vs 157.1 tok/s (−7.4%, with parity-mismatch failures, not promotion-eligible alone); a retuned window measured +0.15%. This kernel ships as part of the fused decode stack (the +7.6% bundle) where its launch-count reduction pays off in aggregate, not on its own delta.

## Correctness — including quality under the fused (approximate-arithmetic) kernel

The whole-MoE fusion reduces the expert output in a different accumulation order than stock `gather_qmm`, so its result differs from stock by ulps (documented tolerance: route ids/scores exact, activations dmax ~0.0013). We verify this precision difference does not change task output:

| Gate | Proves | Result |
|---|---|---|
| Variant exactness | each config == its AR ground truth | byte-identical |
| M1/M2 row parity | single-row MoE == verify MoE, all 40 layers | dmax 0.0 (enforced at load) |
| No-regression | stock (flags-off) config unchanged vs base | byte-identical |
| Task quality | fused-kernel precision does not change output | see below |

The two suites have different denominators — HumanEval is 164 problems, MBPP is 974 — so they are reported separately (fused config vs stock-arithmetic config, full verdict-grade run):

| Suite | Fused (candidate) | Stock arithmetic (control) |
|---|---|---|
| HumanEval | **154 / 164** | 152 / 164 |
| MBPP | **790 / 974** | 792 / 974 |

Parity within single-seed noise (+2 HumanEval, −2 MBPP) — the ulp-level accumulation difference in the fused MoE reduction does not change task output.

## Optional: depth-2 (k=2) speculation — opt-in, default off

A generic 3-row verify path is included so the compiled route + whole-MoE fusion aren't hard-locked at k=1. Default off; the k=1 (199 tok/s) path is byte-identical.

**It is lossless.** Greedy k=2 output is byte-identical to `generate_ar` of the same fused config on both the naturalistic and long_code prompts; the 3-row (m3) MoE kernels are proven bit-exact against the single-row path (m1↔m3 and m2↔m3 parity, dmax 0.0 on the real 35B), enforced by a fail-closed install self-check.

**It loses on this checkpoint and is not a default** (greedy, 256 tokens):

| prompt | k=1 | k=2 | k=2 acceptance (a1, a2) |
|---|---|---|---|
| naturalistic 133-tok | 209.9 | 124.5 | 0.38, 0.10 |
| long_code | 195.7 | 121.8 | 0.37, 0.09 |

The acceptance shown is **artifact-depressed** — this prototype's depth-2 draft producer uses a fresh MTP cache without the committed history the k=1 draft has, so even draft-1 is starved (a1 0.37 vs k=1's ~0.76); it's a fixable prototype limitation, not the structural rate (the depth-2 draft's structural acceptance is ~0.45). But k=2 loses regardless of that fix: the **3-row cycle costs ~1.34× the 2-row cycle**, and A3B ships a **single** trained MTP module (`mtp_num_hidden_layers=1`) reused out-of-distribution for the second draft, so even structural a2 ≈ 0.45 can't cover the cycle-cost overrun. The kernels are included as a byte-exact generic-k foundation — they pay only for a checkpoint/head with a real second draft, not this one.

## Reproduce

```sh
PY=<venv>/bin/python
cd bench/a3b

# headline + end-to-end ladder (byte-gated greedy):
$PY run_a3b_174_variant_exact_gates.py

# single configs (all flags encoded in the config name):
$PY a3b_run.py k1                     # 199 tok/s fullroute
$PY a3b_run.py k1 --sampler contract  # serving shape
$PY a3b_run.py ar                     # AR ground truth

# quality: fused-kernel precision vs output:
MTPLX_QGATE_WORKTREE=<worktree> $PY run_a3b_174_combined_quality_gate.py
```

Per-kernel A/B promotion windows referenced above live in the bench ledger (`bench/a3b/*.summary.json`).

---

<details>
<summary>Original PR description</summary>

Two independent decode improvements for Qwen MoE artifacts (Qwen3.6-35B-A3B), both scoped to the resident model — no expert-streaming machinery involved.

Published A3B depth sweep, used as the reference throughout:

| mode | tok/s | acceptance |
|---|---|---|
| AR baseline | 94.46 | — |
| **D1** | **138.39** | 0.8858 |
| D2 | 135.66 | 0.8701, 0.6409 |
| D3 | 107.67 | 0.8291, 0.5414, 0.2783 |

---

## 1. `mtp_depth_max` is a ceiling, not the default depth

`_model_contract_depth` returned an artifact's `mtp_depth_max` directly as the **default** speculative depth. That field is a *ceiling* — the deepest MTP sidecar the artifact ships — so every model ran at its maximum depth whether or not that depth was its fastest mode.

A3B declares `mtp_depth_max: 3`, so it selects D3. Per-level acceptance decays steeply on this MoE (0.829 / 0.541 / 0.278), so verify cost outruns the extra accepted tokens past D1, and the ceiling costs **−22.2%** against D1 out of the box.

**Repro** — `benchmarks/repro_a3b_depth_default.py`, against a local A3B artifact:

```
contract mtp_depth_max   : 3   (a CEILING)
contract recommended     : sustained
resolved default depth   : 3

STUCK AT 3: the ceiling is returned as the default depth.
```

It passes `fallback=1`, so the `3` can only originate from the contract ceiling, not a caller default. After the change the same script reports `1`. Note the contract is profile-scoped: A3B declares `recommended_profile: sustained`, and the pin only reproduces under the matching profile.

**Fix.** Artifacts may declare `mtp_depth_default`; `mtp_depth_max` then only bounds it. `_MODEL_CONTRACT_DEPTH_DEFAULTS` carries measured defaults for already-published artifacts that predate the field (same measured-win-only shape as the existing `_TURBO_DEFAULT_PUBLIC_MODEL_IDS`). Artifacts declaring neither keep the previous behaviour, and explicit `--depth` / `--speculative-depth` still wins.

## 2. Pack MoE gate/up projections at load time (default off)

Qwen's sparse MoE block issues two matmuls for the gate and up projections in each half of the block: `gather_qmm` twice in the routed `SwitchGLU`, and `quantized_matmul` twice in the shared expert. Both pairs read the same activation and differ only in which output rows they consume, so the two matrices are concatenated along their output-feature axis once at load time and evaluated as a single matmul whose result is split in two.

This targets **dispatch count**, not arithmetic. An A3B decode forward touches roughly 1.4 GB of active weights but issues several hundred kernel launches to do it, leaving it far below the memory-bandwidth roofline and making per-launch and Python graph-construction cost dominant. Packing removes two matmul dispatches — and the graph nodes that build them — from every MoE layer, on the trunk and the MTP draft block alike.

The packed forward is ordinary MLX ops rather than a custom Metal kernel, so unlike an `mx.fast.metal_kernel` lane it still composes with `mx.compile` and the graphbank verify paths.

**Exactness.** Affine quantization groups run along the *input* axis, so concatenating output rows leaves every group intact: no requantization, and each output element is produced by exactly the dot product that produced it before. Tests assert this bitwise (`mx.array_equal`, not a tolerance) on the CPU backend, for the plain and gathered forms, across the `SwitchGLU` sorted and unsorted paths, and end to end through the MoE block.

**Interaction worth noting:** packing changes the projections' type, so two default-off lanes stop seeing them — `native_mlp`'s `MTPLX_MLP_CALL_VARIANT` (dense-targeted, inactive at M=1) and the NAX verify patch (`nn.QuantizedLinear.__call__`). Documented in the module docstring.

Behind `MTPLX_QWEN_MOE_PACK_GATE_UP`, default off.

---

## Tests

`tests/test_moe_packed_projections.py` (29 cases) and `tests/test_public_cli.py` — pytest exit code 0. All MoE-packing tests run inside `with mx.stream(mx.cpu)`; no GPU required.

## Status

Correctness is gated by the parity tests. **Decode throughput on a Qwen MoE artifact is not yet measured** — the depth figures above are the published sweep, not our measurement, and no tok/s claim is made for the packing change. GPU parity on a real artifact plus an A/B decode measurement are the remaining items before this leaves draft.

Supersedes #172 and #173, which are the same two commits split apart.

https://claude.ai/code/session_01YHUiWXTFYuksrafoFWoPbB

</details>

---
**Note (2026-07-31):** the m6 K-split verify kernel introduced here gains
optional 6-bit weight support in stacked PR #200 (additive `bits` parameter,
default 4; rendered 4-bit Metal source test-asserted byte-identical to this
branch's). Motivation: the Qwen3.6-35B-A3B 6-bit "Balance" build's dense
projections, which otherwise fall to the stock path on solo-MTP verify.


davidtai:
My personal analysis is I'm discovering a bunch of bugs and there's some magic around k=4 specfically for turbo mode precision that's throwing my backport off so this might take a little bit.  Also been fighting with the verifier.  Partial backport of my row major kernel optimizations yielding 155tps decode at k=1 but I think we can push it over 200 since theoretical saturation is like 260-300.  Should be done this weekend.

youssofal:
Part 1 of this shipped in 2.2.0, credited to you in the commit, with one correction from fresh measurements.

The ceiling-as-default diagnosis is confirmed: `_model_contract_depth` returned `mtp_depth_max` directly, so the A3B ran its deepest sidecar out of the box. Your `mtp_depth_default` contract field, the measured-defaults map bounded by the ceiling, and explicit `--depth` winning are all in, along with your four tests.

The correction: I re-swept the A3B on an M5 Max under pinned fans before shipping the default and got AR 93.9, D1 130.0 (1.39x), D2 145.0 (1.54x), D3 132.2 (1.41x), with D2 acceptance [0.813, 0.575]. Your table has D1 ahead of D2 by 2%, inside noise; mine has D2 ahead by 10%. Across both datasets D2 is at or within noise of best while the D3 ceiling loses 9 to 22%, so the map pins D2. If your sweep hardware was not M5-class I would genuinely like to know what it was, because a real chip-dependent D1/D2 split would argue for per-tier defaults.

Part 2 (the gate/up packing) stays with you as scoped: once you have GPU parity on a real artifact plus an A/B decode number, it gets reviewed on its own merits. Default-off was the right call.

Your depth work also compounded with #176: the 4B checkpoints drafted at zero acceptance everywhere, so ceiling-as-default had fresh users at roughly half AR speed. The default split plus the artifact fix closes that pair in 2.2.0. Thanks for the quality here.

davidtai:
Will do, I thought I'd have it over 200 by now but there's some really intricated timings on this specific model.

davidtai:
image
I just had to prompt max a little to get over 200tps yay

davidtai:
### Optional depth-2 (k=2) speculation — added for completeness (byte-exact, opt-in, default off)

Pushed a generic 3-row verify path so the compiled route + whole-MoE fusion aren't hard-locked at k=1. It is **default off** and does not touch the shipped k=1 (199 tok/s) path, which stays byte-identical.

**Correctness — it is lossless.** Greedy k=2 output is byte-identical to `generate_ar` of the same fused config on both a naturalistic 133-token coding prompt and long_code. The 3-row (m3) MoE kernels are proven bit-exact against the single-row path (m1↔m3 row parity and m2↔m3 prefix parity, dmax 0.0 on the real 35B), enforced at install by a fail-closed self-check.

**Throughput — it loses on this checkpoint, and is not a default.** Measured (greedy, 256 tokens, M5 Max):

| prompt | k=1 | k=2 | k=2 acceptance (a1, a2) |
|---|---|---|---|
| naturalistic 133-tok | 209.9 tok/s | 124.5 | 0.376, 0.098 |
| long_code | 195.7 tok/s | 121.8 | 0.374, 0.092 |

Two reasons k=2 is slower here: (1) the 3-row cycle costs ~1.34× the 2-row cycle, and (2) depth-2 acceptance is very low because A3B ships a **single** trained MTP module (`mtp_num_hidden_layers=1`) that k=2 must reuse recurrently for the second draft — an out-of-distribution use that caps a2 near zero. This matches the maintainer's own note that D2 is only worth it with sufficient draft acceptance; on the tested MTP head that acceptance isn't there.

**Why include it anyway:** the kernels are a byte-exact generic-k foundation. They pay for any checkpoint/head with a real second draft (a distinct MTP module or a higher-a2 draft head) — for this checkpoint, k=1 is the operating point. The k=1 acceptance itself is prompt-dependent: greedy on the naturalistic prompt runs **209.9 tok/s** (vs ~196 on long_code), so k=1 already clears 200 on realistic inputs.

youssofal:
merged as d3d7cbb with your authorship a few days back, sorry for the slow close. ran the full suite + our own A/B on the M5: flags-off is dead flat on both lanes (27B pooled 54-55, 35B 118.9 vs 119.0), so the stack went in as-is with the D2 default.

the compiled route itself: its request contract (stock cores, cold prompts, no session bank) can never engage under the product serve path. sessionbank is on for every real session and that's not something we'll turn off, warm-prefix reuse is the product. so the 196-207 numbers stay bench/serving-lane for now, and the app default got the reachable subset (moepack + compiled target-prefix + gdn-postconv + row-owned router).

the unlock we actually want from you: make the route compose with sessionbank. engage on warm prefixes, or at minimum tolerate the bank owning cache identity. if the route runs under the product serve path it becomes the 35B app default and that's a headline release.

PR #183, opened 2026-07-19
fix(mtp): correct group_size inference for non-power-of-2 bit widths (5-bit MTP heads)

Jonathangadeaharder:
Fixes #182.

## Problem

`_infer_prequantized_group_size` (`mtplx/mtp_patch.py`) expands packed uint32 columns back to a logical element count using integer-floor division:

```python
values_per_word = max(1, 32 // int(bits))
expanded_cols = packed_cols * values_per_word
```

For a 5-bit MTP head this undercounts. MLX packs `32 / 5 = 6.4` values per uint32 (640 uint32s hold exactly 4096 5-bit values: `640 * 32 / 5 = 4096`), but `32 // 5 = 6` gives `expanded_cols = 640 * 6 = 3840`, so the inferred `group_size = 3840 / 64 = 60` instead of `64`.

`60` does not divide common hidden dims (4096, 2048), so `nn.quantize(group_size=60, bits=5)` rejects the layer:

```
ValueError: [quantize] The last dimension of the matrix needs to be divisible by
the quantization group size 60. However the provided matrix has shape (2048,4096)
```

This blocks every MTPLX model whose MTP head is prequantized at a non-power-of-2 bit width (the Shiftedx `mxfp4-vision-mtplx` family), even though those models declare `mtp_quant_group_size: 64` in their `mtplx_runtime.json`.

## Fix

When `bits` does not evenly divide 32, derive `expanded_cols` from total bits instead of the floor-divided values-per-word:

```python
if 32 % bits_int == 0:
    expanded_cols = packed_cols * values_per_word
else:
    total_bits = packed_cols * 32
    if total_bits % bits_int != 0:
        continue
    expanded_cols = total_bits // bits_int
```

For the 5-bit case: `expanded_cols = (640 * 32) / 5 = 4096`, so `group_size = 4096 / 64 = 64`.

The `32 % bits_int == 0` branch keeps the original code path verbatim for power-of-2 widths (4, 8), so existing 4-bit models are unaffected.

## Verification

Before: loading `Shiftedx/ornith-1.0-35b-abliterated-mxfp4-vision-mtplx` on MTPLX 2.1.0 / MLX 0.31.2 raises the `group size 60` error.

After: the model loads. MTP gate resolves to 5-bit / group 64, matching the contract.

Regression: four 4-bit MTP models still load and run unchanged.

- `wang-yang/Ornith-1.0-35B-MTPLX`
- `Jonandrop/Ornith-1.0-35B-MTPLX-Vision`
[...]

PR #185, opened 2026-07-20
Fix _infer_prequantized_group_size undercount for non-power-of-2 bit widths (fixes #182)

### LINKED ISSUE #182: [BUG] _infer_prequantized_group_size undercounts 5-bit (and non-power-of-2 bit) MTP heads, breaks load with group_size 60
## Summary

`MTPContract` infers the quantization `group_size` for prequantized MTP heads via `_infer_prequantized_group_size` (`mtplx/mtp_patch.py:373`). It uses integer-floor division `values_per_word = 32 // bits` to expand packed uint32 columns back to logical element count. For MTP heads quantized at **5-bit** (and any bit width that doesn't evenly divide 32), this undercounts and produces a wrong `group_size`, causing `nn.quantize` to crash at load with:

```
ValueError: [quantize] The last dimension of the matrix needs to be divisible by the quantization group size 60. However the provided matrix has shape (2048,4096)
```

## Root cause

For a 5-bit MTP head, MLX packs `32 / 5 = 6.4` values per uint32 word (640 uint32s hold exactly 4096 5-bit values: `640 * 32 / 5 = 4096`). The current code computes `values_per_word = 32 // 5 = 6`, giving `expanded_cols = 640 * 6 = 3840`, then `group_size = 3840 / 64 = 60`. The correct group_size is **64** (`4096 / 64 = 64`).

`60` does not divide common hidden dims (4096, 2048), so `mx.quantize(group_size=60, bits=5)` rejects the layer.

## Reproduce

Load `Shiftedx/ornith-1.0-35b-abliterated-mxfp4-vision-mtplx` (or any model whose MTP head is prequantized at 5-bit, group_size 64) on MTPLX v2.1.0 with MLX 0.31.2:

```python
from mtplx.runtime import load
rt = load("/path/to/Shiftedx--ornith-1.0-35b-abliterated-mxfp4-vision-mtplx", mtp=True)
# ValueError: [quantize] ... group size 60 ... shape (2048,4096)
```

The model's `mtplx_runtime.json` correctly declares `mtp_quant_group_size: 64`, but the inference path overrides it from the packed weight geometry and produces 60.

## Fix

When `bits` does not evenly divide 32, compute `expanded_cols = (packed_cols * 32) // bits` instead of `packed_cols * (32 // bits)`:

```python
if 32 % bits_int == 0:
    expanded_cols = packed_cols * values_per_word
else:
    total_bits = packed_cols * 32
    if total_bits % bits_int != 0:
        continue
    expanded_cols = total_bits // bits_int
```

This makes the Shiftedx model load (MTP gate at 5-bit, group 64) and is regression-clean across the 4-bit models I tested (wang-yang/Ornith, Jonandrop/Ornith-vision, Jonandrop/Qwen3.6-27B-vision, Youssofal/Qwen3.6-27B).

## Measured impact

With the fix, the Shiftedx mxfp4-vision model loads and MTP is healthy (Apple M5 Pro 64GB, committed policy, greedy draft, warm):

- AR: 67.6 tok/s
- MTP depth 3: 87.4 tok/s (1.29x), acceptance pos1/2/3 = 0.93 / 0.86 / 0.76

Without the fix, the model is unloadable in stock MTPLX.

## Environment

- MTPLX: 2.1.0 (CLI), repo at `mtplx/mtp_patch.py`
- MLX: 0.31.2
- macOS: 26.5.2, Apple M5 Pro 64GB

Happy to open a PR with the fix if useful.

PhilipJohnBasile:
## Summary

`_infer_prequantized_group_size` expanded packed uint32 columns with floor division (`32 // bits` values per word), which undercounts logical columns for bit widths that don't divide 32. For a 5-bit MTP head with 4096 logical columns (640 packed uint32s) it produced 3840 columns and inferred `group_size = 60` instead of 64, and `nn.quantize` then rejected the layer at load — making 5-bit prequantized models (e.g. `Shiftedx/ornith-1.0-35b-abliterated-mxfp4-vision-mtplx`) unloadable.

Full credit to @Jonathangadeaharder for the diagnosis, proposed fix, and hardware verification in #182.

Fixes #182

## Fix

MLX packs quantized rows bit-contiguously (`packed_cols * 32 == logical_cols * bits`), so recover the logical column count as `packed_cols * 32 // bits`, and skip any layer whose packed bit count isn't divisible by the bit width (no logical column count is recoverable there, so the declared contract value stands). Power-of-2 widths are unaffected: both formulas agree whenever `bits` divides 32.

## Tests

- `test_prequantized_mtp_tensor_geometry_handles_non_power_of_two_bits` — the 5-bit geometry from the issue (640 packed cols, 64 scale groups → group_size 64). Fails with 60 on the old code.
- `test_prequantized_mtp_tensor_geometry_skips_indivisible_bit_geometry` — 3-bit geometry where `packed_cols * 32 % bits != 0`; the declared group_size must stand. Fails with 80 on the old code.

Both new tests verified to fail against the unfixed code and pass with the fix. Full suite: 1933 passed, 5 skipped; ruff clean.

https://claude.ai/code/session_01LdBYNdiLyQ6qkqPA1Z2MkP

### DIFF
--- a/mtplx/mtp_patch.py
+++ b/mtplx/mtp_patch.py
@@ -433,7 +433,7 @@ def _heal_raw_delta_mtp_norms(weights: dict[str, Any]) -> dict[str, Any]:
 def _infer_prequantized_group_size(weights: dict[str, Any], bits: int | None) -> int | None:
     if bits is None or bits <= 0:
         return None
-    values_per_word = max(1, 32 // int(bits))
+    bits_int = int(bits)
     inferred: set[int] = set()
     for key, weight in weights.items():
         if not key.endswith(".weight"):
@@ -452,7 +452,13 @@ def _infer_prequantized_group_size(weights: dict[str, Any], bits: int | None) ->
         scale_groups = int(scales_shape[-1])
         if packed_cols <= 0 or scale_groups <= 0:
             continue
-        expanded_cols = packed_cols * values_per_word
+        # MLX packs quantized rows bit-contiguously: packed_cols * 32 == logical_cols * bits,
+        # even when bits does not divide 32 (e.g. 5-bit). Flooring 32 // bits per word
+        # undercounts logical columns for those widths.
+        total_bits = packed_cols * 32
+        if total_bits % bits_int != 0:
+            continue
+        expanded_cols = total_bits // bits_int
         if expanded_cols % scale_groups == 0:
             inferred.add(expanded_cols // scale_groups)
     if len(inferred) == 1:
--- a/tests/test_mtp_patch.py
+++ b/tests/test_mtp_patch.py
@@ -156,6 +156,52 @@ def test_prequantized_mtp_tensor_geometry_corrects_group_size() -> None:
     assert corrected.mtp_quant_group_size == 32
 
 
+def test_prequantized_mtp_tensor_geometry_handles_non_power_of_two_bits() -> None:
+    class FakeArray:
+        def __init__(self, shape: tuple[int, ...]):
+            self.shape = shape
+
+    contract = MTPContract(
+        mtp_quant_bits=5,
+        mtp_quant_group_size=32,
+        mtp_prequantized=True,
+    )
+    # 4096 logical columns at 5-bit pack into 4096 * 5 / 32 = 640 uint32 columns;
+    # 64 scale groups imply group_size 4096 / 64 = 64.
+    weights = {
+        "layers.0.self_attn.q_proj.weight": FakeArray((2048, 640)),
+        "layers.0.self_attn.q_proj.scales": FakeArray((2048, 64)),
+        "layers.0.self_attn.q_proj.biases": FakeArray((2048, 64)),
+    }
+
+    corrected = _contract_with_prequantized_tensor_geometry(contract, weights)
+
+    assert corrected.mtp_quant_group_size == 64
+
+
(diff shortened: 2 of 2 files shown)


PhilipJohnBasile:
A note for the review queue: I opened this without noticing that @Jonathangadeaharder had already put up #183 — filed a minute after the issue itself, well before this PR. The diagnosis and the fix there are theirs; the two patches are functionally equivalent (same math, mine folds the power-of-2 and non-power-of-2 cases into one formula).

What this PR adds over #183 is the regression coverage: the 5-bit geometry from the issue (640 packed cols / 64 scale groups → group_size 64, fails as 60 on the old code) and the indivisible-geometry guard (3-bit where no column count is recoverable, fails as 80 on the old code), both verified to fail pre-fix.

Whichever way is easiest for you: merge #183 and I'll rebase this down to just the two tests, or take this one — no preference, credit for the find belongs to #183 either way.

youssofal:
Closing in favor of #183, which reported the bug and landed first — the two implementations are functionally identical (for widths dividing 32 the general formula and the branch structure agree in every case). Your two test cases are covered by the six-case regression suite that shipped with the fix (5-bit heal and preserve, 3-bit indivisible skip, 6-bit, 8-bit, padded geometry). Thanks for the clean write-up.

Issue #186, opened 2026-07-20
[Feature] Grammar-constrained structured output: response_format json_schema + strict tool-call arguments, staged to compose with the MTP verify path


PhilipJohnBasile:
## Summary

Proposal: grammar-constrained decoding for MTPLX — honest `response_format` handling (`json_object` / `json_schema`) and opt-in strict tool-call argument enforcement keyed off the request's declared `tools` schemas, staged so the conservative version ships first and composition with the MTP fast paths lands separately, behind measurements.

I'm offering to implement this. Filing first because the later phases touch the verify path and I'd like a design ack before building on it. File:line references are against current `main` (54a1d9a).

## Today's behavior

`response_format` is accepted and silently ignored: it's a field on `ChatCompletionRequest` (`mtplx/server/openai.py:737`) whose only consumer is the observability helper that records it as a "declared extra key" (`mtplx/server/openai.py:13208`). A client sending `{"type": "json_schema", ...}` gets unconstrained text with no warning — arguably worse than a 400, because clients assume the schema was enforced.

Constrained decoding is table stakes in peer runtimes (vLLM/xgrammar, llama.cpp GBNF, SGLang, TGI, Ollama), and the agentic clients this tracker keeps hearing from (#170, #86, #169) are the population that most depends on tool-call validity.

Scope honesty up front: this does **not** claim to fix #170 outright. If that bug's root cause is template-side rendering of *history* (the `|items` hypothesis), a decode-side grammar doesn't touch it. What strict mode guarantees is narrower and still valuable: generated tool-call arguments always conform to the declared schema, so a tool whose schema marks properties `required` can never emit `{}` from the decode side, for every model family, without per-template whack-a-mole.

## Phase 1 — honest `response_format`

1. Reject unsupported `response_format` shapes with a clear 400 instead of silently ignoring them.
2. Enforce `json_object` / `json_schema` with a per-step token mask from a grammar engine, applied to logits before shaping/normalization.

Engine: llguidance is the safer first choice — it ships an ABI3 arm64 macOS wheel covering MTPLX's `requires-python = ">=3.11"` (`pyproject.toml:10`) including 3.13/3.14, whereas xgrammar's latest arm64 wheels currently cover 3.10–3.12 only. Two things it does *not* give us for free, so they're in scope for this phase:

[...]

Issue #170, opened 2026-07-17
[BUG] edit_file tool-call arguments intermittently collapse to {} on Qwen3.6 27B (mode-independent, ~10% of edit turns)


druide67:
## Summary

On agentic tool-call sessions, `edit_file`-style calls (whose arguments carry an array of objects) intermittently arrive with empty arguments (`{}`), while simpler calls on the same session are fine. 2.1.0 fixed the MTP-specific half of this (nice work, details below); a mode-independent residual of about 10% of edit turns remains, and the evidence points at the chat template rather than the decode path.

Same disclosure as #169: measured with [asiai](https://github.com/druide67/asiai) (open-source bench tool, I build it), drafted with AI assistance, everything reproduces from the commands below.

## Environment

Same as #169: M5 Max 128 GB, macOS 26.5.1, MTPLX 2.1.0 (venv-2.1.0, mlx 0.32.0), `Youssofal/Qwen3.6-27B-MTPLX-Optimized-Speed` (`qwen3_next`), serial scheduler, temp 0 suite (deterministic), `--chat-template-profile local_qwen36`, `--ssd-session-cache off`.

## Numbers (8-turn deterministic agentic suite, 3 runs = 24 turns per cell)

| Version | Mode | clean turns | empty-args events | edit turns clean |
|---|---|---|---|---|
| 2.0.2 (measured 2026-07-14) | MTP d3 | 75.0% | 6 | 33% |
| 2.0.2 | AR | 87.5% | 3 | 66.7% |
| 2.1.0 (yesterday) | MTP d3 | 91.7% | 2 | 77.8% |
| 2.1.0 | AR | 87.5% | 3 | 66.7% |

Two reads: (1) the 2.0.2 MTP penalty is gone in 2.1.0, MTP now scores at or above AR on tool calls; (2) the residual failure rate is the same in both generation modes, so it does not look decode-side.

## Why I suspect the template

The upstream Qwen3.6 chat template builds tool-call examples through the Jinja `|items` filter, which several non-Python renderers mishandle; the known symptom is exactly this one, nested object arrays collapsing to `{}`. Same model family served by llama.cpp with the community-fixed template (froggeric's, which replaces `|items`) scores 100% on the identical suite and hardware. I have not read how `local_qwen36` renders, so this is a hypothesis with a matching fingerprint, not a diagnosis.

## Repro

[...]

PR #188, opened 2026-07-20
Structured output phases 2+3: strict tool-call grammars + composition with the MTP verify loop (#186)

PhilipJohnBasile:
Stacked on #187 (this branch contains its commit — review/merge #187 first; this PR's own change is the single top commit).

Implements phases 2 and 3 of #186.

## Phase 2 — strict tool-call arguments (opt-in `MTPLX_TOOL_CALL_STRICT=1`)

When a request declares `tools`, the decode grammar allows free text and native thinking blocks, but any tool-call envelope the model opens is forced to carry a declared tool name and schema-valid arguments — malformed calls (wrong name, missing required keys, out-of-range values, truncated JSON like #170's `{}` collapses) become unsamplable rather than parse-time failures.

- Grammars are hand-rolled lark rather than llguidance `StructTag`: the sugar only specializes the *trigger* into a special-token terminal, so a closing `</tool_call>` inside its `end` string stays bytes and never matches the token Qwen actually emits. Closing markers are referenced at rule level.
- Activates only when the template's tool-call markers are single special tokens (Qwen/Hermes family); other templates 400 with a clear message. `tool_choice` auto/none supported; `required`/pinned-function 400s for now.
- Env-gated and default-off, following how context-copy shipped.

## Phase 3 — grammar constraint composed with the MTP speculative loop

Constrained requests now ride `generate_mtpk` instead of pinning to serial AR. The composition is **target-side only** — no draft-core, reranker, or ensemble surgery:

- The cycle **primary** is sampled under the grammar mask (the guaranteed-progress site).
- **Draft windows stay unmasked**; the accept loop clamps acceptance at the grammar-legal prefix (one stateless `validate_tokens` per cycle).
- Grammar-illegal rejection corrections, bonus tokens, and context-copy residuals are dropped; copy blocks are truncated at the first illegal token.
- The matcher mutates only at the cycle-top sync (before masking, so the mask is computed at the true grammar position), after the primary append, and once post-loop.

Correctness: sample-unmasked → discard-illegal → masked-resample yields exactly the masked target distribution per position (P(x) = p(x) + (1−p(legal))·p_masked(x) = p_masked(x)), so clamping is not an approximation.

**Measured (Qwen3.6-35B-A3B-Optimized-Speed, alternating warmed runs):** constrained MTP at parity with unconstrained — 124.6–129.7 vs 124.2–126.5 tok/s, 82–86% draft acceptance with the constraint active, every output schema-valid. That's ~+37% over the phase-1 AR pin (~92 tok/s): **constrained decoding now keeps the MTP speedup**.

## The template-opens-think discovery (affects phase 1 users too)

Qwen templates end the generation prompt with `<|im_start|>assistant\n<think>\n` — generation starts *inside* a thinking block. A grammar that only accepts paired `<think>…</think>` makes the lone closing tag illegal; in testing, the model (unable to exit reasoning) emitted identical tool calls until the token cap. Both grammar families now accept an optional leading `TEXT </think>` prelude. Side effect: `response_format` requests get their thinking back (phase 1 silently suppressed it); truncation honesty (`constraint_completed: false`) covers reasoning that overruns `max_tokens`.

## Fail-loud hardening

`GrammarConstraint.advance` now raises on matcher error. During development, a mask computed at a stale grammar position produced `{"name": "Qwen", "specialtyty` **labeled `constraint_completed: true`** — the matcher had errored, its stopped-state guard silently disabled masking, and "stopped" read as "completed". A decode/matcher desync is an internal invariant violation; it now crashes the request instead of streaming corrupt output marked valid.

## Verification

- 17 tests in `tests/test_constrained.py` (4 new for the mtpk composition: a scripted-model harness with an evens-only fake grammar where the model always prefers illegal tokens — committed output stays legal purely via mask+clamp; grammar-terminal stop; inactive-stats control). Full suite: **1948 passed, 5 skipped**; ruff clean.
- Hardware: strict tool call → one schema-valid call (integer bounds + enum enforced), `finish_reason: tool_calls`, thinking preserved, greedy + sampled + streaming; `json_schema` at temp 0/0.8 across seeds → schema-valid with 800+ token reasoning preludes; bad shapes and unsupported combos → 400s; unconstrained traffic byte-identical behavior.

## Not in this PR

Draft-side masking (phase 3b — only worth it if acceptance inside constrained spans measures low on adversarial schemas; at 82–86% it doesn't yet), `tool_choice: required`/pinned grammars, and batch-pump integration (constrained requests still bypass the batched AR lane).

https://claude.ai/code/session_01LdBYNdiLyQ6qkqPA1Z2MkP

### DIFF
--- a/mtplx/constrained.py
+++ b/mtplx/constrained.py
@@ -20,6 +20,7 @@ tokens are set to -inf, which survives every downstream shaping step.
 from __future__ import annotations
 
 import json
+import os
 import threading
 import time
 from collections import OrderedDict
@@ -46,6 +47,24 @@ SUPPORTED_RESPONSE_FORMAT_TYPES = ("text", "json_object", "json_schema")
 # JSON value, so the generic grammar pins the top-level type.
 _JSON_OBJECT_SCHEMA = '{"type": "object"}'
 
+# Strict tool-call constraint markers. These are the Qwen/Hermes-family
+# native tool-call and thinking tokens; strict mode only activates when the
+# runtime tokenizer encodes each marker as a single (special) token, so the
+# grammar can reference it as a hard boundary rather than bytes.
+TOOL_CALL_START = "<tool_call>"
+TOOL_CALL_END = "</tool_call>"
+THINK_START = "<think>"
+THINK_END = "</think>"
+
+
+def tool_call_strict_enabled() -> bool:
+    """Opt-in via MTPLX_TOOL_CALL_STRICT=1/true/on. Default off."""
+    return (os.environ.get("MTPLX_TOOL_CALL_STRICT") or "").strip().lower() in {
+        "1",
+        "true",
+        "on",
+    }
+
 
 class ResponseFormatError(ValueError):
     """Invalid or unsupported ``response_format``; message is client-safe."""
@@ -70,12 +89,18 @@ class ConstraintSpec:
 
 def constraint_spec_from_response_format(
     response_format: Any,
+    tokenizer: Any | None = None,
 ) -> ConstraintSpec | None:
     """Parse/validate a request's ``response_format`` into a ConstraintSpec.
 
     Returns None when no constraint applies (absent or ``type: text``).
     Raises ResponseFormatError for anything the server cannot honestly
     enforce — the caller turns that into a 400.
+
+    When a tokenizer is provided and its template family uses native
+    thinking markers, the grammar accepts an optional leading thinking
+    close (chat templates open ``<think>`` inside the generation prompt, so
+    the model's reasoning must be allowed to finish before the document).
     """
     if response_format is None:
         return None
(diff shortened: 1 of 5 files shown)


PhilipJohnBasile:
Post-ship adversarial review of this diff (independent max-effort pass over `git diff a3fab77..HEAD`, findings verified against the code) — results and two tightenings:

**Confirmed by the review:** the matcher sync topology (cycle-top before masking / post-primary / final), the masked primary site, the accept-loop legality clamp including the all-accept + bonus + stop-truncation interplay, the context-copy truncation and correction gates, rule-level grammar markers, and per-retry-attempt matcher rebuild. No correctness defects found.

**Tightening 1 — test coverage (pushed as 07e315a):** the strict tool-call grammar path had hardware verification but no CI-runnable coverage. There is now an end-to-end unit test driving `tool_call_constraint_spec → GrammarConstraint` adversarially with a tiny tokenizer carrying the Qwen-family special markers: free text unconstrained, lone `</think>` legal, forced envelope head, schema-bounded arguments, return to free text, and an undeclared tool name rejected mid-envelope.

**Tightening 2 — bounding the performance claim:** the 124.6–129.7 vs 124.2–126.5 tok/s parity and 82–86% acceptance were measured on JSON-friendly prompts (model already inclined to emit valid JSON) with long unconstrained thinking preludes, greedy, one model (Qwen3.6-35B-A3B-Optimized-Speed), one machine. The honest general statement is: **constrained MTP costs nothing when the model's drafts mostly agree with the grammar, and degrades toward masked-AR (one committed token per cycle) in spans where they fight it.** An adversarial schema (deep nesting, long enums, formats the model dislikes) would push acceptance down and is the falsifying measurement; if that regime matters to users, phase 3b (draft-side masking via llguidance's draft-token bitmask helpers) is the designed answer. Happy to run that benchmark if useful for the review.

Two behavior notes the review surfaced, documented rather than changed: the envelope close is *forced* as `}` + `\n` + `</tool_call>` (models can't emit whitespace variants — that's the mask working, not a rejection path), and in the json-prelude grammar a complete document followed by a stray `</think>` can legally reparse as prelude text — benign for clients since the reasoning splitter owns everything before `</think>`, but worth knowing when reading raw transcripts.

PhilipJohnBasile:
Ran the adversarial-schema benchmark promised above — it earned its keep by catching a real bug in this PR before finding anything about performance.

**Bug found and fixed (3a34ea6):** the optional think-prelude (`TEXT </think>`) was unconditional, so on runs where the template *doesn't* open a think block (`enable_thinking=False`, non-thinking templates), the model could legally write prose forever as "prelude text" and never start the document — `response_format` was honestly flagged `constraint_completed: false` but never enforced. The prelude grammar is now selected only when the rendered prompt's tail actually ends inside an open `<think>` block (scanned from prompt ids); a regression test covers all three cases (open / closed / unknown).

**The falsifying measurement itself** (thinking off, greedy, hostile schema: 3-level nesting, `$`/`~`/`#` keys, 4-value string enums like `"zx91-KKQ"`, tight integer bounds — a shape the model has zero inclination toward):

| config | tok/s | acceptance | valid |
|---|---|---|---|
| unconstrained MTP (same prompt/regime) | 63.6 | 45/291 (15%) | — |
| friendly schema, constrained MTP | 64.9–66.6 | 11/60 | 2/2 VALID |
| **hostile schema, constrained MTP** | 58.1–59.6 | 47/387 (12%) | 3/3 VALID |
| hostile schema, constrained AR | 96.1 | — | VALID |

Two honest readings:

1. **The constraint's within-lane cost stays small even under a hostile schema** (58–60 vs 64 tok/s, acceptance 12% vs 15% — the grammar barely moves acceptance because this regime's acceptance was already low). Validity held 6/6.
2. **The parity claim from the earlier benchmark is a property of the lane, not the constraint**: this no-thinking prose regime has ~15% draft acceptance *unconstrained*, so depth-3 MTP loses to plain AR (~96 tok/s) here regardless of constraints. Where MTP acceptance is healthy (the thinking-on runs above: 82–86%), constrained rides it at parity; where the regime kills acceptance, constrained requests inherit that — same as every other request. Whether constrained requests should auto-drop depth in low-acceptance regimes is the existing adaptive-depth question, not something this PR changes.

Suite now at 1950 passed / ruff clean with the two new regression tests.

PR #189, opened 2026-07-20
Make silent tool-argument loss visible: degradation telemetry for the {} substitutions behind #170

PhilipJohnBasile:
Follow-up to the #170 investigation. Tracing the full path of tool-call arguments (history in, emission out) found the input side is double-guarded — `_template_tool_call` strict-validates arguments to dicts before the template adapter runs, so history rendering cannot mangle nested `edits` arrays — but the **output side has five places where arguments silently become `{}` or a call silently disappears**:

1. missing-`arguments` defaults in the native-parser path (`item.get("arguments", {})`),
2. the same defaults in the JSON payload coercion,
3. the unparseable-bracket fallback,
4. the final serialization catch-all (`_serialize_tool_call_arguments` → `"{}"`),
5. swallowed native parser exceptions dropping the call entirely.

That silence is why #170's ~10% empty-arguments reports have been so hard to localize: from the client's side, "model emitted `{}`", "model emitted malformed args", and "parser lost the arguments" are indistinguishable.

## What this PR does

- Every `{}` substitution is now **counted and categorized** (`missing_arguments` / `unparseable_arguments` / `non_object_arguments`), carried on `ToolCallExtraction` and surfaced as `mtplx_stats.tool_arguments_degraded` + `tool_arguments_degraded_reasons`, with a server-side warning carrying the request id. Swallowed native parser exceptions surface as `tool_parser_exceptions`. Explicit empty objects (`"arguments": {}` emitted by the model) are *not* flagged — the counters isolate genuine loss.
- Behavior is unchanged for clients (still `{}` on the wire — no new failure modes); the change is observability, so the next #170-style report can say *which* of the three mechanisms fired and at what rate.
- The history-side adapter's `_try_parse_json` used to normalize malformed arguments to `{}` silently. Since the chat path strict-validates before the adapter runs, malformed input there is an invariant violation — it now raises instead of degrading (same fail-loud philosophy as #188's matcher).

## Relationship to #170

With this landed, the `count_empty_object_bug` telemetry in the reporter's harness can be cross-read against `tool_arguments_degraded_reasons`: if responses show `{}` with **no** degradation counter, the model genuinely emitted `{}` (a conditioning problem — and #188's strict tool-call mode makes that unsamplable); if counters fire, the parser is losing real arguments and the reason string says where. Either way the ambiguity that kept #170 open dies.

13 new regression tests (nested-arguments roundtrip through parser and adapter, every degradation source, private marker never leaking to clients, strict input path unchanged). Full suite: 1944 passed, 5 skipped; ruff clean. Independent of #187/#188 — branches directly off main.

### DIFF
--- a/mtplx/server/omlx_bridge/adapter.py
+++ b/mtplx/server/omlx_bridge/adapter.py
@@ -65,6 +65,14 @@ def _content_to_text(content: Any) -> str:
 
 
 def _try_parse_json(value: Any) -> dict[str, Any]:
+    """Parse history tool-call arguments to a dict, or fail loudly.
+
+    The chat path strict-validates arguments before this adapter runs
+    (_template_tool_call), so malformed input here means a caller bypassed
+    that contract. This used to fall back to {} silently, which erased
+    tool-call arguments from rendered history with no trace (issue #170's
+    class of failure); an invariant violation must raise, not degrade.
+    """
     if isinstance(value, dict):
         return value
     if value in (None, ""):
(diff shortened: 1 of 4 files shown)


youssofal:
We root-caused #170 independently and the fix ships in 2.3.0: the streaming parser and every extraction dialect now share one contract — a pure JSON-object function body is the arguments payload, unreadable bodies stay visible content, and the fabrication sites that built empty-dict calls are gone (validated with a 168-turn live A/B, zero collapses). That supersedes the detection points this PR instruments — and candidly, the counters as written would not have fired on the mechanisms behind #170: the engine-side substitutions produced real {} dicts, which this PR deliberately exempts, and the streaming translator (where the other half of the loss happened) isn't instrumented. Your identification of the extraction-lane default sites was correct and independent — credit for that. Closing with 2.3.0; if post-release telemetry shows model-emitted missing-arguments at meaningful rates, the bracket/coerce-lane visibility from here is the shape we'd revisit.

PR #190, opened 2026-07-20
Honor parallel_tool_calls instead of client-name sniffing

PhilipJohnBasile:
`parallel_tool_calls` is declared on the request model and logged in observability, but the actual single-vs-parallel tool-call behavior was decided by sniffing the client name (`"pi"`, `"opencode"` + a prompt-text heuristic). A client that explicitly declared `parallel_tool_calls: true/false` was ignored in both directions.

- New `_single_tool_call_stream_policy`: the declared boolean is authoritative — `false` stops the stream after the first complete call, `true` disables truncation even for hint-matched clients. The existing hint heuristics are untouched as the fallback when the field is absent (no behavior change for current clients that don't send it). Non-boolean values are treated as absent.
- Nonstream responses honor `parallel_tool_calls: false` by truncating to the first call per the OpenAI contract, with a `tool_calls_truncated_parallel_disabled` stats marker so the truncation is visible.

Four regression tests: declared-field-wins in both directions, heuristic fallback preserved, nonstream truncation, non-boolean coercion. Full suite: 1935 passed, 5 skipped; ruff clean. Independent branch off main.

### DIFF
--- a/mtplx/server/omlx_bridge/adapter.py
+++ b/mtplx/server/omlx_bridge/adapter.py
@@ -65,6 +65,14 @@ def _content_to_text(content: Any) -> str:
 
 
 def _try_parse_json(value: Any) -> dict[str, Any]:
+    """Parse history tool-call arguments to a dict, or fail loudly.
+
+    The chat path strict-validates arguments before this adapter runs
+    (_template_tool_call), so malformed input here means a caller bypassed
+    that contract. This used to fall back to {} silently, which erased
+    tool-call arguments from rendered history with no trace (issue #170's
+    class of failure); an invariant violation must raise, not degrade.
+    """
     if isinstance(value, dict):
         return value
     if value in (None, ""):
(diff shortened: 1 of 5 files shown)


PhilipJohnBasile:
Note on merge order: this branch is now stacked on #189 (the two PRs edit adjacent regions of openai.py, so independent branches would conflict). Merge #189 first — this PR's diff shrinks to its own commit automatically. Verified locally that #189 → #190 → #191 → #187 → #188 all merge cleanly in sequence with the full suite green on the combined tree (1962 passed).

youssofal:
Adopted — the titular commit is cherry-picked into our 2.3.0 release lane with your authorship preserved (Honor parallel_tool_calls instead of deciding by client-name sniffing), plus two additions on our side: the truncation stat is wired into the dashboard latest-metrics merge list, and an end-to-end streaming test in the real Android Studio payload shape (tools + stream + stream_options + parallel_tool_calls:false) pins that the declared field alone drives the behavior. That client really does send parallel_tool_calls:false today and was being ignored — good find. The degradation-telemetry commit this branch shares with #189/#191 was excluded per the notes on those PRs (our #170 fix supersedes its detection points). Closing since the work ships with 2.3.0, credited in the changelog.

Issue #121, opened 2026-06-25
Prompt cache can't extend across tool-call turns (tool_call_history_rewrite) → agentic workloads re-prefill every turn


mmmugh:
## Summary

For multi-turn **tool-using** (agentic) workloads, the SessionBank caches the prefix up to the **first assistant tool call** and never extends past it — so every subsequent turn re-prefills the growing suffix. On a real coding-agent session this means ~60% of all prompt tokens get re-prefilled, and per-turn latency is dominated by prefill instead of decode (which defeats MTP's decode speedup for exactly the workload agents care about).

I think this is *intentional* (there's a clear correctness guard), so I'm opening this as a question about the design rather than a drive-by PR.

## Repro

Run any tool-using agent (Pi / opencode / cline) against an MTPLX serve and watch the per-request cache telemetry over a multi-turn session. `cached_tokens` freezes at the pre-first-tool-call prefix; the postcommit log shows, every turn:

```
idle async session postcommit: mode=aborted, stored=false,
   reason="foreground_preempted_postcommit",
   unsafe_reason="tool_call_history_rewrite",  best_prefix_len=<frozen base>
```

## Evidence

Qwen3.6-27B (8-bit, `--mtp`), one autonomous coding-agent session, prefill-skip% = `cached_tokens / prompt_tokens` per request:

| | plain-append (no tools) | tool-using agent |
|---|---|---|
| turn 2+ skip% | **~99.96%** | starts ~90%, **decays to ~24%** as context grows |
| token-weighted skip% | ~100% | **40%** (60% of all prompt tokens re-prefilled) |
| per-turn latency | ~0.5 s | 75–250 s (re-prefilling 15–35k tokens) |

The cache *works* for plain appends; it's specifically tool-call turns that can't be stored.

## Root cause (as I read it)

`mtplx/server/openai.py` `_generation_final_postcommit_compatibility()`:

```python
if assistant_tool_calls:
    return {"safe": False, "mode": "unsafe", "reason": "tool_call_history_rewrite"}
```

Unconditional — any turn with assistant tool calls is deemed unsafe to store, so the cache can't grow past it. I believe the reason is sound: the model **generates** `<tool_call>…</tool_call>` as text, but the next turn re-serializes it into a structured tool-call message + tool result that **tokenizes differently**, so the cached KV wouldn't match the re-rendered prefix. Storing it would risk serving stale/incorrect state.

[...]

PR #191, opened 2026-07-20
Reject unsupported OpenAI options (n, logprobs, logit_bias, stream_options) with clear 400s

PhilipJohnBasile:
Same class of gap as `response_format` before #186: the request models use `extra="allow"`, so several standard OpenAI fields were accepted and silently discarded. A client sending `n: 4` got one choice; `logit_bias` got unbiased sampling; `logprobs` got nothing — all with a 200 and no hint. Silently ignoring a field the client is entitled to rely on is worse than refusing it.

- `n`, `logprobs`, `top_logprobs`, `logit_bias` are now declared on the chat model (`n`, `logprobs`, `logit_bias` on completions) and validated: unsupported values get a 400 with a one-line explanation. Benign values (`n: 1`, `logprobs: false`, empty `logit_bias`) stay accepted so existing clients don't break.
- `stream_options` is validated (unknown keys 400; `include_usage` must be boolean). `include_usage` is genuinely honored — the streaming final chunk already carries `usage`, and a regression test now pins that behavior. (MTPLX sends usage in streams regardless of the flag; that's a strict superset of the OpenAI contract and matches what the #121 thread asked for.)

16 regression tests. Full suite: 1945 passed, 5 skipped; ruff clean. Independent branch off main; pairs with #190 (`parallel_tool_calls`) as the "no silently-ignored API surface" cleanup.

### DIFF
--- a/mtplx/server/omlx_bridge/adapter.py
+++ b/mtplx/server/omlx_bridge/adapter.py
@@ -65,6 +65,14 @@ def _content_to_text(content: Any) -> str:
 
 
 def _try_parse_json(value: Any) -> dict[str, Any]:
+    """Parse history tool-call arguments to a dict, or fail loudly.
+
+    The chat path strict-validates arguments before this adapter runs
+    (_template_tool_call), so malformed input here means a caller bypassed
+    that contract. This used to fall back to {} silently, which erased
+    tool-call arguments from rendered history with no trace (issue #170's
+    class of failure); an invariant violation must raise, not degrade.
+    """
     if isinstance(value, dict):
         return value
     if value in (None, ""):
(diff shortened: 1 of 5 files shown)


PhilipJohnBasile:
Note on merge order: stacked on #190 (adjacent openai.py regions). Merge #189 → #190 first and this diff reduces to its own commit. The full five-PR chain (#189/#190/#191/#187/#188) merges cleanly in that order; combined-tree suite: 1962 passed, 5 skipped.

youssofal:
Appreciate the spec rigor, but we're not taking this direction. Our compatibility policy for benign-but-unsupported options is deliberately ignore-with-telemetry (the existing client_*_fields_ignored regime): our most-used client sends stream_options on every request today, and an unknown-key 400 there is a pre-armed outage the day any SDK grows that object — a camelCase variant already exists in the wild. There's also a concrete break in the current diff: on /v1/completions a non-dict stream_options reaches set()/.get and 500s. Note this branch also contains #190 wholesale plus an undisclosed telemetry commit — please keep PRs independently mergeable in future. Thanks for the test matrix; the accept/reject boundary cases were genuinely useful.

PR #195, opened 2026-07-22
67.6tps Laguna S-2.1 AR support (mlx-community oQ4e)

davidtai:
## What changed

Adds target-only AR support for the exact `mlx-community/Laguna-S-2.1-oQ4e` checkpoint at revision `8e3f5cad513746264940c1c4195de48d7ea345a5`. The earlier `pipenetwork/Laguna-S-2.1-MLX-4bit` pin is now rejected like any other unadmitted variant.

- bundles the Apache-2.0 Laguna MLX model implementation with construction-time geometry checks
- pins the Hugging Face revision and verifies all 13 shard sizes plus exact config, index, tokenizer, generation config, special tokens map, and Poolside template hashes (7 sidecar sha256s)
- installs a dedicated `laguna_ar` runtime with native Laguna KV caches, cache-only prefill, final-logit slicing, and no Qwen optimization hooks or hot-path fallback
- rejects MTP before weight loading because this checkpoint has no native MTP head
- owns Laguna sampler, reasoning, Poolside native tool protocol, 32K context/response defaults, integration budgets, and memory/disk preflights through the backend descriptor
- updates CLI/server inspection, docs, dashboard parser controls, and tracked dashboard assets
- handles the oQ4e export layout specifically: every tensor sits under a `language_model.` prefix, which the runtime strips from both weights and the quantization map so the vendored module tree (no such wrapper) matches by path; the BF16 router lives in a `gate.proj` submodule with the load-balancing bias parked under `gate`
- quantization is a mixed-precision imatrix map read directly from the checkpoint's config: 4-bit/gs128 base (MoE experts), 8-bit `lm_head` and embeddings at gs64, 5/6-bit layer-0 dense MLP at gs64, per-layer self-attention overrides (mostly 5-bit with several layers promoted to 8-bit), and unquantized BF16 routers
- `Model.sanitize()` folds the `language_model.` layout onto the module tree, and the runtime feeds the prefix-stripped quant dict through `load_model(model_config=...)` so mlx-lm's config-driven quantizer addresses module paths directly; the old hard-coded 8-bit-router `quant_predicate` was dead code under mlx-lm 0.31.3 and is removed
- `generation_config.json` and `chat_template.jinja` are byte-identical to the previous pin, so reasoning/tool defaults carry over unchanged

- ports the batched AR decode lane (ragged-KV batched driver) and opens `decode_mode="ar"` to target-only runtimes: neither prefill nor the AR loop requests hidden states, and no MTP head is required
- fixes batched single-token rope: `mx.fast.rope`'s single-step fast path writes only batch row 0 when `T==1` and the offset is scalar, so every batched decode row past the first came back as uninitialized buffer; a per-row offset vector routes the call to the general kernel (this hits the 36 sliding-window layers, whose caches keep scalar offsets whenever context exceeds the window)
- adds seven env-gated fused decode paths, all default-off, with the per-path measurements in the "Decode kernel stack" section below and CPU contracts in `tests/test_laguna_fused.py`

- adds a compile-able single-token decode step (`mtplx/laguna_compiled_step.py`): the B=1 step re-expressed as a pure function over explicit tensor cache leaves so `mx.compile` traces it once and replays it from the C++ runtime — measurements in the "Compiled decode step" subsection below

## Why

The checkpoint is not loadable through the current pinned `mlx-lm` model registry, and MTPLX previously treated runnable models as MTP-only. Laguna also has different attention-head geometry, RoPE, router, cache, and Poolside tool-protocol requirements, so it needs an exact target-owned route rather than a topology transplant or silent stock fallback.

The admitted artifact moved from the uniform-4bit `pipenetwork` build to the `mlx-community` oQ4e mixed-precision imatrix export because the export layout differs mechanically — the `language_model.` prefix and the per-path quantization dict cannot be expressed by a hard-coded predicate — so the runtime now reads the checkpoint's own quantization config instead of assuming a fixed bit-width scheme.

## How to verify

```bash
PYTHONPATH="$PWD" python -m pytest \
  tests/test_laguna_model.py tests/test_artifacts.py tests/test_hf_loader.py \
  tests/test_metal_memory_caps.py tests/test_reasoning_stream_split.py \
  tests/test_generation_sustained.py tests/test_public_cli.py \
  tests/test_server_openai.py -q

ruff check mtplx tests/test_laguna_model.py tests/test_artifacts.py \
  tests/test_hf_loader.py tests/test_metal_memory_caps.py \
  tests/test_public_cli.py tests/test_reasoning_stream_split.py \
  tests/test_server_openai.py tests/test_generation_sustained.py

cd dashboard
npm run typecheck
npm run build
```

CI runs the no-mlx-installed lane as a single pytest invocation; the affected files are all in that list:

```bash
python -m pytest tests/test_no_mlx_imports.py tests/test_public_cli.py \
  tests/test_runtime_kpis.py tests/test_server_openai.py \
  tests/test_openai_bridge.py tests/test_thermal.py tests/test_cache_state.py
```

Pinned artifact validation:

```bash
mtplx pull mlx-community/Laguna-S-2.1-oQ4e --progress-json
mtplx inspect mlx-community/Laguna-S-2.1-oQ4e --json
```

Guarded real-checkpoint hosting and generation completed against the oQ4e pin; see Benchmarks below.

## Benchmarks

Measured 2026-07-22 on an M5 Max (128 GiB unified, 614 GB/s) inside an exclusive guarded window (no other model processes). Server benchmarks ran against `mtplx serve --no-mtp` on localhost via the OpenAI-compatible endpoint, serial requests, `temperature 0`; decode rates use two-point timing (`max_tokens 1` vs `max_tokens N` on the identical prompt) so prefill and connection overhead cancel out.

| Measurement | Prompt tokens | Generated | Result |
|---|---|---|---|
| Prefill (server) | 690 | — | 513.8 tok/s |
| Prefill (server) | 14,872 | — | 698.9 tok/s |
| Decode (server) | 690 | 780 (natural EOS before the 1024 cap) | **52.3 tok/s** |
| Decode at ~15K context (server) | 14,872 | 256 | 47.3 tok/s |
| Decode (`mtplx run --json` runtime receipt) | short prompt | 291 | 55.3 tok/s |

Speculative-decoding stats in every receipt are structurally zero — `generation_mode="ar"`, `mtp_depth 0`, `verify_calls 0`, `accepted_by_depth []` — confirming the target-only AR route: one token per forward pass, no draft/verify cycle exists for this checkpoint.

Two caveats frame these numbers. First, they predate the decode kernel stack below: `laguna_ar` runs stock kernels by default (no tuned kernel routes, no prefill-layout work; `tune_policy` is disabled for this backend by design). Second, the checkpoint ships no MTP head, so the decode figures above are the plain-AR floor — speculative decoding, MTPLX's primary decode lever, is unavailable until an MTP-equipped Laguna export exists. Both are properties of the artifact and the deliberately exact-scope route, not measurement conditions.

## Decode kernel stack (env-gated, measured)

Added 2026-07-24 on top of the AR route above. All paths are opt-in via environment variables; default behavior is unchanged unless a flag is set. Each path ran as an arm inside a guarded GPU window (M5 Max, ctx 1024, greedy, 96 decode steps) against an in-window stock arm, with a repeated stock cell in the same window bounding drift. Greedy-token identity between fused and stock arms was gated at B=1.

| Path (env var) | What it fuses | Exactness | B=1 solo | B=2 solo |
|---|---|---|---|---|
| `MTPLX_LAGUNA_KERNEL_ATTN_GATE` | Per-head attention gate: softplus + broadcast-multiply chain in one Metal kernel | Bit-exact | +1.68% | +0.46% |
| `MTPLX_LAGUNA_KERNEL_ROUTER` | MoE router epilogue: sigmoid → bias → top-k → normalize → scale in one kernel, row-gated to ≤4 rows | Identical expert set; weight reassociation bounded at ~1e-8 | +2.25% | +2.51% |
| `MTPLX_LAGUNA_FUSED_GATE_UP` | Expert gate/up projections concatenated into one gather matmul | Bit-exact (quantization groups run along the input dim, so concatenating output rows preserves per-row scales/biases) | +1.94% | +1.18% |
| `MTPLX_LAGUNA_KERNEL_QK_ROPE` | q/k RMSNorm + transpose + rope (YaRN partial-rotary on full-attention layers, plain rope on sliding layers) for q and k in one dispatch per layer | Bit-exact (reproduces the stock kernels' exact lane layout, rounding order, and angle math) | -0.23% | +1.11% |
| `MTPLX_LAGUNA_KERNEL_COMBINE` | MoE weighted expert combine + shared-expert add in one dispatch, eliminating the `[rows, top_k, hidden]` intermediate | Bit-exact (reproduces the strided-reduction accumulation order) | +0.03% | +2.03% |
| `MTPLX_LAGUNA_FUSED_SHARED_GATE_UP` | Gate/up concatenation applied to the shared experts and the dense layer-0 MLP | Bit-exact | -0.01% | +0.15% |
| `MTPLX_LAGUNA_CACHED_LHS` | Cached gather `lhs_indices` + constant router-bias cast, removing ~190 dispatches/step | Value-identical inputs; the explicit index array can select a different gather kernel tiling at prefill, which was observed to flip greedy near-ties | neutral eager; ~-0.2 ms inside the compiled lane | neutral |
| `MTPLX_LAGUNA_FUSED_QKVG` | q/k/v/gate projections concatenated into one quantized matmul per attention block | Projection outputs bit-exact; the wider matmul changes prefill kernel selection, which was observed to flip greedy near-ties when composed with the full stack | neutral (the four projections already execute concurrently) | neutral |
| `MTPLX_LAGUNA_KERNEL_ROUTER_GEMV` | The router's BF16 gemv folded into the routing path: a simdgroup-per-expert logits kernel feeding the fused top-k kernel — two dispatches replace gemv + cast + eleven-op epilogue | Logits within one bf16 ulp of the stock gemv (~1 in 12k values); zero selection flips observed; kept the stock token stream in the exact composition's measured windows | -0.5 ms inside the compiled lane (neutral eager) | untested |

*Solo figures come from two windows: Window A (stock B=1 18.18 ms/step; repeated-stock drift −0.4% at B=1, −0.32% at B=2) covers `KERNEL_ATTN_GATE`, `KERNEL_ROUTER`, `FUSED_GATE_UP`. Window B (stock B=1 17.92 ms/step; drift +1.0% at B=1, −0.68% at B=2) covers `KERNEL_QK_ROPE`, `KERNEL_COMBINE`, `FUSED_SHARED_GATE_UP`, `CACHED_LHS`.*

**Stacked**

Two stacked configurations were measured in Window B against the same in-window stock arm. The all-reversible stack combines `KERNEL_ROUTER` + `KERNEL_ATTN_GATE` + `KERNEL_QK_ROPE` + `KERNEL_COMBINE`. The candidate-best stack adds `FUSED_GATE_UP` + `FUSED_SHARED_GATE_UP` on top of that set. Both produce token sequences identical to stock at B=1.

Per cell: speedup vs in-window stock, ms/step, then per-prompt and raw (aggregate) throughput.

| Stack | B=1 | B=2 | B=4 |
|---|---|---|---|
| stock (baseline) | 17.92 ms — 55.8 per-prompt / 55.8 raw tok/s | 25.52 ms — 39.2 per-prompt / 78.4 raw | 37.13 ms — 26.9 per-prompt / 107.7 raw |
| all-reversible | +4.78% — 17.07 ms — 58.6 per-prompt / 58.6 raw | +5.50% — 24.12 ms — 41.5 per-prompt / 82.9 raw | +5.55% — 35.07 ms — 28.5 per-prompt / 114.1 raw |
| candidate-best | +7.44% — 16.59 ms — **60.3 per-prompt** / 60.3 raw | +7.24% — 23.67 ms — **42.3 per-prompt** / 84.5 raw | +5.04% — 35.26 ms — 28.4 per-prompt / 113.4 raw |

A second confirmation window anchored candidate-best B=1 in the 16.55–16.73 ms range.

Solo deltas at B=1 sit inside the launch-overlap shadow of the large kernels, so small-kernel dispatch removal only registers once several paths run in combination — the stack total exceeds the sum of the solo deltas. A residual+norm fusion path also exists but stays off by default: a single fused custom-kernel launch costs more host-side than the two stock ops it replaces.

### Compiled decode step (tensor-leaf caches)

`mtplx/laguna_compiled_step.py` re-expresses the B=1 decode step as a pure function over explicit tensor state leaves: fixed-capacity full-attention KV written at a traced offset with an in-graph additive mask, and the sliding windows as ring buffers written at a traced slot index — unmasked at steady state, since every slot is live once context exceeds the window and softmax is permutation-invariant over the ring. That makes the whole step `mx.compile`-able: the graph is traced once and replayed from the C++ runtime, removing the per-token python graph reconstruction (~1,600 op builds/step). The lane seeds from the eager caches after a normal prefill; the supported regime is context ≥ sliding window.

One guarded window (M5 Max, ctx 1024, greedy, 96 measured steps, B=1; stock anchors 18.07/18.13 ms bracketing the arms):

| Configuration | ms/step | tok/s (per-prompt = raw at B=1) |
|---|---|---|
| stock python-cache lane | 18.00 | 55.6 |
| tensor-leaf lane, eager | 17.82 | 56.1 |
| tensor-leaf lane, compiled | 16.12 | 62.0 |
| stock lane + kernel stack | 16.54 | 60.5 |
| compiled + kernel stack | **15.73** | **63.6** |

A second window reproduced the compiled + kernel stack cell at 15.55 ms (64.3 tok/s; anchors 17.79/18.09 ms), again with one shared token digest across every lane. The fused q/k norm+rope kernel also runs inside the compiled graph (traced position input) — token-safe, measured speed-neutral there: the dispatches it removes were already inexpensive inside the compiled graph.

**Choosing a configuration.** Two vetted stacks, differing only in whether the greedy token stream is identical to stock:

| Configuration | Env on top of the compiled lane | ms/step | tok/s (B=1) | Output contract |
|---|---|---|---|---|
| exact | GATE_UP + SHARED_GATE_UP + KERNEL_ROUTER + KERNEL_ATTN_GATE + KERNEL_QK_ROPE + KERNEL_COMBINE + KERNEL_ROUTER_GEMV | **14.84** | **67.4** | Greedy tokens identical to stock in every measured window (in-window stock reference; the router kernels' bounded reassociations are present but altered no token) |
| inexact | exact + CACHED_LHS + FUSED_QKVG | **14.83** | **67.4** | Value-safe: identical arithmetic inputs, but different prefill kernel selection flips occasional greedy near-ties; within-configuration stock/eager/compiled lanes agree exactly |

Without the router-gemv fold the same stacks measure 15.53 / 15.18 ms; the exact figure includes the strided-GLU activation kernel on the dense gate/up paths (bit-exact, source-matched rounding). Sizing the compiled lane's full-attention capacity to the workload (constructor argument; 1152 for this 1024+decode shape instead of the default 2048) measures 14.80 ms / 67.6 tok/s with the same stock-identical digest. A packed k/v leaf layout was also built and measured ~0.07 ms slower than unpacked in-model, so it ships default-off. This exact configuration is the shipped result of the optimization pass: 17.9-18.2 ms stock to 14.84-14.80 ms, +21.6%, greedy tokens identical to stock in every measured window. The fold's token-stream preservation in the exact stack is a measured property of its rounding contract (the dot is rounded to bf16 before the sigmoid, matching the stock gemv's own output precision), not a bitwise guarantee — a logit within one bf16 ulp of a routing tie could still flip; none did across the measured windows.

Both configurations pass every within-arm identity gate (python-cache, eager tensor-leaf, and compiled lanes produce one shared digest). The inexact stack's divergence from stock is the same benign class as batched-vs-solo non-invariance: equally valid float reassociation, not a precision change.

Greedy tokens were identical across every lane and configuration in the window — one shared digest, including compiled-vs-stock. Inside the compiled lane the router, attention-gate, and combine kernels plus both gate/up concatenations engage; the q/k norm+rope fusion is bypassed because the compiled step's own attention covers that chain (compile fuses it). CPU contracts live in `tests/test_laguna_compiled_step.py` (18 tests: ring-write equivalence against `RotatingKVCache` across wraps, snapshot aliasing, compiled-vs-eager equality, quantized/YaRN/tied-embedding variants).

## Notable decisions

- Support is deliberately exact-checkpoint-only. Other Laguna variants remain blocked, including the earlier `pipenetwork` build, which is now rejected the same way any unadmitted variant is.
- MTP is rejected explicitly; there is no enabled-lane fallback.
- The 32K default is a resource-safe operating default, while the model maximum remains 1,048,576 and larger server contexts must pass the active Metal-memory budget.
- Weights are 64,122,027,323 bytes (59.72 GiB); the full pinned snapshot is 64,129,728,868 bytes (64.13 GB). The resident floor at the 32K default context is 74,398,072,123 bytes (~69.3 GiB), plus a 16 GiB system reserve; the docs round this up to the practical tier: at least 96 GiB unified memory, 128 GiB recommended.
- The test-env isolation fix (turned CI green): the laguna one-shot defaults test drives `_generate_one_shot_public` with the sustained profile, which writes roughly three dozen `MTPLX_*` keys into process env. Under CI's single-process pytest invocation those keys leaked into later tests, and leftover `MTPLX_VLLM_METAL_PAGED_*` keys flipped env-driven cache selection and failed `test_cache_state.py`. The test now snapshots and restores `os.environ` around itself so the profile application stays observable but process-local state doesn't leak.

### DIFF
--- a/mtplx/runtime.py
+++ b/mtplx/runtime.py
@@ -716,11 +716,12 @@ def load(
             adapter_merge_report = merge_installed_mtp_lora_adapters(model)
     elif merge_mtp_adapter:
         raise RuntimeError("merge_mtp_adapter requires mtp_adapter")
+    fused_report: list[dict[str, Any]] = []
     if _is_laguna_s_2_1_mlx_4bit_config(config):
         # Env-gated fused decode paths (MTPLX_LAGUNA_*): with no switches set
         # this returns an empty report and changes nothing, so default serving
         # behavior is untouched; a serving wrapper that exports the measured
-        # stack gets it engaged at load, visible in this log line.
+        # stack gets it engaged at load.
         from .models.laguna_fused import install_from_env as _laguna_install_fused
 
         fused_report = _laguna_install_fused(model)
@@ -759,6 +760,9 @@ def load(
         )
         runtime.a3b_whole_moe_installed = True
         logger.info("[a3b-whole-moe] %s", whole_moe_report)
+    # The server prints this as its startup engagement receipt; logger.info
+    # alone is invisible under `python -m mtplx.server.openai` (no handler).
+    runtime.laguna_fused_report = fused_report
     return runtime
 
 
--- a/mtplx/server/omlx_bridge/tool_calling.py
+++ b/mtplx/server/omlx_bridge/tool_calling.py
@@ -474,8 +474,10 @@ def parse_tool_calls(
         cleaned, calls, malformed = _parse_xml_tool_calls(cleaned_text)
         filtered_calls = _filter_known_tools(calls, tools)
         if calls and not filtered_calls and tools:
-            first_name = calls[0].get("function", {}).get("name", "unknown")
-            malformed = f"unknown tool '{first_name}'"
+            # OpenAI passes unknown-named calls through; the client owns the
+            # rejection (and answers the model, letting it self-correct).
+            # Degrading the whole turn to prose costs the agent a strike.
+            filtered_calls = calls
         if filtered_calls:
             return ToolCallExtraction(
                 cleaned_text=cleaned,
--- a/mtplx/server/omlx_bridge/tool_calling.py
+++ b/mtplx/server/omlx_bridge/tool_calling.py
@@ -243,6 +243,47 @@ def _parse_xml_tool_calls(text: str) -> tuple[str, list[dict[str, Any]] | None, 
                 break
             calls.append(_tool_call(name, params))
             continue
+        # Poolside dialect: `name<arg_key>k</arg_key><arg_value>v</arg_value>...`
+        # (the Laguna family's native emission; same contract as the strict
+        # parser in openai.py — every non-pair byte after the name is residue
+        # and marks the call malformed rather than silently dropping args).
+        poolside_match = re.match(r"\s*([A-Za-z_][\w.-]*)", content)
+        if poolside_match:
+            poolside_name = poolside_match.group(1)
+            body = content[poolside_match.end():]
+            pairs = list(
+                re.finditer(
+                    r"<arg_key>\s*(.*?)\s*</arg_key>\s*<arg_value>\s*(.*?)\s*</arg_value>",
+                    body,
+                    re.DOTALL,
+                )
+            )
+            if pairs:
+                params = {}
+                residue_parts = []
+                cursor = 0
+                valid = True
+                for pair in pairs:
+                    key = pair.group(1).strip()
+                    if not key or key in params:
+                        valid = False
+                        break
+                    value = pair.group(2).strip()
+                    try:
+                        params[key] = json.loads(value)
+                    except (TypeError, ValueError):
+                        params[key] = value
+                    residue_parts.append(body[cursor:pair.start()])
+                    cursor = pair.end()
+                residue_parts.append(body[cursor:])
+                if valid and not "".join(residue_parts).strip():
+                    calls.append(_tool_call(poolside_name, params))
+                    continue
+                malformed_reason = (
+                    f"tool '{poolside_name}' contains malformed Poolside arguments"
+                )
+                calls = []
+                break
         malformed_reason = "unrecognized <tool_call> payload"
     if not calls:
         return text, None, malformed_reason
--- a/mtplx/server/openai.py
+++ b/mtplx/server/openai.py
@@ -6360,6 +6360,69 @@ def _parse_poolside_tool_call(block: str) -> tuple[str, Any] | None:
     return name, arguments
 
 
+def _scan_bracket_tool_call(
+    text: str,
+    start: int,
+) -> tuple[int, str, Any] | None:
+    """Balanced-scan one `[Calling tool: name({...})]` block at `start`.
+
+    The JSON object is walked with string/escape awareness, so `}` or `)]`
+    sequences inside argument strings (a JS file body, a task_progress
+    checklist) cannot end the block early — the failure mode of the
+    non-greedy regex this replaces. Returns (end_exclusive, name, arguments)
+    for a complete block, or None when no complete well-formed block starts
+    at `start`.
+    """
+    match = re.compile(
+        r"\[(?:Calling tool|Tool call):\s*([A-Za-z_][\w.-]*)\s*\(",
+        re.IGNORECASE,
+    ).match(text, start)
+    if match is None:
+        return None
+    name = match.group(1)
+    i = match.end()
+    tail = re.compile(r"\s*\)\s*\]").match(text, i)
+    if tail is not None:
+        return tail.end(), name, {}
+    if i >= len(text) or text[i] != "{":
+        return None
+    depth = 0
+    in_string = False
+    escaped = False
+    j = i
+    while j < len(text):
+        ch = text[j]
+        if in_string:
+            if escaped:
+                escaped = False
+            elif ch == "\\":
+                escaped = True
+            elif ch == '"':
+                in_string = False
+        elif ch == '"':
+            in_string = True
+        elif ch == "{":
+            depth += 1
+        elif ch == "}":
+            depth -= 1
+            if depth == 0:
+                j += 1
+                break
+        j += 1
+    if depth != 0:
+        return None
+    close = re.compile(r"\s*\)\s*\]").match(text, j)
+    if close is None:
+        return None
+    try:
+        arguments = json.loads(text[i:j])
+    except (TypeError, ValueError):
+        return None
+    if not isinstance(arguments, dict):
+        return None
+    return close.end(), name, arguments
+
+
 def _tool_marker_pairs_from_tokenizer(tokenizer: Any | None) -> list[tuple[str, str]]:
     if tokenizer is None:
         return []
@@ -6391,19 +6454,33 @@ def _iter_generated_tool_call_envelopes(
         )
         for match in pattern.finditer(text):
             envelopes.append((match.start(), match.end(), match.group(1).strip(), None))
-    for match in _BRACKET_TOOL_CALL_RE.finditer(text):
-        name = match.group(1).strip()
-        raw_args = (match.group(2) or "").strip()
-        if raw_args:
-            try:
-                arguments: Any = json.loads(raw_args)
-            except json.JSONDecodeError as exc:
-                raise _tool_protocol_error(
-                    f"bracket tool_call '{name}' arguments are not valid JSON"
-                ) from exc
-        else:
-            arguments = {}
-        envelopes.append((match.start(), match.end(), "", (name, arguments)))
+    # Bracket dialect (`[Calling tool: name({...})]`) via the balanced scanner:
+    # the old non-greedy regex ended the block at the first `})]`, which any
+    # code-file argument contains inside a string, so large bracket calls
+    # always failed JSON decode and fell back to prose.
+    search_from = 0
+    lowered_text = text.lower()
+    while True:
+        candidates = [
+            idx
+            for idx in (
+                lowered_text.find(prefix.lower(), search_from)
+                for prefix in _BRACKET_TOOL_PREFIXES
+            )
+            if idx >= 0
+        ]
+        if not candidates:
+            break
+        found = min(candidates)
+        scanned = _scan_bracket_tool_call(text, found)
+        if scanned is None:
+            # Unterminated or malformed: not an envelope — the text stays
+            # visible content, same terminal state as the old decode error.
+            search_from = found + 1
+            continue
+        end, name, arguments = scanned
+        envelopes.append((found, end, "", (name, arguments)))
+        search_from = end
     envelopes.sort(key=lambda item: item[0])
     for previous, current in zip(envelopes, envelopes[1:]):
         if current[0] < previous[1]:
--- a/mtplx/server/openai.py
+++ b/mtplx/server/openai.py
@@ -305,8 +305,15 @@ STREAM_SILENCE_WARN_INTERVAL_S = 60.0
 STREAM_STALL_DEADLINE_S = float(
     os.environ.get("MTPLX_STREAM_STALL_DEADLINE_S") or 300.0
 )
-STREAM_HIDDEN_TOOL_GUARD_TOKENS = 2048
-STREAM_HIDDEN_TOOL_GUARD_S = 30.0
+# Runaway-hidden-generation backstop. Native-tool agent workloads stream
+# multi-thousand-token arguments (whole files) as legitimate hidden text, so
+# the ceilings are env-tunable; the defaults keep the original chat-UX guard.
+STREAM_HIDDEN_TOOL_GUARD_TOKENS = int(
+    os.environ.get("MTPLX_STREAM_HIDDEN_TOOL_GUARD_TOKENS", "2048")
+)
+STREAM_HIDDEN_TOOL_GUARD_S = float(
+    os.environ.get("MTPLX_STREAM_HIDDEN_TOOL_GUARD_S", "30")
+)
 STREAM_TOOL_CALL_FINISH_GRACE_S = 0.05
 TOOL_PROTOCOL_BOUNDARY_GRACE_S = 0.05
 _REASONING_DETAILS_RE = re.compile(
--- a/mtplx/server/openai.py
+++ b/mtplx/server/openai.py
@@ -1082,6 +1082,15 @@ def _startup_line(text: str = "") -> None:
     _safe_stdout_print(text)
 
 
+def _laguna_fused_startup_line(runtime: Any) -> str | None:
+    """Engagement receipt for the env-gated fused stack (grep: [laguna-fused])."""
+
+    report = getattr(runtime, "laguna_fused_report", None)
+    if not report:
+        return None
+    return "[5/6] [laguna-fused] " + json.dumps(report, ensure_ascii=False)
+
+
 def _startup_server_url(args: argparse.Namespace) -> str:
     return local_url_for_bind(
         str(getattr(args, "host", "127.0.0.1")),
@@ -1765,6 +1774,9 @@ class ServerState:
             _startup_line(
                 f"[5/6] {self.backend_descriptor.display_name} drafter is active"
             )
+        fused_line = _laguna_fused_startup_line(self.runtime)
+        if fused_line:
+            _startup_line(fused_line)
         if self.backend_descriptor.uses_draft_lm_head and self.runtime.mtp_enabled:
             self.draft_lm_head = self.model_scheduler.submit_foreground(
                 _install_draft_lm_head,
--- a/mtplx/server/openai.py
+++ b/mtplx/server/openai.py
@@ -6423,6 +6423,74 @@ def _scan_bracket_tool_call(
     return close.end(), name, arguments
 
 
+def _classify_bracket_tool_call(text: str, start: int) -> str:
+    """Classify the bracket block starting at `start` (which sits on a full
+    `[Calling tool:`/`[Tool call:` prefix).
+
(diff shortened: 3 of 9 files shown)


davidtai:
The following renders show the mlx-profiler dispatch-census command-buffer overlap timeline for the two Laguna S-2.1 decode lanes under comparison — stock python-cache decode and the compiled tensor-leaf lane with the kernel stack — each as a full measured-phase overview (8 synchronize-bracketed decode steps) and a 2-step zoom exposing per-command-buffer structure.

Stock python-cache decode lane, full measured decode phase (8 steps, 1526 command buffers)
Stock python-cache decode lane, 2-step zoom
Compiled + kernel stack decode lane (pre-router-gemv, 15.55 ms era), full measured decode phase (8 steps, 1209 command buffers)
Compiled + kernel stack decode lane (pre-router-gemv, 15.55 ms era), 2-step zoom

| Lane | Dispatches/step | Command buffers/step | GPU busy | GPU wall |
|---|---|---|---|---|
| Stock (python-cache) | ~2,419 | ~145 | 15.7 ms | 17.8 ms |
| Compiled + kernel stack (pre-router-gemv, 15.55 ms era — not the final 14.98 ms configuration) | 1,668 | ~98 | 14.5 ms | 15.4 ms |

Captures are curated to the measured-phase command-buffer window with `metal-dispatch-viz/scripts/curate_trace.py` (referentially closed, hash-verified) before rendering, since the raw stock census is 12.8 MB.

davidtai:
Follow-up to the timeline renders above — the numbers they illustrate have since moved; final shipped figures for this branch (M5 Max, B=1, ctx 1024, greedy, guarded windows, in-window stock anchors, one shared token digest per configuration):

| configuration | ms/step | tok/s | tokens |
|---|---|---|---|
| stock python-cache lane | 17.9–18.2 | 55.2–55.8 | reference |
| kernel stack, eager | 16.5–16.6 | ~60.3 | identical to stock |
| compiled tensor-leaf lane + kernel stack | 15.55–15.73 | 63.6–64.3 | identical to stock |
| + router-gemv fold + strided-GLU (**shipped exact config**) | **14.84** | **67.4** | identical to stock |
| same, with lane capacity sized to the workload (1152 vs default 2048) | **14.80** | **67.6** | identical to stock |
| inexact option (+ cached-lhs + qkvg) | 14.83 | 67.4 | value-safe; prefill near-tie flips |

The two rendered captures correspond to the stock lane (~2,419 dispatches/step, ~145 cbs) and the compiled+kernels lane of the 15.55 ms era (1,668 dispatches, ~98 cbs); the shipped head further folds the router gemv into the routing kernel and the dense-MLP activation into one strided-GLU kernel.

Updated findings since the renders, all measured on this branch:

- The remaining small dispatches are not the cost: removing hundreds of them solo moves the wall by ~zero (they execute concurrently under the serialized matmul chain). The serial-spine chains were folded instead; dependency-DAG analysis shows 57% of commands on one serial chain with a +1.85% reordering ceiling, which is why whole-step ICB replay measures 0.94x.
- The expert gather is bandwidth-healthy: 428 GB/s cold (82% of the machine's measured 538 GB/s streaming ceiling). Earlier "slow gather" readings were cache-served microbenches; the cold-bank, randomized-index methodology is in `bench` as the standard for gather measurements.
- The step is bandwidth-bound at ~5.9 GB/token; the single-prompt AR ceiling at these bytes is ~68–69 tok/s. A verify-cost sweep (T=1/2/4/8 = 20.2/25.8/40.0/65.3 ms) prices a future MTP-equipped export at 100+ tok/s at typical draft-head acceptance.
- A packed k/v leaf layout and prompt-lookup speculative drafting were both built, measured, and rejected on the numbers (packing −0.07 ms the wrong way; lookup drafting ~60 tok/s pooled on real code transcripts — below the AR floor).

Branch state: 26 commits, 112 tests across the three Laguna suites, every configuration digest-gated against stock in its measurement window.

davidtai:
### What the latest 8 commits are for

These bring Laguna serving in line with the reference configuration published in
[Blackwellboy's `laguna-s21-lab`](https://github.com/Blackwellboy/laguna-s21-lab),
which is the operator guide this checkpoint is documented against. On vLLM that
configuration is:

```
--enable-auto-tool-choice
--tool-call-parser poolside_v1
--reasoning-parser poolside_v1
```

plus `"enable_thinking": false` in `chat_template_kwargs`.

MTPLX already had the equivalent surface — `--tool-prompt-mode native` (which the
`laguna_ar` descriptor requires) renders the request's tools through the
checkpoint's own template, and `--reasoning-parser poolside_v1` is wired. What was
missing was the decode half: the checkpoint emits its calls as
`<tool_call>name<arg_key>k</arg_key><arg_value>v</arg_value></tool_call>`, a payload
shape the omlx bridge did not recognise, so every call fell through to
"unrecognized `<tool_call>` payload" and was returned as prose in `content`. The
guide's own check — send a request with a `tools` array and see whether a real
`tool_calls` array comes back — did not pass before these commits.

The eight commits:

| Commit | |
|---|---|
| `2221152` | Parse the Poolside `arg_key`/`arg_value` dialect |
| `ef8403f` | Pass unknown-named tool calls through rather than degrading the turn |
| `f8440e7`, `ea86403` | Make the hidden-tool stream guard ceilings env-tunable (the 2048-token default truncated whole-file tool arguments) |
| `fc7dd74`, `adeebb8` | Rescue the textual `[Calling tool: name({...})]` drift with a balanced-brace scanner, buffered rather than double-delivered while streaming |
| `1ad1c43` | Print the `[laguna-fused]` install report as a startup line — it went through `logger.info`, which has no handler under `python -m mtplx.server.openai`, so the engagement receipt never reached the log |
| `ba8e0c6` | Update two unknown-tool tests that still asserted the pre-`ef8403f` degrade-to-prose contract |

Verified against a live serve on an M5 Max: buffered `tool_calls`, incremental
`delta.tool_calls` while streaming, and a multi-turn tool-result round trip.

Suite at this head: 2167 passed, 1 failed — `test_onboarding.py::test_start_invokes_onboarding_when_no_explicit_flags`, which fails identically at the previous head (`b3dbb97`), so it is not from this work.

One naming note for anyone wiring a client: the guide describes vLLM streaming
reasoning as `delta.reasoning`, whereas this server emits `reasoning_content`. The
input side accepts either spelling on resent assistant turns, so multi-turn
preservation works with both conventions — it only matters if reasoning is enabled,
and this lane runs with it off.

davidtai:
### Rebased onto current `main` — this branch is now only the un-landed delta

`main` already carries this PR's original substance as
`f108318 feat(models): Laguna S-2.1 oQ4e exact-pin AR-only support + compiled tensor-leaf decode (PR #195)`,
so the branch's 28 pre-existing commits were re-applying work that is already
upstream. That is what made it conflict — the same changes on both sides, plus
rebuilt dashboard asset bundles.

The branch is now `main` plus the 8 commits whose content is **not** upstream yet
(the tool-calling dialect work and the fused-stack startup receipt described in the
comment above). Result:

- **36 commits → 8**, 9 files, +578/−55
- conflicts gone: fast-forward from `main`, no merge commit needed
- full suite at this head: **2647 passed, 0 failed** (the `test_onboarding` failure
  that was red on the old head is fixed by `main`)

`f182ac8`'s honest-`finish_reason` work is upstream of these commits and the
tool-call tests pass against it, so the two changes compose rather than collide.

The previous head was `ba8e0c6` if any of it needs recovering, though everything it
carried beyond these 8 commits is already in `main` via `f108318`.

youssofal:
the rebased delta is merged and shipped in 2.4.0 (single curated commit, your authorship): balanced-brace scanner, stream buffering, poolside dialect, unknown-tool pass-through, env guard ceilings, startup line. the double-delivery find was real, any task_progress checklist tripped the old gate. one integration note from the original round still stands: your _apply_runtime_compatibility_mode crashed bare `mtplx start` for verified models (inspect returns compatibility as a string tier), we normalized it in integration, and we rebuilt the dashboard bundles from source. laguna sits in the app catalog with an ar-only flag.

PR #200, opened 2026-07-24
A3B concurrent decode: row-packing aggregate lane + parallel-decoding docs/evals + 6-bit verify kernels (stacked on #174)

davidtai:
# A3B concurrent decode: a row-packing aggregate-throughput lane (stacked on #174)

> **Stacked on #174 (`perf/a3b-decode-stack`).** This branch is the 12 commits on top of #174's
> head. Until #174 merges, GitHub's diff here also shows #174's commits — **review the 12 commits
> from `perf(a3b): Phase-1 multi-stream batched greedy decode driver` onward**, and merge only
> after #174 lands (the diff collapses to just those 12 once it does). With one flagged exception, commits here
> are new files plus additive methods, default-off. The exception: the 6-bit
> verify-kernel commit (`ea89f45`) extends #174's `nax_verify.py` m6 K-split
> kernel with a `bits` parameter — additive, default `bits=4`, and the rendered
> 4-bit Metal source is test-asserted byte-identical to #174's.

**Stacked on #174.** #174 makes *single-stream* decode fast (MTP speculation + six kernel/schedule
wins, ~200 tok/s for one prompt). This PR adds the complementary thing: an *aggregate* lane that
serves many prompts at once. It shares #174's kernels and adds no risk to that work — everything
here is new files plus a few additive methods on `cache_state.py`, all default-off. #174's decode
path is byte-for-byte unchanged when the new modes aren't selected.

Two operating points, one driver:

| lane | who it's for | rows / request | tokens / request / cycle |
|---|---|---|---|
| **spec** (MTP, #174's speculation) | latency — one or few prompts | 2 (draft + verify) | up to 2 if the draft is accepted |
| **AR** (this PR's default aggregate mode) | throughput — many prompts | 1 | exactly 1 |

---

## Why batching helps here — the reasoning, not just the result

Decode is **memory-bandwidth-bound**, not compute-bound. Each step reads the model's weights out of
RAM to produce tokens, and that read is what the step time is mostly spent on. One stream reads
~all the weights to make **one** token. Put B streams through **one** weight read and you make **B**
tokens for the same read — so aggregate throughput ≈ `B / cycle_time(B)`.

For a **dense** model `cycle_time` barely grows with B until you run out of compute, so you'd batch
until the GPU's FLOPs saturate. A3B is a **sparse Mixture-of-Experts**, which bends the story: each
token routes to its own top-8 of 256 experts, so as B grows **more distinct experts** get pulled in
and the expert bytes read per cycle *grow* with B — until every expert is resident (saturation),
past which extra streams cost **no** additional expert bytes. So the curve has three regimes:

```
aggregate tok/s
   ^
   |                                        ____________  compute roofline (GPU saturated)
   |                                   ____/               ms/token hits its floor
   |                             _____/
   |                       _____/   <- experts saturate: extra rows ~free on bytes
   |                 ____/
   |            ___/   <- expert working set still expanding (bytes grow with B)
   |        __/
   |     _ /
   +----+----+----+----+----+----+----+----+---------> batch size B
        (bandwidth-bound)              (compute-bound)
```

The peak is a **bandwidth / working-set** limit reached *before* compute saturation — it is **not**
"fill all the FLOPs." Concretely, because every stream emits exactly one token per cycle,
**aggregate = 1000 / (ms-per-token)**, so the ceiling is simply wherever ms-per-token bottoms out
(the compute-bound floor).

### How the batch sizes were chosen — and a correction

I started with a doubling ladder (1, 2, 4, 8, 16). That is a cheap way to see the *shape* of the
scaling curve — each doubling tells you whether you're still amortizing — but powers of two are a
**convention, not a physics boundary**. The spec lane caps at B=8 (2 rows × 8 = the 16-row kernel
budget), and AR reaches B=16 at the same budget.

Then a subtle trap: a first wide probe (B=24) came in *below* B=16, and I initially read that as
"16 is the peak." It was not — that probe changed **two** variables at once. It put 24 real streams
in a **32-slot cohort** (8 dummy slots burning compute) *and* crossed the row-count where the tuned
kernels stop. Fixing both — each B measured at **its own cohort** (no dummy waste), on **one
consistent stock kernel config** across the whole range — the curve keeps climbing straight through
16 with nothing special happening there:

```
  clean stock-kernel sweep, each B at its own cohort (code prompts, 256 new tokens, full occupancy)
  B          8    16   24   32   48   64   96   128  192  256
  aggr tok/s 374  479  563  612  649  696  753  758  800  813   <- still rising at 256
  ms/token  2.66 2.08 1.77 1.63 1.53 1.43 1.28 1.31 1.25 1.22   <- knee ~96-128, slow tail
```

**16 was a kernel-lane artifact, not a throughput peak.** (Details on the kernels below.) The true
ceiling is the compute roofline, located by sweeping multiples of 8 until ms/token stops falling —
measured tables at the end.

### Is 16 special? It's a specific kernel width

The number 16 comes from a family of **hand-written Metal kernels dimensioned for 1–16 rows**
("M1–M16"), used by #174's tuned path:

- the **row-owned MoE router** (`qwen_row_owned_router.py`, `_EXACT_ROWS = 1..16`) — grid-dynamic,
  so it *structurally* runs at any width but is only validated to 16;
- the **NAX verify matmul** (`nax_verify.py`, `m16_nax_eligible: 1 <= M <= 16`) — a fixed 16-row
  tile; above 16 it's ineligible and falls back to stock `mx.quantized_matmul`;
- the **combine tail**, coupled to the router.

16 also happens to equal the spec lane's ceiling (8 requests × 2 rows) — which is exactly why "the
peak is at 16" looked convincing and was a coincidence. Above 16 rows the forward runs on stock MLX
kernels (generic, any width). Stock is slightly slower per token than the tuned kernels, so a wide
sweep is run entirely on stock for a fair, discontinuity-free curve; whether a *tuned* wide kernel
(M32/M48/…) is worth writing depends on that curve's shape (it rises → yes).

---

## The construction: a stack of small, verifiable layers

Each commit is one idea, each is bitwise-gated against the layer below it. Bottom to top:

```
  #174 head (perf/a3b-decode-stack)
   │
   ├─ 1. Phase-1 driver        multi-stream [B,·] greedy decode + per-stream sha gate
   ├─ 2. force-unsorted pin     make the MoE gather order content-independent (gate can pass)
   ├─ 3. Build-1 cohort loop    FIXED-SHAPE cohort + one device-side sync/cycle
   ├─ 4. R1 ragged KV           per-row offsets + scatter writes (opt-in cache)
   ├─ 5. R2 ragged attention    per-row causal mask builder
   ├─ 6. R3 host capacity bound  no device read on the hot path
   ├─ 7. R3 masked restore       per-row recurrent-state rewind
   ├─ 8. R3 fold-in loop         reject = replay one row back, no separate repair forward
   ├─ 9. tax fix                 drop the per-cycle recurrent-state clone
   ├─ 10. refill / admission      continuous batching: a finished slot re-admits the next request
   ├─ 11. AR row-packing lane     decode_mode=ar: [B,1], the aggregate lane
   └─ 12. tight frozen KV cap     correctness-neutral memory trim
```

### How the pieces work

**Fixed-shape cohort (the correctness backbone).** MLX picks different matmul kernels at different
row counts, and the MoE's token-sort changes float accumulation with content. So bitwise
"batched == single-stream" is unattainable across *different* shapes. The fix: decode always runs a
**constant** `[cohort, ·]` shape, padding empty slots with dummies; the reference for stream *b* is
that same prompt **alone in an identically-shaped cohort**. Same shape → same kernels → a bitwise
match rests only on per-row independence, which the force-unsorted pin guarantees.

```
  cohort of N slots (fixed shape every cycle):
  ┌────┬────┬────┬────┬────┬────┬────┬────┐
  │ r0 │ r1 │ r2 │ r3 │ ·· │ ·· │dum │dum │   real streams fill slots; dummies pad the rest
  └────┴────┴────┴────┴────┴────┴────┴────┘   shapes never change as load varies
   gate: stream r2 here  ==  r2 alone in an N-slot cohort   (byte-for-byte)
```

**Ragged KV + fold-in (spec lane).** Under speculation a stream can be one token ahead or behind its
neighbors, so each row keeps its **own** KV offset and an explicit per-row causal mask. A rejected
draft doesn't trigger a second "repair" forward — the row simply **replays one position back** in
the next shared forward (fold-in), with its recurrent state rewound per-row. One weight read per
cycle, always.

**Refill = continuous batching.** Slots don't drain and wait. When a stream finishes, its slot
re-admits the next queued request at the following cycle boundary, through one cohort-shaped
prefill: the admitted rows start from offset 0 over zeroed recurrent state, every other row's state
is restored and its offsets put back, and the KV capacity is frozen constant so shapes never shift.

```
  time ──▶
  slot0: [req A decoding.....done][req D decoding.........]
  slot1: [req B decoding...done  ][req E decoding.........]
  slot2: [req C decoding.........................done     ][req F ...]
                     ▲ finished slot re-admits the next request at the cycle boundary
```

**AR row-packing (this PR's default aggregate lane).** Plain autoregressive: one `[B,1]` forward,
argmax, one token per stream — **no draft, no verify, no MTP**. That's the whole point. Speculation
spends 2 rows per request; plain AR spends 1, so the same row budget holds **twice** the requests:

```
  16-row budget:

  SPEC (MTP):  r0    r1    r2    r3    r4    r5    r6    r7        8 requests
               [dv]  [dv]  [dv]  [dv]  [dv]  [dv]  [dv]  [dv]      each: 1+accept tok/cycle
               └2┘   └2┘                                          aggregate ≈ 8 × 1.55 ≈ 12 tok/cycle

  AR:          0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15               16 requests
               [x][x][x][x][x][x][x][x][x][x][x][x][x][x][x][x]    each: exactly 1 tok/cycle
                                                                   aggregate = 16 × 1 = 16 tok/cycle
```

At the measured acceptance (~0.5–0.6), AR's 16 tokens/cycle beats spec's ~12 — AR wins **because**
it drops MTP. MTP/spec isn't wasted; it's the right lane when latency per request matters and only a
few streams are in flight. The two are complementary SKUs, and #174's speculation is what makes the
spec SKU good.

**The decode loop (both lanes).** One blocking sync per cycle. The argmax that feeds the next cycle
is ready at submit time, the forward runs asynchronously, and the loop keeps exactly one forward in
flight — so the host never stalls the GPU:

```
  cycle t:   submit forward(t) ──async──▶  [GPU runs forward t]
             commit tokens from t-1  (host, overlapped)
             read decision(t-1)  ← the one sync, already finished
  cycle t+1: submit forward(t+1) using forward(t)'s output ...
```

---

## Results

All measured on an M5 Max (614 GB/s), full occupancy, 256 new tokens, code prompts, per-stream
sha-gated byte-identical to single-stream greedy. Interactive charts: throughput scaling +
per-thread tradeoff → **https://claude.ai/code/artifact/e55bc3a5-880a-4af3-9600-badafad14606**
(private Artifact; ask the author to share).

### Aggregate scales with batch — no plateau in range (813 tok/s at 256 streams)

On one consistent kernel config, total throughput climbs monotonically; the knee (where the steep
gains stop) is ~96–128, with a slow tail past it.

| AR streams | 8 | 16 | 24 | 32 | 48 | 64 | 96 | 128 | 192 | 256 |
|---|--:|--:|--:|--:|--:|--:|--:|--:|--:|--:|
| total tok/s | 374 | 479 | 563 | 612 | 649 | 696 | 753 | 758 | 800 | **813** |
| tok/s/thread | 46.7 | 29.9 | 23.5 | 19.1 | 13.5 | 10.9 | 7.8 | 5.9 | 4.2 | 3.2 |

### Two lanes, compared at equal GPU rows (equal compute)

AR wins total; Spec/MTP wins per-request — the throughput vs latency tradeoff. This comparison is
at equal *rows* (equal GPU work); Spec runs 2 rows/request, so a Spec point serves half the streams
of the AR point beside it.

| GPU rows | AR streams | AR total | AR /req | Spec streams | Spec total | Spec /req |
|---:|---:|---:|---:|---:|---:|---:|
| 8 | 8 | 374 | 46.7 | 4 | 236 | **59.0** |
| 16 | 16 | 479 | 29.9 | 8 | 279 | **34.9** |
| 32 | 32 | 612 | 19.1 | 16 | 393 | **24.6** |
| 64 | 64 | 696 | 10.9 | 32 | 450 | **14.1** |

### The per-thread floor picks the operating point

If each request must stay above ~40 tok/s to feel responsive, batching is capped well below the
aggregate ceiling: AR's per-thread speed drops through 40 tok/s at **~11 concurrent threads
(~416 tok/s total)**. The 813 tok/s peak at 256 threads is real but delivers only 3.2 tok/s to each
thread — throughput that a faster per-thread lane (more compute, or a stronger MTP head) would
unlock while staying responsive.

### Long context

Batched decode holds up at 1024-token context (AR 687 tok/s at 64 rows) but **16k context is
memory-bound, not throughput-bound**: AR at just B=8 / 16k already peaks at ~151 GiB — over the
100 GiB working-set budget — so wide 16k batching needs paged/ragged batch-KV (the R1 lane below
lays its foundation) or more RAM. It is not a serving config as-is.

### Reading the numbers
- **Aggregate vs per-request is a tradeoff.** Total climbs with batch; per-request (what one user
  feels) falls. Pick the operating point against a per-thread SLO.
- **AR is the aggregate lane; Spec/MTP is the latency lane.** At equal GPU rows Spec gives faster
  individual threads (higher acceptance would widen that edge); at equal *request counts* AR's row
  efficiency wins both, so AR is the batched-serving default and Spec is for low concurrency.

## Correctness
- Per-stream **fixed-shape sha gate**: every batched stream is byte-identical to that prompt alone
  in an identically-shaped cohort. Green at every tested B on the real model.
- `tests/test_batched_decode.py`: 62 CPU tests over a deterministic row-isolated fake — driver
  bookkeeping, fold-in replay, admission/refill, AR, and the AR-equals-spec committed-sequence
  invariance (greedy is decode-strategy-invariant).
- All new behavior is **default-off**; `decode_mode="spec"` with no refill is #174's path unchanged.

## Scope / caveats
- Throughput numbers are full-occupancy (every slot busy). Real variable-length traffic loses some
  to admission churn; admission hysteresis (batch admissions) recovers most and is a follow-up.
- Wide-row (>16) runs use stock kernels; a tuned wide kernel is a possible follow-up if the curve
  justifies it.
- Quality: the gate proves byte-identity to greedy single-stream, so per-token quality is unchanged;
  a full HumanEval pass is owed before any serving-default cutover.

---

## Additions (2026-07-31): parallel-decoding docs, eval evidence, and 6-bit verify kernels

Three additive commits since the original 12:

- **`docs/perf/PARALLEL_LOCKSTEP_KERNELS.md`** — how the two concurrency
  mechanisms divide the space: width-2 lockstep MTP pairing (measured
  1.41–1.55× aggregate when both streams are decode-ready) vs this lane's
  row-packed batched AR (row-bound MoE decode, and prefill-heavy concurrency
  where pairing measurably collapses). Documents the M6 verify-kernel family's
  exactness contract and bit-width support.
- **`docs/perf/QWEN36_QUANT_EVAL_MATRIX.md`** — full EvalPlus
  (HumanEval+/MBPP+, greedy) across the 4/6/8-bit Qwen3.6 builds with locally
  measured bf16 references and serving rates. Relevant here as end-to-end
  evidence that the concurrency lanes and verify kernels are output-neutral:
  quantized builds land at/above published references through these stacks.
  Headline: the 35B-A3B 6-bit build scores at or above its own bf16 reference
  on all four tiers.
- **6-bit weights on the M6 K-split verify kernel** (`ea89f45`): the kernel
  previously decoded 4-bit only; 6-bit was excluded because 32/6 values
  straddle uint32 boundaries. The addition decodes MLX's continuous
  little-endian bitstream in 3-word/16-value blocks (LCM(6,32)=96 bits, all
  shifts compile-time). Dequant verified bit-identical to `mx.dequantize`
  (one-hot probe, dmax exactly 0.0 at gs 32/64/128); full matmul sits in the
  same reduction-order tolerance band as the existing 4-bit lane (worst rel
  1.3e-3) with per-row argmax parity — the family's documented contract. The
  rendered 4-bit Metal source is asserted byte-identical to the pre-change
  original. Gives the 35B-A3B 6-bit build's dense projections a kernel path
  for solo-MTP verify instead of falling to stock. 35 new tests; bit width is
  part of the kernel cache key and name.


youssofal:
merged as d3c2e04 on top of #174, shipped in 2.4.0. per-stream byte-identity held in our suite. the AR refill lane needed a hidden=None guard once #195's target-only runtimes landed next to it (test gap on the intersection, fixed in integration).

PR #206, opened 2026-07-25
Serving correctness fixes, load-time resident quantization, and a measured-knob registry

davidtai:
## Serving correctness fixes, load-time resident quantization, and a measured-knob registry

Ten self-contained commits, ordered small-to-large. Everything measured on Apple silicon (M-class, 614 GB/s) against `mlx-community/Hy3-oQ2e` (295B MoE, 2-bit experts / q8 residents) served fully resident — numbers inline.

### Fixes

- **Env-flag parsing** (`b2fa09a`-era commit): four variables had divergent readers. The dangerous one: the paged-KV layout pick used a raw membership test that missed spellings (`8`, `8bit`, `uint8`) the rest of the stack normalizes to q8 — silently choosing the dense-decode layout **with a quantized cache**. All readers now route through one strict parser per variable (raises on unknown spellings); `MTPLX_SKIP_VERIFY_SNAPSHOT`'s two-element allowlist inverts to a safe-list so a new verify strategy keeps the snapshot instead of silently inheriting skip; `--chat-template-profile` survives explicit typing; CLI flag recording canonicalizes abbreviated tokens so config files stop overwriting explicitly typed values. 62-test parsing suite included.
- **MTP payload guards** (deepseek / mimo / nemotron_h): a non-empty weight map with no real MTP tensors passed the `if not mapped` completeness check and injected a headless draft surface. Each guard now demands its layer's marker tensors per declared MTP layer.
- **Decode trace**: lane-specific counter sets no longer KeyError the first AR bucket under `MTPLX_DECODE_TRACE_JSONL`.
- **Publish metadata scrub**: strips home-directory paths from `mtplx_runtime.json` before upload while keeping downstream-useful provenance (source repo, SHA, recipe, versions). A published manifest was leaking local paths.

### Features

- **proj-quant / proj-requant** (`load(...)` kwargs or `MTPLX_PROJ_QUANT` / `MTPLX_PROJ_REQUANT`): load-time q4/q8 gs64 affine quantization of the trunk `*_proj` Linears; requant takes pre-quantized residents down (q8 → q4 via dequantize + `QuantizedLinear.from_linear`; double quantization disclosed — BF16-derived stays the higher-quality route). Scope is exactly attention + dense/shared-MLP projections: expert banks excluded by module type, routers/embeddings/head/norms by path; applied before MTP injection so draft heads are never reduced. Measured: per-token reads ~11.9 → ~7.9 GB; AR decode 33.9 → 43.1 tok/s short-context (36.6 @ 1k-token real-code context); paired HumanEval unchanged (McNemar p = 1.0).
- **`serve --draft-core`**: the depth-N device draft core (single sync per cycle) existed in the engine but only bench/tune could select it; the server always ran the stock core. Measured on the K=1 committed-history lane: device 37.9 tok/s vs stock 35.3–36.5 at accept 0.903 — the difference between speculation losing to AR and beating it. Default remains stock.
- **Optimization-profile registry** (advisory v1): per-model measured knob decisions as reviewable in-repo data — each entry carries a state (`default_on` / `default_off` / `not_applicable` / `unvalidated`) and provenance naming the measurement. Seeded with the Hy3-oQ2e set, including the negatives (capture-strategy accept regression, compiled-AR null, kv-quant not_applicable).
- **HumanEval/MBPP harness**: offline pass@k with an explicit `allow_execution` gate (subprocess + timeout containment); validated against canonical solutions (HumanEval 164/164; MBPP 972/974 — two tasks carry upstream-broken float-equality asserts). Plus a real-code programming-prompt corpus for accept/throughput benchmarking at representative prefill shapes.
- **Compiled AR forward** (flag-gated, default off): traces the target call once with per-layer KV state threaded as compile I/O (N layers → 3N leaves). Shipped with its honest measurements: engagement-proofed **null** on hy_v3 (35.1 vs 35.9 — the host graph-rebuild tax it targets isn't present on the lean mlx-lm forward) and a greedy token-parity flip from quantized-kernel selection; both recorded in the registry as `default_off`.
- **Diagnostics**: roofline profiler + env-gated cProfile of AR decode forwards (`MTPLX_AR_PROFILE_TOKENS`).

### Evidence

Full suite: 2,169+ passing (two pre-existing vllm-metal dylib environment failures unrelated to this branch). Serving quality gate on the assembled stack (full HumanEval + MBPP, chat endpoint, temperature 0, seed 42): see the companion hy_v3 PR for the paired table.

### DIFF
--- a/mtplx/benchmarks/code_eval.py
+++ b/mtplx/benchmarks/code_eval.py
@@ -1,0 +1,405 @@
+"""HumanEval / MBPP code-generation evaluation.
+
+Every benchmark in this repo so far measures throughput or acceptance. None
+measures whether the output is still *correct*. That matters here because the
+shipping configuration quantizes aggressively — Q2 routed experts, Q4 trunk
+projections — and a throughput win that silently costs pass@1 is not a win.
+
+This module is the scoring half: load the tasks, build prompts, extract code
+from a completion, execute it against the reference tests, and compute pass@k.
+Getting completions from a model is the driver's job
+(``scripts/code_eval_gate.py``), so this stays importable with no MLX, no
+network, and no model.
+
+## Executing model-generated code
+
+Scoring HumanEval means running code the model wrote. That is inherently
+unsafe, so it is deliberately not the default and not silent:
+
+- every candidate runs in a **fresh subprocess**, never in-process ``exec``
+- a hard wall-clock timeout, then SIGKILL of the whole process group
+- ``resource`` limits on CPU time, address space, and core dumps
+- an empty-ish environment and a scratch cwd that is deleted afterwards
+- the caller must pass ``allow_execution=True`` explicitly; the default
+  refuses, matching the ``--allow-degraded-mtp`` convention elsewhere
+
+This is the standard HumanEval methodology and it is appropriate for
+benchmarking a local model on your own machine. It is *not* a container, and
+it should not be pointed at completions from an untrusted source.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import re
+import shutil
+import subprocess
+import sys
+import tempfile
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any, Sequence
+
+__all__ = [
+    "CodeTask",
+    "CodeEvalError",
+    "TaskResult",
+    "load_humaneval",
+    "load_mbpp",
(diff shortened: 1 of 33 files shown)


davidtai:
Closing — consolidating into the existing hy3 PR (#152).

PR #208, opened 2026-07-25
Hy3: 48 tok/s full-residency serving via 4-bit dynamic requant (quality-gated) + corrected MTP draft contract — continues #152

davidtai:
Serves Tencent Hy3 (295B MoE) from the standard `mlx-community/Hy3-oQ2e` checkpoint, fully resident, with working K=1 speculation and a load-time requantization mode that is the recommended way to run it. Continues #152 — same goal, without the SSD-streaming runtime that PR proposed. The hy_v3 backend's contract gate asked for a hardware-measured runtime contract; this PR supplies it.

| Flagship: 4-bit dynamic requant, full residency | |
|---|---|
| Throughput | **48 tok/s** (K=1, this hardware, this recipe, development lane) · this branch receipts **43.1** — gap attribution below |
| Quality | HumanEval statistically unchanged vs q8 (paired full-164, McNemar p = 1.0) · served gate: **0.8659** HumanEval / **0.7947** MBPP, zero request errors |
| Memory | ~90 GB resident (vs 94–95 GB default) · 16k context fits (default does not) |
| vs default | **+19–23% decode** at every measured shape |

Hardware for all numbers: Apple silicon M-class, 614 GB/s unified memory, mlx-lm PR #1211.

## Running it

```bash
MTPLX_PROJ_REQUANT=q4 mtplx serve --model mlx-community/Hy3-oQ2e \
    --mtp --depth 1 --verify-strategy batched --verify-core stock \
    --draft-core device --profile performance-cold
```

The checkpoint's `mtplx_runtime.json` (included) carries the measured contract: arch id, depth max 1, recommended profile, and the env overrides that keep verify snapshots on.

## Dynamic requant

Decode at this scale is bytes-per-token bound, and the always-read set — attention `q/k/v/o_proj` plus dense/shared-MLP projections — dominates. The checkpoint ships those at q8; `MTPLX_PROJ_REQUANT=q4` (or `load(proj_requant="q4")`) re-quantizes exactly that set at load time, from the published artifact, no republishing:

- Pre-quantized residents are taken down via dequantize → canonical `QuantizedLinear.from_linear` at q4/gs64 — byte-identical to what a load-time quantization would produce. The q8→q4 double quantization is deliberate and disclosed; a BF16 source (via the sibling `proj_quant`) remains the higher-quality route when available.
- Expert banks are excluded by module type, routers/embeddings/lm_head/norms by path, and the pass runs before MTP injection, so the draft head's precision is never touched. Per-token reads drop ~11.9 → ~7.9 GB.

## Results

Decode tok/s, real-code prefill, greedy 256-token decode, bf16 KV. Peak GB in parentheses.

| Context | Lane | Default (q8 residents) | Dynamic requant (q4) |
|---|---|---|---|
| 256 | AR | 30.9 (94.4) | **37.9** |
| 256 | MTP K=1 | 31.2 · accept .796 | **37.2** (90.5) · accept .861 |
| 1,024 | AR | 29.5 (95.4) | **36.2** (91.6) |
| 1,024 | MTP K=1 | 32.9 · accept .875 | **37.0** (91.6) · accept .875 |
| 16,384 | AR | does not fit (101 GB projected) | 24.6 — see note |
| 16,384 | MTP K=1 | does not fit | blocked by the long-context guard |

16k note: the cell ran, but the non-chunked prefill route materializes full hidden/logits intermediates (observed 105 GB peak, decode degraded under the pressure), and the engine's guard blocks the MTP equivalent outside the sustained profile — correctly. Long context belongs to the sustained/chunked lane and is out of scope here; 64k is structurally infeasible fully resident (21.4 GB of KV on top of ~90 GB of weights).

Acceptance trends up with context (.80–.86 at 256 → .875–.903 at 1k) as committed history accumulates.

> **Known issue — MTP acceptance instability.** Acceptance on this family is not stable run-to-run in longer field use: it flips from 90+% to below 50% with no configuration change. The MTP numbers above are samples from the good mode. At sub-50% accept, K=1 drops below AR (the win margin is ~1–5 tok/s), so treat AR as the stable floor and MTP as an opportunistic win until this is root-caused. Suspects: committed-history/session state across requests; near-tie numerics in the draft path.

### Kernel profile

Queued lane (N submits, one eval), real loaded modules, B=1 decode shapes, L2 defeated by cycling layers and random expert slots. ms/call and achieved GB/s vs the 614 GB/s ceiling:

| Kernel | Default q8 | Requant q4 |
|---|---|---|
| attention projs (q,k,v,o) | 0.141 / 570 (93%) | **0.081** / 523 (85%) |
| MoE experts, 8-of-192 gather | 0.105 / 406 (66%) | 0.267 / 159 ¹ |
| shared MLP | 0.037 / 538 (88%) | **0.023** / 454 (74%) |
| router gate (bf16, real path) | 0.011 | 0.011 |
| lm_head (q8, untouched) | 0.856 / 614 (100%) | 0.901 / 584 (95%) |

¹ Expert banks are byte-identical in both variants; this delta is bench variance, not a requant effect. ² An earlier revision showed the router at 63–90 µs — a mixed-dtype bench artifact (fp32 input against bf16 weights), corrected here; see the erratum comment.

Reading: the byte-carrying kernels run at 85–100% of DRAM ceiling, and the requant column is the same kernels moving half the bytes — that is the whole +23%. Summed kernel time accounts for ~60% of the measured AR step; the remaining gap to the 48 flagship (AR step 27.9 ms vs ~13 ms of bytes at 1k) is distributed attention/eval overhead with no single hot kernel, and its attribution is the branch's active work along with the K=1 verify's byte-floor gap (cycle 1.88 → 1.66 AR-steps).

## MTP: what was broken, what was fixed

Acceptance was 0.12–0.26 when this work started. Three measured causes:

1. **The draft head consumes the post-final-norm trunk hidden.** Teacher-forced 2×2 probe over a 1024-token real-code prompt: 0.773 agreement post-norm vs 0.387 pre-norm vs 0.000 with the eh_proj concat flipped. The pre-norm contract previously documented in the backend is refuted by measurement and corrected.
2. **Draft history must be primed over the prompt** (`committed` policy) — unprimed, the head is context-starved at 0.26.
3. **The device draft core** (single sync per cycle, already in-tree) was unreachable from serving; `serve --draft-core` exposes it. It is the difference between speculation losing to AR and beating it.

Supporting engine work, each with regression tests:

- **Checkpoint graft** — released mlx-lm sanitizes the appended NextN layer away; injection now builds the draft head from the checkpoint's canonical `model.layers.{N}.*` tensors (index-driven, per-module quant overrides honored). Native `model.mtp` binding remains the fast path for when mlx-lm's MTP revision lands; the registry gate accepts the tencent-native layout alongside repacked `mtp.*`.
- **Snapshot-free rejection repair** — product profiles set `MTPLX_SKIP_VERIFY_SNAPSHOT=1`; under batched verify the first rejected draft then had no repair path and the request 500'd. Attention-only caches now repair by trimming the uncommitted verify tail, token-for-token identical to the snapshot arm.
- **Compiled-AR hook reachability** — MTP wrappers advertise `emit_logits` via `**kwargs`, which made the decode fast branch unreachable on exactly the runtimes it targets.

## Quality

Full suites over HTTP against the served model (requant arm), chat endpoint, temperature 0, seed 42, zero request errors:

| Arm | HumanEval pass@1 | MBPP pass@1 |
|---|---|---|
| MTP K=1 device | **0.8659** (142/164) | **0.7947** (774/974) |
| AR | 0.8598 (141/164) | — |

One task apart — the paired McNemar-p≈1.0 shape; exact verification is quality-neutral as promised.

## Also in the diff

- Canonicalized env-flag parsing (one strict reader per variable; the paged-KV membership test silently picked the dense-decode layout for a quantized cache) — 62-test suite.
- MTP payload guards for deepseek/mimo/nemotron_h (non-empty weight maps with no real draft tensors no longer inject headless surfaces).
- Decode-trace tolerance for lane-specific counter sets; publish-metadata scrub (a published manifest leaked home paths).
- Per-model measured-knob registry (advisory): every decision above as reviewable data with provenance, including the negatives — capture-class verify (slower here, accept .903→.776), compiled-AR (flag-gated; engagement-proofed null with a greedy parity flip), kv-quant (not applicable).
- HumanEval/MBPP harness (offline pass@k, explicit `allow_execution` gate; validated 164/164 and 972/974 on canonical solutions), real-code prompt corpus, roofline/AR-profile diagnostics.

Suite: 2,171 passing (two pre-existing vllm-metal dylib environment failures predate this branch; hy_v3 test files skip on released mlx-lm until the model lands).

### DIFF
--- a/mtplx/backends/hy_v3_mtp.py
+++ b/mtplx/backends/hy_v3_mtp.py
@@ -2,7 +2,7 @@
 
 Hy3 ships one appended MTP layer (num_nextn_predict_layers=1): a full MoE
 decoder layer fed concat[RMSNorm(next-token embedding), RMSNorm(trunk
-pre-final-norm hidden state)] through an eh_proj down-projection, sharing the
+post-final-norm hidden state)] through an eh_proj down-projection, sharing the
 trunk's embeddings and lm_head. The MLX reference implementation exposes it as
 ``Model.predict_next_tokens(hidden, token_ids, cache)`` with
 ``return_hidden_states=True`` on the trunk forward (see
@@ -57,8 +57,8 @@ class HyV3MTPBackend(MTPBackend):
                 "Single appended NextN layer with its own 192-expert MoE MLP, "
                 "sigmoid top-8 routing with expert bias, eh_proj over "
                 "concat[enorm(embedding), hnorm(hidden)], shared embeddings "
-                "and head. Draft layer consumes the trunk pre-final-norm "
-                "hidden state. Verification is exact rejection sampling in "
+                "and head. Draft layer consumes the trunk POST-final-norm "
+                "hidden state (measured 0.773 vs 0.387 pre-norm). Verification is exact rejection sampling in "
                 "generation.py; the MLX reference (mlx-lm hy_v3 MTP revision) "
                 "verifies greedily and is temp-0 exact."
             ),
--- a/mtplx/backends/registry.py
+++ b/mtplx/backends/registry.py
@@ -816,23 +816,35 @@ _HY_V3_MTP_MARKER_SUFFIXES = (
 
 
 def _passes_hy_v3_gate(inspection: Any) -> bool:
-    """Hy3's appended MTP block lives directly under an ``mtp.`` prefix
-    (``mtp.enorm.weight``, ``mtp.hnorm.weight``, ``mtp.eh_proj.weight``,
-    ``mtp.final_layernorm.weight``, ``mtp.layer.*``) rather than the
-    ``mtp.layers.{idx}.`` nesting DeepSeek/GLM/Step use, so it needs its own
-    gate instead of `_passes_appended_layer_gate` (verified against the
-    shipped `hy3-demolition-mlx-*-mtp` checkpoints' safetensors index)."""
+    """Hy3's appended MTP block ships in one of two layouts: repacked
+    checkpoints put it directly under an ``mtp.`` prefix (``mtp.enorm.weight``,
+    ``mtp.eh_proj.weight``, ``mtp.layer.*`` — verified against the shipped
+    `hy3-demolition-mlx-*-mtp` indexes), while tencent-native exports keep the
+    canonical appended-layer form ``model.layers.{num_hidden_layers}.*``
+    (``...enorm.weight``, ``...self_attn.*``, ``...mlp.*``). Neither uses the
+    ``mtp.layers.{idx}.`` nesting DeepSeek/GLM/Step share, so this stays a
+    dedicated gate instead of `_passes_appended_layer_gate`."""
     keys = _weight_keys(inspection)
     if not keys:
         return False
     count = int(getattr(inspection, "mtp_num_hidden_layers", 0) or 0)
     if count <= 0:
         return False
-    return _has_marker_under_prefixes(
+    if _has_marker_under_prefixes(
         keys,
         ("mtp.",),
         _HY_V3_MTP_MARKER_SUFFIXES,
         ("mtp.layer.",),
+    ):
+        return True
+    start = int(getattr(inspection, "num_hidden_layers", 0) or 0)
+    if start <= 0:
+        return False
+    return _has_marker_under_prefixes(
+        keys,
+        (f"model.layers.{start}.",),
+        _HY_V3_MTP_MARKER_SUFFIXES,
+        ("self_attn.", "mlp."),
     )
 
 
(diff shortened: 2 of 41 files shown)


davidtai:
Erratum on the kernel table: an earlier revision showed the router gate as a 5 ms/token launch-bound outlier. That was a measurement artifact — the bench fed fp32 input to bf16 weights, hitting the mixed-dtype matmul slow path (78 µs vs 10.9 µs on the real bf16 path). The table and gap attribution are corrected; the router is negligible, and the remaining 43.1→48 distance is attributed to AR-step occupancy, not any single kernel.

youssofal:
merged as ec844ea, shipped in 2.4.0. the post-final-norm draft contract finding checks out and it matters more than the hy3 lane itself: it is the same signature we later found on the 35B qwen lane (two independent heads pinned to the same live ceiling). recurrent-cache loud failure kept as you had it. the F401 got cleaned in integration.

PR #210, opened 2026-07-27
Fix slow Hugging Face config probes in Forge

PhilipJohnBasile:
## What changed

- Raise the app probe request timeout from 12 to 30 seconds for large generated MLX configs.
- Recover transport, transient HTTP, and decode failures through Hugging Face model metadata.
- Retry the complete config at the immutable repository SHA so compact indexed metadata cannot create false `no MTP` verdicts or lose quantization details.
- Preserve the existing auth, missing-repo, GGUF, mirror, and pair-bundle behavior.
- Add a regression using `philipjohnbasile/hy3-demolition-mlx-reap25-v1-mtp`, including its real source revision and sparse indexed-config shape.

## Why

The Forge picker fetched only `<repo>/resolve/main/config.json` with a 12-second timeout. The Hy3 MLX artifact has a generated config of roughly 539 KB with a large per-tensor quantization map. A slow raw-file response therefore surfaced as `Couldn't fetch config.json` / `Unknown format`, even though the CLI correctly recognizes the public artifact as forgeable `hy-v3-mtp`.

## Impact

Slow but valid public MLX/MTP repositories can now pass preflight instead of being falsely rejected, while negative MTP classification still requires the complete authoritative config.

## Validation

- `HuggingFaceProbeForgeTests`: 28 tests passed, 0 failures.
- Dedicated timeout fallback harness: 3 tests passed, 0 failures.
- `git diff --check`: clean.
- Cross-fork compare is based directly on current upstream `main`: 2 commits ahead, 0 behind, only the probe and its tests changed.

The full Swift package gate could not run in the Codex sandbox because external package resolution is network-blocked and Xcode-beta's SwiftData macro plugin cannot launch under the nested sandbox. The focused production probe suite compiled and passed against the exact modified source.

### DIFF
--- a/apps/MTPLXApp/Sources/MTPLXAppCore/Onboarding/HuggingFaceProbe.swift
+++ b/apps/MTPLXApp/Sources/MTPLXAppCore/Onboarding/HuggingFaceProbe.swift
@@ -204,7 +204,15 @@ public struct HuggingFaceProbe: Sendable {
     /// object. `nil` on any failure — callers treat these fetches as
     /// best-effort signals, never hard errors.
     private func fetchRepoJSON(repo: String, path: String) async -> [String: Any]? {
-        guard let url = URL(string: "\(endpointBase)/\(repo)/resolve/main/\(path)") else {
+        await fetchRepoJSON(repo: repo, revision: "main", path: path)
+    }
+
+    private func fetchRepoJSON(
+        repo: String,
+        revision: String,
+        path: String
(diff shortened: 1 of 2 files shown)


youssofal:
merged after our own review, shipped in 2.4.0. the pinned-SHA retry is the right shape, and refusing to treat indexed-config absence as "no MTP" avoids exactly the false-negative class we care about in forge. both regression tests run green in our suite.

PR #212, opened 2026-07-28
feat(server+app): serve embeddings and reranking from the MTPLX daemon

Cyb3rb1ade:
## What

Adds `POST /v1/embeddings` (OpenAI shape) and `POST /v1/rerank` (Cohere/Jina shape) to the MTPLX daemon, so a retrieval-backed setup — RAG, agent memory — no longer needs a second inference server running beside MTPLX purely to answer embedding calls. Configurable from the CLI and the macOS app, and observable in the dashboard.

```bash
mtplx serve \
  --embedding-model mlx-community/Qwen3-Embedding-8B-4bit-DWQ \
  --reranker-model vserifsaglam/Qwen3-Reranker-4B-4bit-MLX
```

Both flags repeat, accept a Hugging Face id or a local path with an optional `REF=served-id` alias, and are selected per request via `"model"`. Also on `quickstart`, in the app under **Settings → Retrieval endpoints**, and in `~/.mtplx/config.toml` as `embedding_models` / `reranker_models` / `retrieval_max_resident`.

## Why not through MTP

Retrieval models deliberately bypass the MTP path. Multi-token prediction makes *next-token* decoding cheaper, which means nothing for a model that returns a vector instead of a token stream — there is no draft to accept or reject. They run the transformer stack directly:

- **Embedding** — final hidden state of the last real token, cast to float32, L2-normalised. Qwen3-Embedding appends `<|endoftext|>` and pools that position, which `pooling="last"` reproduces.
- **Reranking** — the official yes/no judging prompt, then a softmax over the `yes`/`no` logits at the last real position.

Batches are padded on the **right**. With causal attention a real token never attends to a later pad token, so right padding leaves every real position bit-identical to an unpadded run; left padding would corrupt it.

## Memory behaviour

- **One model, many roles.** Backends are keyed by resolved filesystem path, so the same weights registered under several served ids, under both roles, or under both a Hugging Face id and the local path it resolves to are loaded once.
- **Lazy.** Models load on first request; an unused endpoint costs nothing.
- **Capped, without pulling the rug.** `--retrieval-max-resident` (default 2) unloads the least recently used backend beyond the cap, but a backend stays pinned for its whole load-and-inference lifetime and is never evicted while in use. When every resident backend is busy the cap is briefly exceeded — the alternative is unloading weights an in-flight request is about to use, which frees nothing and forces an immediate reload.

## Additive by construction

With no flag configured the endpoints answer 404 and the chat runtime is untouched. `/v1/models` gains a `capability` field (`chat` / `embedding` / `rerank`) alongside the existing entry.

**Three wiring points, worth knowing if this area is refactored:** the server argv is rebuilt rather than inherited, so the flags are threaded through the `serve`/`quickstart` parsers, the child argv builder in `commands/public.py`, **and** the server module's own parser. Missing any one leaves the endpoints silently unconfigured on every path but a bare `mtplx serve`. The same applies to the cache directory: the chat model reaches the server pre-resolved, so retrieval references need `--retrieval-cache-dir` to find models pulled into a custom `--cache-dir`.

## Observability

Configuration that cannot be observed is a guess, so the dashboard snapshot carries a `retrieval` section: per model the role, load and residency state, request/item/token/error counts, cumulative compute, load duration, derived throughput and average latency, and the last error. The key is always present, letting a client tell "no retrieval configured" apart from "this daemon predates retrieval".

Load time is measured but **excluded** from request latency. Counting it made a cold first request read as ~10 s average latency on a model that serves in ~450 ms — precisely the number a user would act on.

The app renders this in the Activity tab and distinguishes the states that matter: configured but never loaded (normal, because loading is lazy), loaded but unused, working, and failing.

## Evidence

Hardware: Apple Silicon, 48 GB unified memory. Models: `mlx-community/Qwen3-Embedding-8B-4bit-DWQ` (4-bit DWQ) and `vserifsaglam/Qwen3-Reranker-4B-4bit-MLX` (4-bit). Profile `sustained`, fan mode default (no fan-controlled runs). Base commit 97fb388.

- **Tests**: 48 Python tests (`tests/test_retrieval.py`) and 15 Swift tests, covering flag parsing, served-id resolution, resolved-path backend sharing, pinning against eviction, slot accounting, metric derivation, error recording, token usage, settings round trip, backward-compatible decoding, and the HTTP contract including the chat-only 404 path. Full suites green: 365 Python in the touched areas, 533 Swift.
- **Live, through the real routes**: 4096-dim vectors with norm `1.000000`; reranker separating a matching from a non-matching document `0.9994` vs `0.0000`; base64 decoding to the same vectors; `13.4 texts/s` at `447 ms` per request with the `6.0 s` weight load reported separately; real token usage (`14` for two short embedding inputs, `160` for a two-document rerank); `/v1/chat/completions` answering from the same daemon throughout.
- **Vector parity**: against the same two models served by another local runtime, worst-case cosine similarity **0.9998** with identical reranker ordering — existing indexes do not need re-embedding.

No performance claim is made about the chat path, which is unchanged.

## Commits

1. `feat(server+app)` — the endpoints, CLI/app/config wiring, tests, docs.
2. `fix(retrieval)` — round one of review: resident-slot reservation, forwarding the cache directory, honouring `encoding_format` (base64, unknown values rejected rather than silently misread).
3. `feat(retrieval)` — live load state and throughput in the snapshot and the app.
4. `fix(retrieval)` — round two of review: pinning backends through inference, keying by resolved path, recording failures, and reporting real token counts.

## Contribution gate

`tests/test_no_mlx_imports.py` and `tests/test_runtime_kpis.py` pass; `python -m build` produces both artifacts with `mtplx/retrieval.py` in the wheel.

`tests/test_public_cli.py` has one failure, `test_start_hermes_live_path_writes_profile_and_handoff`, which asserts `port == 18085`. It fails identically on a pristine checkout of the base commit in this environment because port 18085 was already bound — verified in a separate `git worktree` at `HEAD`, so it is not a regression from this change.


Cyb3rb1ade:
Thanks — all three findings were reproducible and are fixed in `3193172`, each with a regression test.

**P1 — resident slots (`mtplx/retrieval.py`).** Confirmed. Loading happens outside the registry lock, so two concurrent first-use requests each saw the other as not yet loaded, skipped eviction, and both stayed resident. Residency is now reserved on *acquisition* rather than on completed load, so an in-flight load occupies its slot. Test: `test_a_slot_is_reserved_before_the_weights_finish_loading` acquires a second model while the first is still unloaded and asserts the first is evicted — it fails on the previous implementation.

**P2 — cache directory.** Confirmed, and the cause is slightly different from the description, which matters for the fix: the server subprocess has no cache-directory argument *at all*. The chat model is resolved to an absolute path before launch, so nothing downstream needs one; retrieval references stay symbolic and were resolved against the default cache. Added `--retrieval-cache-dir`, forwarded from the public CLI's `cache_dir` whenever retrieval models are configured, and `registry_from_args` now prefers it over the in-process `cache_dir` / `model_dir` fallbacks. Tests cover both the forwarded value and the fallback.

**P2 — `encoding_format`.** Confirmed. The field was accepted and ignored. `base64` now returns the little-endian float32 buffer OpenAI clients decode with `numpy.frombuffer(..., dtype="float32")`, `float` remains the default, and any other value is rejected with 400 instead of being silently misread. Tests cover all three paths.

Verified live against `Qwen3-Embedding-8B-4bit-DWQ` through a running daemon: base64 decodes to 4096 dimensions with norm `1.000000`, `float` still returns a JSON array, and `float16` is refused with 400.

Python retrieval tests are now 34, all passing, alongside the 8 Swift tests.

Cyb3rb1ade:
@codex review

Two commits landed since the first pass: the three findings from that review are fixed in `3193172`, and `3c6c2db` adds retrieval status and throughput to the dashboard snapshot and the app. Fresh eyes on those would be welcome — particularly the slot-reservation change in `RetrievalRegistry._backend` and the metric accounting in `RetrievalStats`.

Cyb3rb1ade:
All four findings from the second pass were reproducible and are fixed in `fd2a238`, each with a regression test.

**P1 — pin acquired backends.** Correct, and my previous fix was insufficient: reserving a slot still left the window where a request had exited the acquisition path but not finished, so eviction unloaded weights it was about to use and they were immediately reloaded — two models live, one counted. A backend is now pinned for its whole load-and-inference lifetime, and eviction skips pinned backends. When every resident backend is busy the cap is briefly exceeded, which is the lesser evil versus pulling weights out from under an in-flight request.

**P2 — key by resolved path.** Correct, and it contradicted this module's own docstring. Lookup used the raw reference, so `org/model` and `/models/model` produced two backends for one directory. Resolution is now cached per reference and used as the residency key. This surfaced a second defect while fixing it: `descriptors()` still looked backends up by raw reference, so load state was reported against the wrong entry whenever an alias shared a backend — also fixed.

**P2 — record failures.** Correct. `last_error` was only ever cleared, never set, so the dashboard row I added for it could never appear and a model that failed to resolve read as idle. Failures are recorded before re-raising and still cleared by the next success; an `errors` counter distinguishes "never used" from "only ever failed".

**P2 — real token usage.** Correct. Counts now come from the encoding already performed per batch. Verified live: 14 prompt tokens for two short embedding inputs, 160 total for a two-document rerank.

Suites after the change: 48 Python retrieval tests, 533 Swift, all green.

Cyb3rb1ade:
@codex review

`fd2a238` addresses the second pass. The concurrency change is the subtle one and deserves a third look: backends are now pinned for their whole load-and-inference lifetime in `RetrievalRegistry._acquire`, eviction skips pinned entries, and residency is keyed by resolved filesystem path rather than the raw reference. Description updated to match.

Cyb3rb1ade:
Both findings were reproducible and are fixed in `8c7890c`.

**P1 — enforce the cap on release.** Correct. Acquisition skipped pinned victims as intended, but the release path only decremented the pin count, so a legitimate overshoot from overlapping requests stayed loaded until the next request happened to trigger eviction. Eviction now runs again as backends become unpinned.

Writing the regression test surfaced a second defect in the same code: eviction could drop the *most recently used* backend whenever an older one was pinned, which inverts LRU — with a cap of 1, finishing the inner of two overlapping requests evicted the model that had just been used and kept the older one. The pass now never considers the most recently used entry, so a surplus resolves by dropping the oldest unpinned model. Same rule at both call sites, which also removed the `protect` parameter.

**P2 — role-shared identities.** Correct, and it hit the case this feature advertises: one reference in both roles yields two entries under the same served id, so `Identifiable` collided and a row could be dropped or reused. Identity is now `role:servedId`; `servedId` stays available for display and for the `"model"` field clients send. Covered by a test that decodes one reference in both roles and asserts two distinct identities with per-role counters.

Suites: 49 Python retrieval tests, 16 Swift retrieval tests, 366 Python and 533 Swift overall — green. Verified live on the running daemon afterwards.

youssofal:
this is a serious contribution and the self-review discipline shows. it's on the bench for a heavy pass this weekend: embedding/rerank correctness against reference implementations, residency behavior under the memory pillar, concurrent load next to a decode session. not merging before that, but expect movement.

youssofal:
direction accepted — a retrieval surface in the daemon is something we want, and this is a strong writeup (right-padding for causal exactness and excluding load time from request latency are both the correct calls, and the three-parser argv-threading warning is appreciated).

timing honesty so you're not left guessing: the current cycle is fully consumed by agent-lane work, so the proper review pass here (argv threading, eviction pinning under concurrency, the app surface) lands next cycle rather than this week. no changes requested yet — we'll review from whatever head is current when we get there.

Cyb3rb1ade:
One thing worth settling before the heavy pass — found while running this branch's build in daily use, not in review.

## `/v1/models` mixes capabilities, and discovery clients cannot tell them apart

With `--embedding-model` and `--reranker-model` set, `GET /v1/models` returns:

```json
{"id": "hawhyhb-qwen36-35b-a3b-uncensored-...", "capability": "chat"}
{"id": "jina-embeddings-v5-text-small-mlx",     "capability": "embedding"}
{"id": "jina-reranker-v3.5-mlx",                "capability": "rerank"}
```

`capability` is our extension. It is not part of the OpenAI models schema, so a
client that discovers models has no way to filter on it — and several do
discover. Hermes auto-populated its model picker from this endpoint and wrote
all three into its profile config as selectable chat models:

```yaml
providers:
  custom:
    discover_models: true
    models:
      mtplx-qwen36-35b-a3b-optimized-speed: {}
      hawhyhb-qwen36-35b-a3b-uncensored-heretic-mtplx-4bit-fp16: {}
      jina-embeddings-v5-text-small-mlx: {}      # offered as a chat model
      jina-reranker-v3.5-mlx: {}                 # offered as a chat model
```

Nothing in the client is at fault here. It asked what models are available and
believed the answer.

## It fails quietly rather than loudly

Selecting one of those does not produce an error. The daemon accepts any model
id and answers with whatever chat model is loaded:

```
loaded per /health : hawhyhb-qwen36-35b-a3b-uncensored-heretic-mtplx-4bit-fp16
requested          : mtplx-qwen36-27b-optimized-speed
answered by        : hawhyhb-qwen36-35b-a3b-uncensored-heretic-mtplx-4bit-fp16
```

So picking `jina-embeddings-v5-text-small-mlx` as a chat model returns ordinary
chat completions from the loaded model. The user gets a plausible answer from a
model they did not choose, with no signal that the selection was ignored.

The substitution behaviour predates this PR — a single-model daemon ignoring the
requested id is defensible on its own. What this PR changes is that the model
list now contains entries which *cannot* serve chat, so a wrong pick is no
longer merely cosmetic.

## Options

1. **Filter on request** — `GET /v1/models?capability=chat`, with chat-only as
   the default response. Discovery clients then see only what they can use, and
   anything wanting the full inventory asks for it.
2. **Separate the surfaces** — keep `/v1/models` chat-only and expose retrieval
   models under their own path.
3. **Reject mismatched ids** — return an error when a request names a model with
   a capability that cannot serve the endpoint. Fixes the silent-substitution
   half regardless of which listing option is chosen.

Happy to prepare whichever fits the direction you have in mind.

youssofal:
Merged for 2.6.0. The commits landed through the release lineage rather than the merge button, so GitHub will not auto-mark this; closing manually. Everything shipped with your design intact: /v1/embeddings (OpenAI shape, `dimensions` truncation), /v1/rerank (Cohere/Jina shape), opt-in per-model flags, lazy load, LRU resident cap, idle release, capability-separated /v1/models, and the trust-remote-code gate. 17 commits, credited in the release notes.

Issue #196, opened 2026-07-22
Tool calls for Qwen3.6 are emitted as malformed / unterminated JSON (blocks writing source files)


jeffwood-lab:
### mtplx doctor --json

Environment

MTPLX version: 2.3.0
Model: Qwen3.6-27B-MTPLX-Optimized-Speed-FP16 (also seen on Qwen3.6-35B-A3B-...)
Launch command / flags: <fill in — e.g. --reasoning-parser, --tool-prompt-mode, --chat-template-profile>
Bridge: OpenAI-compatible /v1 (also reproduced on Anthropic /v1/messages)
Client: opencode via @ai-sdk/openai-compatible

Summary
When the model calls a tool whose arguments contain a large or heavily-escaped string (e.g. a write with file content full of " { } < > \n), MTPLX emits the tool call as invalid or unterminated JSON, or leaks it into the assistant text channel instead of a structured tool_call. Small prose writes succeed; code/markup writes fail — pointing at JSON string-escaping of the arguments.

Observed behavior

Hard error on a write of an HTML/CSS file:

error: { "name": "UnknownError", "data": { "message": "malformed tool_call: unterminated stream" } }

Earlier instance — a bash tool call with a broken escape (timeout\":), never validly closed, spilling into free text:

{"name": "bash", "arguments": {"command": "cat design/spec.md", "tool": "bash", "timeout\": 10000}Let me get the full spec content via bash. ...

That response then ran to finish: "length" with output: 32000 tokens (a runaway, downstream of the broken tool call).

Expected behavior
Tool calls should always be emitted as complete, well-formed, properly-escaped JSON in the structured tool_calls field — or fail cleanly — regardless of payload size or special-character density.

Impact
Blocks the core operation of any coding agent: the model cannot reliably write source files (which are escaping-heavy by nature). Prose writes work; code writes do not.

Repro
Ask the model to write a ~150+ line file containing HTML/CSS/JS (quotes, braces, angle brackets, newlines) in a single tool call.

Questions / things tried

Does MTPLX_TOOL_CALL_STRICT=1 enforce valid/terminated tool-call JSON?
Does --tool-prompt-mode native vs hybrid change the escaping behavior?
Is there a max tool-call argument length or streaming buffer limit being hit?

### Exact command

<fill in — e.g. --reasoning-parser, --tool-prompt-mode, --chat-template-profile>

### Model path or repo id

Qwen3.6-27B-MTPLX-Optimized-Speed-FP16

[...]

PR #213, opened 2026-07-28
Bound the think prelude so an unclosed </think> can't run away

amoreo140:
Bound the think prelude so an unclosed `</think>` can't run away

## The problem

The think prelude from #186 exists because Qwen-style templates open `<think>` inside the
generation prompt, so generation starts mid-reasoning and the grammar has to let the model
back out. Its terminal is unbounded free text:

```lark
start:    prelude? doc
prelude:  TAG_TEXT </think>
TAG_TEXT: /(.|\n)*/
doc:      %json {…}
```

That makes one failure mode **legal**. A model that never emits `</think>` is not violating
the grammar — it is staying inside the prelude, and it will happily fill `max_tokens` with
prose and return no document at all. The grammar cannot stop it, because the grammar permits it.

## Evidence

Hardware/config: Apple M5 Max 128 GB, macOS 15, mtplx 2.3.0 (commit 97fb388), Qwen3.6 VL MoE
(35B-A3B lineage) 6-bit MLX with a grafted MTP sidecar, `--generation-mode mtp --depth 1`,
`--no-stats-footer`, temperature 0.6 (server default) and 0.2, non-streaming
`/v1/chat/completions` with `response_format: json_schema`, 2026-07-28/29.

With thinking on, **~25 % of requests never produced a document**: `finish_reason: length`,
`reasoning_content` empty, and 13 000–41 000 characters of prose in `content` beginning
`"Here's a thinking process that leads to…"`, with `mtplx_stats.constraint_completed: false`.

Raising the budget does not fix it — it only makes each failure more expensive:

| arm | valid JSON | median |
|---|---:|---:|
| thinking off | 6/6 | 0.57 s |
| thinking on, `max_tokens` 4000 | 4/6 | 8.9 s |
| thinking on, `max_tokens` 16000 | 5/6 | 9.0 s (the failure cost **167 s**) |

The failure is nondeterministic — repeated runs of the same suite gave 2/8 and 3/8 — which
matches the nondeterminism reported at temp 0 in #196.

## The change

Give the prelude its own terminal and bound it. Default 4000 characters;
`MTPLX_THINK_PRELUDE_MAX_CHARS=0` restores the previous behaviour verbatim. The bound is part
of the grammar cache key, so changing it cannot return a stale compiled grammar.

`tail`/free text keep the unbounded `TAG_TEXT`, so the assistant's **visible answer is never
capped** — only the reasoning segment that precedes constrained output is.

The tool-call grammar shares the same prelude and is bounded the same way. That is the symptom
reported in #196: *"the content channel fills with the model's reasoning narration … until
`finish: length`, with no tool call ever emitted."*

## Results after the change

Same hardware and settings:

| | before | after |
|---|---|---|
| schema + thinking, valid documents | 4/6 and 5/6 | **62/62** |
| arithmetic suite (n=8), thinking on | 5–6/8 correct + 2–3 runaways | **8/8, zero runaways** |
| schema, thinking **off** (regression check) | 6/8 | 8/8, cap never engaged |

The cap engages only when it is needed — 3/10 on a short prompt, 0/20 on a longer one — and
every capped request then emitted a valid document with `finish_reason: stop`. Quality did not
regress: with the bound in place, single-call constrained reasoning reaches the same 8/8 as a
two-pass reason-then-format pipeline, without the second call.

Temperature was checked separately (n=12 each): 12/12 valid at both 0.2 and 0.6, with the cap
engaging in neither.

## Why in the grammar rather than in a thinking budget

A scheduler-side budget is the usual answer, and MTPLX already has one
(`--agent-thinking-budget`) — but it only arms for requests that declare `tools` and have
thinking enabled, so plain `response_format` requests never get it.

More importantly, a grammar bound rides the sampling-time token mask instead of scheduler
state, so it cannot go stale under speculative decoding. That is not hypothetical: vLLM's
`thinking_token_budget` silently did nothing whenever MTP was on — budget enforcement ran
against pre-incremented token state during multi-token acceptance batches, so the stop
condition fired late or never — and it went unnoticed for over a month, fixed only in
vLLM 0.21.0. MTPLX's guard is the same shape of mechanism and MTPLX ships MTP on by default.

## Tests

Nine tests added to `tests/test_constrained.py`: default bound, configurability, `0` restoring
the old behaviour, the bound participating in the cache key, the tool-call prelude being
bounded while `tail` stays free, no prelude terminal when thinking is off, and llguidance
compiling all variants.

```
pytest tests/test_constrained.py -q     28 passed
pytest tests/test_no_mlx_imports.py -q  15 passed
```

## Trade-off, stated plainly

A hard character cap cuts mid-sentence, where vLLM's and llama.cpp's designs (llama.cpp
PR #25961) look for a paragraph break first. A grace window before the forced close would be
a natural follow-up. It is a ceiling, not a floor — the model is still free to finish earlier,
and in the majority of our requests it did.

### DIFF
--- a/mtplx/constrained.py
+++ b/mtplx/constrained.py
@@ -290,6 +290,43 @@ def _lark_string(text: str) -> str:
     return json.dumps(text)
 
 
+_THINK_PRELUDE_DEFAULT_MAX_CHARS = 4000
+
+def _think_prelude_max_chars() -> int:
+    """Character cap on the reasoning segment that precedes constrained output.
+
+    The think prelude exists because Qwen-style templates open ``<think>`` inside
+    the generation prompt, so generation starts mid-reasoning and the grammar must
+    allow the model back out (see #186). That prelude was unbounded free text, which
+    makes one failure mode *legal*: a model that never emits ``</think>`` stays inside
+    the prelude and fills ``max_tokens`` with prose, returning no document at all.
+    Reported symptom on the tool-call side in #196 ("the content channel fills with
+    the model's reasoning narration ... until finish: length, no tool call emitted").
+
+    Bounding the prelude regex makes the grammar itself force the close. Because the
+    bound is carried by the sampling-time token mask rather than by scheduler state,
+    it cannot go stale under speculative decoding -- the failure mode that silently
+    disabled vLLM's thinking budget whenever MTP was on, fixed only in vLLM 0.21.0.
+
+    ``MTPLX_THINK_PRELUDE_MAX_CHARS=0`` restores the previous unbounded behaviour.
+    """
+    raw = os.environ.get("MTPLX_THINK_PRELUDE_MAX_CHARS")
+    if raw is None or raw.strip() == "":
+        return _THINK_PRELUDE_DEFAULT_MAX_CHARS
+    try:
+        value = int(raw)
+    except ValueError:
+        return _THINK_PRELUDE_DEFAULT_MAX_CHARS
+    return value if value > 0 else 0
+
+
+def _prelude_terminal(max_chars: int) -> str:
+    """The prelude's own terminal, so bounding it never touches tail/free text."""
+    if max_chars <= 0:
+        return "PRELUDE_TEXT: /(.|\\n)*/\n"
+    return f"PRELUDE_TEXT: /(.|\\n){{0,{max_chars}}}/\n"
+
+
 def _tool_call_lark_grammar(
     functions: list[tuple[str, dict[str, Any]]],
     *,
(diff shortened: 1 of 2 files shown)


youssofal:
merged, shipped in 2.4.0. clean find, the prelude being legally unbounded was our miss in #186, and your cache-key catch saved us a stale-grammar bug on top. we kept the 4000-char default and the MTPLX_THINK_PRELUDE_MAX_CHARS=0 escape hatch, and re-ran your grammar cells plus our strict tool-call battery on the merged tree, plus a live check on all four 27B models. one repo convention note: co-author trailers get stripped at integration, your authorship is preserved on the commit.

PR #214, opened 2026-07-31
Add experimental MoE DFlash integration and evidence gates

PhilipJohnBasile:
## Status

Draft and experimental. This is not a release candidate: the full correctness
gate remains red, and DFlash stays disabled unless
`MTPLX_DFLASH_DRAFT=1` is set.

## What changed

- adds a pinned DFlash draft source with five target hidden-state taps and
  commit-only companion-cache handling
- adds two experimental MoE lanes:
  - wide B8 one-hot verification
  - staged-K1 B8 over the compiled A3B target-prefix and whole-MoE route
- adds compiled target-prefix tap outputs and fail-closed route checks
- adds protocol manifests that pin revisions, prompts, sampling, thinking
  mode, execution switches, and aggregation
- records the runtime-observed MTP contract, generation mode, speculative
  depth, forward-call counts, compiled route, and fallback count
- recovers legacy `forge_provenance.mtp_contract` metadata while preserving
  canonical top-level `mtp_contract` precedence
- includes the Phase-0 measurements, adversarial corrections, and raw receipts
  that led to the current decision

## Corrected artifact-contract evidence

Target:
`Youssofal/Qwen3.6-35B-A3B-MTPLX-Optimized-Speed`

Drafter:
`z-lab/Qwen3.6-35B-A3B-DFlash@f181eece646affea2c38b2765f1aaa01a9734ccd`

24 coding prompts, greedy, thinking disabled, 112-token cap:

| Lane | AR | MTP D1 | DFlash B8 | Equality |
|---|---:|---:|---:|---|
| Wide capture-commit | 109.57 tok/s | 112.86 (1.03x) | 130.65 (1.19x) | DFlash=AR 9/24; MTP=AR 11/24; three-way 8/24 |
| Staged-K1 compiled whole-MoE | 129.92 tok/s | 176.08 (1.36x) | 171.26 (1.32x) | DFlash=MTP 24/24; both=AR 13/24 |

The staged DFlash lane introduces no additional divergence relative to its MTP
K1 control, but that control still diverges from authoritative AR on 11/24
prompts. The PR intentionally preserves that failed gate rather than redefining
success around MTP parity.

## Validation

- focused loader/protocol/DFlash/compiled-route/generation regression suite:
  passed, with one expected skip
- Ruff on every changed Python file: passed
- both replacement 24-prompt receipts record the recovered
  `pre_norm` / `post_norm` / `local` contract
- staged receipts prove the compiled A3B K1 route executed with zero fallbacks

## Review scope note

This branch contains the complete 17-commit research lineage: 83 files and
roughly 1.48 million added lines, mostly raw benchmark receipts. They are
included so the negative results and protocol corrections remain auditable.
If a smaller review surface is preferred, I can follow with a code-and-tests
only branch while keeping the evidence branch linked.

## Remaining blocker

Fix or explain the authoritative AR versus compiled M1/M2 target-prefix stream
divergence before considering enablement or release.


PhilipJohnBasile:
Hi @youssofal — I wanted to share this as an evidence-first experimental
implementation, not present it as merge-ready.

The encouraging result is that the staged DFlash path reaches about
171 tok/s on Qwen3.6-35B-A3B and matches the existing MTP K1 control on all
24/24 prompts. The important blocker is that both speculative paths match
authoritative AR on only 13/24 prompts, so I have kept the feature opt-in and
the promotion gate red.

The branch includes the full research and correction trail, including the
failed approaches and replacement receipts after discovering that this
artifact's forged MTP contract lived under legacy
`forge_provenance.mtp_contract`. The benchmark now records the resolved
contract and the route actually executed rather than trusting requested
settings.

I would especially value your guidance on two things:

1. Is the compiled M1/M2 target-prefix divergence from `generate_ar` already
   known or expected anywhere in the current runtime?
2. Would you prefer a much smaller code-and-tests-only PR, with these raw
   receipts left on this evidence branch?

No expectation to review the 1.48M-line evidence diff as-is. I am happy to
reshape this into the smallest useful upstream contribution.

youssofal:
read the whole protocol. the blocker: the receipts are greedy-only (24 prompts, greedy, thinking disabled) and dflash=ar equality is 9/24 even there. our line for any speculative path is exact probabilistic speculative sampling at the product sampler (temp 0.6, top_p 0.95, top_k 20), the same bar the MTP path clears with probability-ratio acceptance + residual correction. a greedy-only lane can't ship as a product route, flag-gated or not. if the dflash lane grows correct variable-temperature acceptance, we're interested, and the evidence-gate structure you built is the right harness to prove it with.

PhilipJohnBasile:
Implemented the requested product-sampler correctness lane in `edb6cf4`.

What is now in the branch:

- Target and DFlash both run the product sampler: `temperature=0.6`, `top_p=0.95`, `top_k=20`.
- DFlash samples each row independently on the host and passes the exact sparse proposal law `q` to the engine through the published, pinned `dflash-mlx==0.1.0` API. The branch no longer depends on the unpublished `dflash.model_mlx` package.
- The reference verifier advances the target one scalar AR row at a time, then uses strict probability-ratio acceptance (`u < min(1, p(x)/q(x))`) and residual `p-q` correction. Multi-row capture-commit and staged target-prefix remain diagnostic lanes, not correctness evidence.
- The receipt fails closed unless **every** proposal reports soft `q`, the engine-observed support/probabilities are valid, the expected sampler and acceptance lane actually ran, and a real residual-correction rejection was exercised.
- Hub draft revisions are resolved to immutable commit SHAs. Receipts also record engine HEAD, dirty state, and a dirty-diff digest.
- Logical verify windows, physical verification forwards, and physical rejection-repair forwards are now separate metrics, so the scalar oracle does not understate per-forward cost.
- Exceptional generator exits close target taps. Staged queues are invalidated after rejection/residual repair. The permutation diagnostic refuses under-resolved configurations and still treats marginal non-rejection as inconclusive.

CPU-only validation on the pushed commit:

- full suite: 2,730 collected, all passing with 4 expected skips;
- focused sampled-`q`/DFlash/protocol suite: 137 passing;
- server/public-CLI telemetry selection: 19 passing;
- Ruff, `py_compile`, and `git diff --check`: clean.

I deliberately did not run a model/GPU benchmark because the machine's GPU is occupied by an active training job. Accordingly, the PR remains draft and the product/distribution gate remains **HOLD**: there is no real `.6/.95/20` model receipt or throughput claim yet. The next evidence step is the preregistered GPU A/B with enough independent AR/DFlash samples, an exercised residual path, a powered synthetic falsifier/equivalence tolerance, and a joint-sequence-law check; position-marginal non-rejection alone will not promote the lane.

youssofal:
verdict: not taking this lane into the product tree — closing as research-complete rather than wrong.

the honesty of the red gate is exactly the discipline we want here, and it's also the answer: the product bar for a draft lane is committed-sequence identity with the authoritative model, and by your own receipts DFlash sits at 9/24 (wide) with the staged lane inheriting its MTP-K1 control's 13/24 vs AR. an experimental lane that can't clear its own gate stays research, and a 1.48M-line branch isn't reviewable or carryable in-tree regardless of verdict.

what's preserved and valuable: the negative result, the adversarial protocol corrections, and the corrected artifact-contract evidence — all linkable from this PR. if a future drafter clears spec==AR at depth on real prompts, open a fresh code-and-tests-only PR (the split you already offered) and that gets a real review.

Issue #86, opened 2026-05-30
MTPLX gets stuck when used with Hermes Agent


benmarte:
### mtplx doctor --json

```
{
  "diagnostics": {
    "checks": [
      {
        "command": null,
        "docs_url": "https://ml-explore.github.io/mlx/build/html/install.html",
        "expected": "macOS >= 14.0 on Apple Silicon",
        "fix": "Upgrade to macOS 14+; MLX does not support older macOS.",
        "id": "os.macos_version",
        "observed": "26.5",
        "severity": "error",
        "status": "pass"
      },
      {
        "command": "python3 -c \"import platform; print(platform.machine(), platform.processor())\"",
        "docs_url": "https://ml-explore.github.io/mlx/build/html/install.html",
        "expected": "native arm64 Python, not Rosetta",
        "fix": "Install/use a native arm64 Python. If needed, reinstall via Homebrew arm64 or uv.",
        "id": "python.native_arm64",
        "observed": {
          "machine": "arm64",
          "processor": "arm"
        },
        "severity": "error",
        "status": "pass"
      },
      {
        "command": null,
        "docs_url": "https://ml-explore.github.io/mlx/build/html/install.html",
        "expected": "Python >= 3.10",
        "fix": "Install Python 3.10 or newer.",
        "id": "python.version",
        "observed": "3.13.13",
        "severity": "error",
        "status": "pass"
      },
      {
        "command": "python3 -m pip install mlx",
        "docs_url": "https://ml-explore.github.io/mlx/build/html/install.html",
        "expected": "mlx importable",
        "fix": "Install MLX into this same native Python environment.",
        "id": "mlx.import",
        "observed": {
          "default_device": "Device(gpu, 0)",
          "get_active_memory": 0,
          "get_peak_memory": 0,
          "mlx": "0.31.2",
          "mlx_lm": "0.31.3"
        },
        "severity": "error",
        "status": "pass"
      },
      {
        "command": null,
        "docs_url": null,
        "expected": "estimated peak <= 80% of unified memory; 48 GiB+ recommended",
        "fix": null,
        "id": "resource.memory",
        "observed": {
          "estimated_peak_gib": 35.3,
          "unified_memory_gib": 128.0
        },
        "severity": "warning",
        "status": "pass"
      },
      {
        "command": null,
        "docs_url": null,
        "expected": "free space for model + temp download + safety headroom",
        "fix": "Free disk space or set MTPLX_MODEL_DIR to a larger volume.",
[...]

PR #220, opened 2026-08-02
feat(app): embed isolated Hermes profile sessions

Cyb3rb1ade:
## What

- adds Hermes profile selection in Settings and saved session/agent selection in the native MTPLX UI
- lets a selected profile accept a first prompt immediately and creates the new Hermes agent on first send
- launches an MTPLX-owned `hermes -p <profile> serve --isolated` sidecar for the selected profile
- connects through authenticated gateway RPC for session discovery, creation, resume, prompt streaming, tools, and approval prompts
- persists profile/session selection without changing shared Hermes routing
- includes the approved design and implementation notes

## Isolation and lifecycle guarantees

- applies endpoint, key, and model overrides only to the MTPLX-owned sidecar process
- leaves the Root Gateway, Telegram routing, and foreign Hermes processes untouched
- preserves the selected profile `config.yaml` and `.env` byte-identically
- keeps externally active sessions visible but read-only and offers a fresh session instead
- records exact process identity and ownership before teardown or orphan cleanup
- fails closed for ambiguous routing, ownership, request IDs, and approvals
- restores normal behavior by ending only the marked MTPLX sidecar when switching profiles or exiting MTPLX

## Validation

- `swift test --scratch-path /tmp/mtplx-hermes-textfield-tests --filter Hermes`: 134 passed, 0 failed
- `swift test --scratch-path /tmp/mtplx-hermes-textfield-tests`: 663 executed, 1 skipped, 0 failed
- `swift build --scratch-path /tmp/mtplx-hermes-textfield-build --product MTPLXApp`: passed with Xcode-beta
- `uv build --wheel --out-dir dist`: passed
- `scripts/fresh_venv_smoke.sh`: passed
- `git diff --check upstream/main...HEAD`: passed
- required Python selection: 251 passed, 2 failed because upstream tests still expect the former Qwen default while current `upstream/main` uses Ornith; this PR changes no Python or Python-test files

## Scope

This branch is rebuilt directly on current `upstream/main` and contains only the Hermes app integration, its Swift tests, and the corresponding design/plan documents. Retrieval, embeddings, reranking, packaged-runtime, and local build-environment commits are deliberately excluded.


youssofal:
Review queued for the next cycle. The isolated sidecar (`hermes -p <profile> serve --isolated`) is the right containment; the gateway RPC auth path is the part I want to walk through carefully.

youssofal (changes requested):
This is titled as an app feature but ships an engine policy change: _session_keep_live_refs_for_request hard-returns False for any Hermes client. 2.8.0's session-cache fix routes oversized snapshots through the live-ref lease that this disables, so merging as is reinstates the ~38k committed-frontier freeze for exactly the client class in #86. CI also never ran on the head. Split the app feature from the engine change and the app half can go in.

PhilipJohnBasile:
Superseded by #373.

The replacement is rebuilt from current `main`, keeps the diff app-only, and removes the engine/session-cache policy rejected here. It includes authenticated Hermes sidecar readiness, lifecycle guards, credential redaction, and externally active session handling.

Validation: 145 focused Hermes tests, 774 full Swift tests with 1 existing skip, and `swift build` passed.

Issue #202, opened 2026-07-24
Need ability to have server restart after a failure


NorthernMan54:
### mtplx doctor --json

{
  "diagnostics": {
    "checks": [
      {
        "command": null,
        "docs_url": "https://ml-explore.github.io/mlx/build/html/install.html",
        "expected": "macOS >= 14.0 on Apple Silicon",
        "fix": "Upgrade to macOS 14+; MLX does not support older macOS.",
        "id": "os.macos_version",
        "observed": "26.6",
        "severity": "error",
        "status": "pass"
      },
      {
        "command": "python3 -c \"import platform; print(platform.machine(), platform.processor())\"",
        "docs_url": "https://ml-explore.github.io/mlx/build/html/install.html",
        "expected": "native arm64 Python, not Rosetta",
        "fix": "Install/use a native arm64 Python. If needed, reinstall via Homebrew arm64 or uv.",
        "id": "python.native_arm64",
        "observed": {
          "machine": "arm64",
          "processor": "arm"
        },
        "severity": "error",
        "status": "pass"
      },
      {
        "command": null,
        "docs_url": "https://ml-explore.github.io/mlx/build/html/install.html",
        "expected": "Python >= 3.10",
        "fix": "Install Python 3.10 or newer.",
        "id": "python.version",
        "observed": "3.14.5",
        "severity": "error",
        "status": "pass"
      },
      {
        "command": "python3 -m pip install mlx",
        "docs_url": "https://ml-explore.github.io/mlx/build/html/install.html",
        "expected": "mlx importable",
        "fix": "Install MLX into this same native Python environment.",
        "id": "mlx.import",
        "observed": {
          "default_device": "Device(gpu, 0)",
          "get_active_memory": 0,
          "get_peak_memory": 0,
          "mlx": "0.31.2",
          "mlx_lm": "0.31.3"
        },
        "severity": "error",
        "status": "pass"
      },
      {
        "command": null,
        "docs_url": null,
        "expected": "estimated peak <= 80% of unified memory; 48 GiB+ recommended",
        "fix": "Close other heavy apps or use a smaller model/profile.",
        "id": "resource.memory",
        "observed": {
          "estimated_peak_gib": 35.3,
          "unified_memory_gib": 24.0
        },
        "severity": "error",
        "status": "fail"
      },
      {
        "command": null,
        "docs_url": null,
        "expected": "free space for model + temp download + safety headroom",
[...]

PR #221, opened 2026-08-02
feat(app): add bounded daemon crash recovery

PhilipJohnBasile:
## Summary

Implements the narrow app-managed supervision slice accepted in #202.

- Adds an opt-in automatic daemon restart setting (off by default).
- Retries abnormal exits at most three times with bounded exponential backoff and a visible circuit-breaker state.
- Makes explicit Stop and disabling recovery cancel scheduled or in-flight retries and reap only the exact app-owned process family.
- Keeps restart recipes, including the API key, in memory only and releases them after clean exit, disable, or exhaustion.
- Preserves lifecycle ownership across Pi, Hermes, and OpenCode handoffs using per-launch UUID or PID-plus-launch-date identities, including exact cleanup after restarts.
- Removes one-use Hermes command and receipt artifacts as soon as a receipt is collected so launch secrets are not left on disk.
- Exposes recovery status/count in the app and guards transport, fan, metrics, and client state against stale async completions.

An already-running daemon adopted from a previous app session is intentionally shown as unprotected. The user must start a fresh app-managed daemon before automatic recovery can own it safely.

## Scope

This does **not** claim to fix the underlying Metal out-of-memory crash, persist daemon logs, or add memory telemetry. Those are separate parts of #202 and need their own investigation and validation.

## Verification

CPU-only / no model weights:

- Swift parser: all 10 changed/new Swift files
- `git diff --check`
- `DaemonSupervisorTests`: 35 passed
- `TerminalHandoffLeaseTests`: 12 passed
- Full `MTPLXApp` suite: 593 tests, 1 skipped, 0 failures
- Six high-risk lifecycle/reparenting races: 6 passed, repeated three times
- Post-test process scan: no leaked supervisor or handoff processes

The final diff received independent Terra and Sol review approval before commit.

Closes the app-managed restart slice of #202; leaves the crash root cause, persistent logging, and memory telemetry open.

### DIFF
--- a/apps/MTPLXApp/Sources/MTPLXAppCore/Models/AppConfiguration.swift
+++ b/apps/MTPLXApp/Sources/MTPLXAppCore/Models/AppConfiguration.swift
@@ -108,6 +108,10 @@ public struct MTPLXAppConfiguration: Codable, Equatable, Sendable {
     public var streamSnapshotIntervalMs: Int
     public var performanceLock: Bool
     public var launchDaemonOnOpen: Bool
+    /// Opt-in app-managed recovery after an abnormal daemon exit. It applies
+    /// to launches created in this app session; the launch command and API key
+    /// remain process-memory-only in DaemonSupervisor.
+    public var automaticDaemonRestart: Bool
     /// When on, the app launches Hermes in auto-approve ("YOLO") mode so
     /// the agent runs tools without prompting. Off makes Hermes ask for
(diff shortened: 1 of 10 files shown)


youssofal:
Review queued for the next cycle. This matches the slice accepted in #202 (opt-in, three bounded retries, visible circuit breaker, in-memory recipes); the process-family reaping is the part I will verify first.

PR #222, opened 2026-08-02
Laguna S-2.1: mlx.fast XS2.1 port — residual+router fusion + async decode scheduling (+5.8% decode, digest-exact)

davidtai:
## What

Ports the mlx.fast "Laguna XS2.1" challenge optimized runtime to Laguna S-2.1 (affine
oQ4e), as a standalone alternative decode+prefill lane (`mtplx/laguna_alt_step.py`)
benchmarked head-to-head against the `install_from_env` reference lane. Every challenge
kernel is ported and measured; each sits behind an `AltConfig` flag (default off, fail-loud)
so nothing is silently skipped and the two lanes stay drop-in comparable.

## Results (decode, B=1, ctx 1024 / decode 96 — the reference's shape; digest-exact)

| lane | tok/s | Δ |
|---|---|---|
| reference (`install_from_env` + `LagunaCompiledLane`) | 67.3 | — |
| **alt (D1 + async)** | **71.2** | **+5.8%** |

Token digest identical to the reference across all runs (`9098436fbc29879b`).

## Optimization ledger — worked / tried / not tried

Every verdict below is from a fair, S-2.1-shaped vehicle. Two conclusions were re-tested
after a self-audit found weak vehicles (see "Fair-vehicle re-tests").

### ✅ Worked — shipped
| opt | Δ | note |
|---|---|---|
| **D1** residual+RMSNorm+router-GEMV fusion | +1.0% | folds router GEMV into the post-attn norm dispatch, 47 sparse layers; bit-exact |
| **S1** async-eval decode scheduling | +4.0% | full-state `async_eval` per token → host/GPU overlap; value-preserving |
| **D1 + S1 (best)** | **+5.8%** | 71.2 vs 67.3, digest-exact |

### ⏭️ Tried — no win (stock retained)
| opt | verdict |
|---|---|
| D6 SDPA-vector (group-3 GQA), decode | −1.6% in the compiled runtime. (Isolated, the op is faster at N≤8k; the runtime loss is `mx.compile` graph-composition of a custom kernel. See re-tests: the KV-reuse *technique* is real, just not as a userland kernel.) |
| D7–D9 affine MoE SwiGLU-QMV, decode (per-token) | −25%; inherent — at B=1 the per-(token,expert) layout lights 10 threadgroups (GPU-underfilled); stock `gather_qmm` amortizes |
| affine MoE gather-GEMM, prefill | no lever; stock `gather_qmm` is MLX's steel MMA GEMM and already handles sparse/empty experts for free (see RUNSKIP re-test) |
| D14 lm-head top-1 (qmv8) | −0.7% (though digest-exact); head read dominates, top-k machinery adds overhead over argmax |
| interval async ladder (1,7,15,…) | worse than async-every-step |
| D1 at prefill | −5% (router becomes a GEMM; per-row fusion loses to stock GEMM) |
| counting-sort route table | no win — loses ~15% at small M, ties at M=10240 (host-dispatch-bound); exact stable-argsort drop-in but negligible vs the GEMM |

### ✅ Active via installed reference kernels (in the runtime, digest-matched)
D2/D3 qk-norm+rope · D5 attn-gate · D10 moe-combine · D12 router · gate_up fusion
(loads default-on as `FusedGateUpSwitchGLU`).

## Fair-vehicle re-tests (self-audit corrections)

A review flagged that two negative verdicts were tested in weak vehicles; both were re-run:

- **GQA KV-reuse — CORRECTED to a real lever.** The original no-go used a *single-pass* SDPA
  that under-parallelizes at long N. In a fair **2-pass KV-split (flash-decode)** vehicle,
  GROUP≈3 reuse **decisively beats the same-tiling per-head control and the win grows with N**
  (~2.6× at 128K; tracks the KV-traffic ratio → genuinely bandwidth). Interior optimum at
  **GROUP≈3** (max reuse G=gqa tips ALU/occupancy-bound). *But* no hand kernel beats stock
  `mx.fast.SDPA` (asymptotes ~1.7× slower — MLX's flash-decode is too tuned to replace). So
  KV-reuse is validated as a technique but is realizable only by injecting GROUP≈3 reuse
  **inside MLX's flash-decode** — an ml-explore/mlx candidate (long-context/high-GQA), out of
  scope for this PR.
- **Prefill MoE / RUNSKIP — no headroom (confirmed, correct reasoning).** Real routing is
  imbalanced (25–45% empty experts/layer, not the 0 a synthetic uniform test showed), but
  stock `gather_qmm` **already handles empties for free**: at fixed token count it gets
  *faster* as the empty fraction rises (−10% at 25% empty, −23% at 45%), and empty placement
  is irrelevant — it only tiles present sorted segments. RUNSKIP would skip already-free work.

## Root finding

XS2.1 shipped as **NVFP4** (`poolside/Laguna-XS-2.1-NVFP4-mlx`), which MLX has no native
kernel for — so the challenge's hand kernels had no tuned stock rival. Re-expressed for
S-2.1's **affine oQ4e**, they race MLX's already-tuned stock primitives (`gather_qmm` steel
MMA / flash SDPA / argmax) and lose. The transferable levers are the **quant-agnostic** ones
— cross-op **fusion (D1)** and **scheduling (S1)** — which is what shipped (+5.8%).

Repro: `tests/test_laguna_alt_step.py`; per-kernel scripts + receipts under
`docs/laguna-mlxfast-port/`.

## No-regression
Additive only — new module + kernels behind default-off flags + docs/tests. The reference
lane and every existing installer are untouched.

### DIFF
--- a/docs/laguna-mlxfast-port/PORT_KERNEL_LEDGER.md
+++ b/docs/laguna-mlxfast-port/PORT_KERNEL_LEDGER.md
@@ -1,0 +1,71 @@
+# Laguna S-2.1 port — completeness matrix (every challenge kernel ported + tested)
+
+This replaces the earlier "active via installed reference kernels" hand-wave. Every
+challenge optimization is now ported as the **challenge's own implementation, reshaped to
+S-2.1's geometry** (not MTPLX's installed kernel, not "argued-equivalent"), CPU-validated
+bit-exact/allclose against a pure-mx reference and the stock module, and measured under the
+GPU flock. Δ column filled from the flock window (`f04_batch_check_runner.py`, one guarded
+window, 13 checks).
+
+## S-2.1 shape corrections the ports had to make (vs the challenge's XS2.1 geometry)
+
+These are the concrete "customized to the Laguna S shape" changes — each was a real
+divergence from the challenge Swift, caught during the port:
+
+| kernel | XS2.1 (challenge) | S-2.1 (this port) |
+|---|---|---|
+| D2 YaRN mscale | 1.3465… (= factor 32) | **1.4852030263919618** (0.1·ln(128)+1, = attention_factor) |
+| D2 rot dims | — | 64 of 128 (partial-rotary 0.5), θ=500000, factor 128 |
+| D3 sliding heads | 64 q | **72 q** / 8 kv, full-rotary 128, θ=10000 |
+| D5 o_proj quant | group-32 INT8 only | **affine gs64, 5- and 8-bit** (all S-2.1 layers), inline gate (occupancy) |
+| D9 merged slots | top-8 → 9 slots | **top-10 → 11 slots** (10 routed 4-bit + 1 shared 8-bit) |
+| D11 dense layer-0 | plain BF16, 2048/8192 | **mixed affine** (gate/up 5-bit, down 6-bit, gs64), 3072/12288, 48 layers |
+| D12 router sort | top-8 bitonic | **top-10** Batcher bitonic over 256, +pow2-experts guard (latent bug fixed) |
+| D10 combine reduce | in-order bf16 (exact at k=8) | **TY=8 col_reduce order** to stay bit-exact at k=10; scale 2.5 pre-baked upstream |
+| P1 batched rope | — | length-T positions vector (batched-rope offset trap), mscale 1.4852 |
+
+## Port + test status — MEASURED (flock window `bd7tgynzj`, 2026-08-02)
+
+Decision lane = **chained** (B=1 serial decode link) for decode, **queued** for prefill.
+For the `port | mtplx | stock` triples the numbers are µs/call → **lower is faster**.
+
+| id | kernel | GPU correctness | isolation speed (decision lane) | verdict |
+|---|---|---|---|---|
+| D2 | qk-norm + YaRN rope (full) | **bit-exact** vs stock | chained 25.82 vs stock 19.57 | ❌ LOSS 0.76× |
+| D3 | qk-norm + rope (sliding) | **bit-exact** vs stock | chained 17.01 vs stock 5.73 | ❌ LOSS 0.34× |
+| D4 | input-norm + QKV + gate | 8-bit allclose ✓; **5-bit diverges** vs stock (0.11) | queued 0.534× | ❌ LOSS + 5-bit gap |
+| D5 | gated o-proj | 8-bit allclose ✓; **5-bit allclose FAIL** (0.09) | chained 252–391 vs stock 40 | ❌ LOSS ~6–10× |
+| D7 | shared-expert SwiGLU-QMV | ALL PASS | queued x0.48–0.70 | ❌ LOSS |
+| D9 | 11-slot merged SwiGLU-QMV | ALL PASS | queued x0.48–0.65 | ❌ LOSS |
+| D10 | MoE combine tail (+residual) | **bit-exact** | chained rows=1: 16.21 vs mtplx 11.26 vs stock 14.25 | ❌ LOSS at decode (µs epilogue) |
+| D11 | dense layer-0 MLP (5/6-bit) | pass | stock/fused 0.206× | ❌ LOSS ~5× (~2% share) |
+| D12 | router top-10 (bitonic) | **0 selection flips** | chained rows=1: 30.53 vs mtplx 21.20 vs stock 27.22 | ❌ LOSS vs both (shape-dependent) |
+| D13 | embed + rope-atlas | n/a (pure memcpy) | challenge's own −0.23…−0.7% | ⊘ NO LEVER (no distinct decode op) |
+| **P1** | **prefill qk-norm + rope** | compile bug (`T` shadow) → **FIXED** → **bit-exact** (max\|d\|=0.0, rows distinct) | **queued 1.48× (FULL)** / 0.96× (sliding); per-call ~1.0–1.08× | ✅ **mild WIN** (FULL) / neutral (sliding) |
+| P2 | steel flash-attention (prefill) | ctx1024 PASS; **ctx8192 FAIL** | ratio x0.05–0.16 (full), x0.99 (sliding 8k) | ❌ LOSS (no `mlx::steel` MMA) |
+| P3 | prefill router top-10 | **0 flips** | queued 0.266× @M1024, 1.021× @M10240 | ❌ LOSS @1024 / neutral @10240 |
+| **P5** | **prefill MoE combine tail** | **bit-exact** (max\|d\|=0.0) | **queued 4.745× / per-call 1.603×** | ✅ **WIN** (isolation) |
+
+**Scoreboard:** **2 new bit-exact isolation WINS** (P5 MoE-combine, P1 qk-norm+rope), 0
+decode wins, 11 losses/neutral, 1 no-lever. Both wins are **prefill fusions** and both are
+**bit-exact** (digest-safe). This sharpens the D1 thesis: cross-op **fusion** transfers to
+S-2.1's affine geometry (D1 decode, P1+P5 prefill); hand **GEMV/attention** kernels do not,
+because they race MLX's `mlx::steel` MMA (`gather_qmm`, flash-SDPA) which `mx.fast.metal_kernel`
+cannot reach. The prefill combine/rope wins exist precisely where stock has **no MMA-backed
+path** to beat. End-to-end ceilings are bounded by prefill's share (0.25 of score) and each
+op's slice of prefill, so they are validated op-level wins layered on the shipped D1+S1
+(+5.8% decode), not headline replacements.
+
+Already shipped (unchanged): **D1** residual+RMSNorm+router-GEMV fusion (+1.0%), **S1**
+async-eval scheduling (+4.0%), best **+5.8%** (71.2 vs 67.3, digest-exact).
+Already measured losers (prior fair-vehicle runs): **D6** SDPA-vector −1.6%, **D8** routed
+per-token SwiGLU-QMV −25%, **D14** lm-head top-1 −0.7%, **P4** prefill gather-GEMM (stock
+handles empties free).
+
+## Recurring root cause (why the hand kernels are expected to lose on affine)
+`mx.fast.metal_kernel` compiles a **standalone snippet** and cannot reach MLX's `mlx::steel`
+16×16 simdgroup-matrix (MMA) headers — the exact machinery that makes stock `gather_qmm` /
+flash-SDPA fast. On XS2.1's NVFP4 (no native MLX kernel) the challenge's hand kernels had no
+tuned rival; re-expressed on S-2.1's **affine oQ4e** they race MLX's tuned stock and the only
+transferable wins are quant-agnostic **fusion (D1)** + **scheduling (S1)**. The per-kernel Δ
+below either confirms this or finds an exception — measured, not assumed.
--- a/docs/laguna-mlxfast-port/PORT_KERNEL_LEDGER.md
+++ b/docs/laguna-mlxfast-port/PORT_KERNEL_LEDGER.md
@@ -56,6 +56,35 @@ path** to beat. End-to-end ceilings are bounded by prefill's share (0.25 of scor
 op's slice of prefill, so they are validated op-level wins layered on the shipped D1+S1
 (+5.8% decode), not headline replacements.
 
+## Prefill runtime A/B — P5 wired end-to-end (2026-08-02)
+
+The isolation "5.1× queued" for P5 was measured vs the **naive** combine. Wired into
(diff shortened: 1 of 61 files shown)


davidtai:
### ✅ D1 — fused residual+RMSNorm+router-GEMV (**+1.0%**)

Folds the MoE router GEMV into the post-attention residual+norm dispatch across the
47 sparse layers. Kernel bit-exact at the S-2.1 shape (residual / norm / logits 0.0
diff, top-10 overlap 10/10).

Benchmark (B=1, ctx 1024 / decode 96, one window, all digest-match):

| lane | ms/step | tok/s | digest |
|---|---|---|---|
| reference | 14.81 | 67.5 | `9098436fbc29879b` |
| alt[stock] | 14.87 | 67.2 | match |
| **alt[D1]** | **14.72** | **67.9** | match |

Beats the reference's separate `kernel_router_gemv` even paying for a separate
argpartition top-k. Receipt: `docs/laguna-mlxfast-port/benchmarks/laguna-alt-ab-d1-20260801-200550.json`.

davidtai:
### ✅ S1 — async-eval decode scheduling (**+4.0%**)

`mx.async_eval` the full step state (token + offset + ring + KV leaves) each token,
so the host encodes step N+1's graph while the GPU runs step N. Value-preserving —
digest identical to per-step sync.

Benchmark (B=1, ctx 1024 / decode 96, one window, all digest-match):

| lane | ms/step | tok/s | digest |
|---|---|---|---|
| reference | 14.86 | 67.3 | `9098436fbc29879b` |
| alt[stock \| sync] | 14.81 | 67.5 | match |
| **alt[stock \| async]** | **14.24** | **70.2** | match |
| **alt[D1 \| async]** | **14.05** | **71.2** | match |

**D1 + async together = 71.2 vs 67.3 = +5.8%**, token-for-token identical. The
interval ladder (async every ~8 steps) measured *worse* (68.6 vs 70.0). Receipt:
`docs/laguna-mlxfast-port/benchmarks/laguna-alt-ab-s1-20260801-201437.json`.

davidtai:
### Correction — prefill MoE claim (thanks to a reviewer catch)

To be precise about what was measured: the affine MoE kernel benched across T=1…1024 is the
challenge's **decode** per-token SwiGLU-QMV. Benched at prefill it loses simply because it
re-reads each expert's weights per routed token — the wrong algorithm for prefill.

The challenge's actual **prefill** MoE kernel is different: a **grouped gather-GEMM
(split-K / RUNSKIP / SM-16 tiling)** — the same class as stock `mx.gather_qmm(sorted_indices)`,
which `SwitchGLU` already uses. That kernel was **not directly ported**, so "no prefill MoE
lever" is *expected* (a grouped gather-GEMM should land ≈ tuned stock) but **not yet proven**.
A port of the real gather-GEMM is in progress to confirm; I'll post the numbers here.

Context: XS2.1 shipped as NVFP4 (`poolside/Laguna-XS-2.1-NVFP4-mlx`), which has no native MLX
kernel — so the challenge's hand kernels had no tuned stock rival there. On S-2.1's affine
oQ4e they race MLX's already-tuned `gather_qmm`/SDPA, a much higher bar.

The decode headline (D1 + S1, **+5.8%**, digest-exact) is unaffected.

davidtai:
### Prefill MoE gather-GEMM — gap closed (ported + measured, no lever)

Ported the challenge's *actual* prefill MoE kernel. Key finding: `fp_gather_qmm_rhs_nax` is
**a fork of MLX's own steel tiled gather-GEMM** — the exact kernel `mx.gather_qmm(sorted_indices=True)`
dispatches (16×16 `simdgroup_matrix` MMA, BK double-buffered) — with micro-levers (RUNSKIP,
wider loads, register prefetch) that default OFF. So the challenge never hand-wrote a GEMM to
beat stock; it forked stock and toggled knobs. Re-expressed for affine 4-bit gs128, measured
vs stock at prefill (256 experts, top-10, moe_inter 1024, hidden 3072, queued lane):

| T | M (sorted rows) | hand (ms) | stock (ms) | ratio | win? |
|---|---|---|---|---|---|
| 256 | 2,560 | 32.0 | 3.19 | 10.0× | no |
| 512 | 5,120 | 65.1 | 3.72 | 17.5× | no |
| 1024 | 10,240 | 128.0 | 5.61 | 22.8× | no |

allclose PASS at every T. A hand `mx.fast.metal_kernel` uses scalar FMA and can't match the
hardware MMA units; RUNSKIP is inert (0 empty experts at top-10/256), split-K is slower.

Three independent lines now agree there is **no prefill MoE lever** on affine: (1) the challenge
kernel is itself stock + off-by-default micro-levers, (2) the challenge team's own notes record
prefill MoE as "NO HEADROOM" (staging "shelved-regressed", split-K +0.18%), (3) this direct
affine measurement (−10…23×). Keep stock `gather_qmm` on the prefill path. Kernel:
`mtplx/kernels/laguna_moe_gather_gemm.py`; check: `docs/laguna-mlxfast-port/bench/laguna_moe_gather_check.py`.

davidtai:
## Completeness pass — every challenge kernel now ported to S-2.1 & measured

Earlier this PR shipped **D1 + S1 (+5.8% decode, digest-exact)** and marked the rest as
"active via installed kernels" or "argued-equivalent." That was a gap: those weren't the
**challenge's own** kernels reshaped to S-2.1 — they were MTPLX's installed substitutes, and
several challenge kernels (D4/D7/D9/D11/D13, prefill P1/P2/P3/P5) were never attempted.

This pass closes it. **Every** challenge optimization is now ported as the challenge's own
implementation, reshaped to S-2.1's real geometry, CPU-validated against a pure-mx reference
+ the stock module, and **measured under the GPU flock** (`388f5d22`; raw receipts in
`docs/laguna-mlxfast-port/benchmarks/laguna-kernel-checks-flock-20260802.txt`).

### S-2.1 shape corrections each port had to make (vs the challenge's XS2.1)
| kernel | XS2.1 (challenge) | S-2.1 (this port) |
|---|---|---|
| D2 YaRN mscale | 1.3465… (factor 32) | **1.4852030263919618** (0.1·ln 128 + 1) |
| D3 sliding heads | 64 q | **72 q** / 8 kv, θ=10000 |
| D5 o_proj quant | group-32 INT8 | **affine gs64, 5- & 8-bit** (inline gate) |
| D9 merged slots | top-8 → 9 | **top-10 → 11** (10 routed 4-bit + 1 shared 8-bit) |
| D11 dense-0 | plain BF16, 2048/8192 | **mixed 5/6-bit affine**, 3072/12288 |
| D12 router | top-8 bitonic | **top-10** + pow2-experts guard (latent mis-reduce fixed) |
| D10 combine | in-order bf16 (exact @k=8) | **TY=8 col_reduce order** (bit-exact @k=10) |
| P1 rope | — | length-T positions vector (batched-rope offset trap) |

### Measured verdicts (isolation; decision lane = chained for decode, queued for prefill)
| id | kernel | GPU correctness | speed (decision lane) | verdict |
|---|---|---|---|---|
| D2 | qk-norm+YaRN (full) | bit-exact | chained 25.8 vs stock 19.6 µs | ❌ 0.76× |
| D3 | qk-norm+rope (sliding) | bit-exact | chained 17.0 vs 5.7 µs | ❌ 0.34× |
| D4 | input-norm+QKV+gate | 8-bit ✓, 5-bit diverges | queued 0.53× | ❌ |
| D5 | gated o-proj | 8-bit ✓, 5-bit allclose FAIL | chained 253–391 vs 40 µs | ❌ ~6–10× |
| D7 | shared SwiGLU-QMV | pass | queued 0.48–0.70× | ❌ |
| D9 | 11-slot merged SwiGLU-QMV | pass | queued 0.48–0.65× | ❌ |
| D10 | MoE combine tail | bit-exact | chained rows=1 16.2 vs stock 14.3 / mtplx 11.3 µs | ❌ (µs epilogue) |
| D11 | dense-0 MLP (5/6-bit) | pass | stock/fused 0.206× | ❌ ~5× (~2% share) |
| D12 | router top-10 (bitonic) | 0 selection flips | chained rows=1 30.5 vs 27.2 / 21.2 µs | ❌ (shape-dependent) |
| D13 | embed + rope-atlas | n/a (pure memcpy) | challenge's own −0.23…−0.7% | ⊘ no lever |
| **P1** | **prefill qk-norm+rope** | **bit-exact** (fixed a `T`-shadow compile bug) | **queued 1.48× (FULL)** / 0.96× sliding | ✅ **mild WIN** |
| P2 | steel flash-attn (prefill) | ctx1024 ✓, **ctx8192 FAIL** | ratio 0.05–0.16× | ❌ (no `mlx::steel` MMA) |
| P3 | prefill router top-10 | 0 flips | queued 0.27× @M1024 / 1.02× @M10240 | ❌ / neutral |
| **P5** | **prefill MoE combine tail** | **bit-exact** | **queued 5.1× / per-call 1.79×** | ✅ **WIN** |

### What it means
- **Two new bit-exact wins, both prefill fusions** — **P5** (MoE-combine, removes the
  `[M,K,H]` intermediate) and **P1** (qk-norm+rope, FULL family). Both digest-safe.
- **Everything else loses or ties**, confirming and sharpening the root cause:
  `mx.fast.metal_kernel` compiles a standalone snippet and **cannot reach MLX's `mlx::steel`
  16×16 MMA** (`gather_qmm`, flash-SDPA). On XS2.1's NVFP4 the challenge's hand kernels had no
  tuned rival; re-expressed on S-2.1's **affine oQ4e** they race MLX's tuned stock and lose.
  The wins are exactly where stock has **no MMA-backed path** to beat — cross-op **fusion**
  (D1 decode, P1/P5 prefill) and **scheduling** (S1). That is the transferable lever.
- Notes worth keeping honest: D4/D5's affine GEMV **diverges from stock at 5-bit** (matches
  its own reference; not chased since both lose on speed), and **P2 fails correctness at
  ctx 8192** — both moot for shipping given the decisive speed losses, but recorded.

P1/P5 are **op-level** wins; their end-to-end ceiling is bounded by prefill's 0.25 score
weight × each op's slice, so they layer on the shipped D1+S1 rather than replacing the
headline. Happy to wire P1+P5 into the prefill lane for an end-to-end A/B if that's wanted.

Full per-kernel ledger: `docs/laguna-mlxfast-port/PORT_KERNEL_LEDGER.md`. Every check script
is under `docs/laguna-mlxfast-port/bench/`.

davidtai:
## Prefill follow-up: P5 wired end-to-end + prefill tok/s now tracked

Two things landed (commit `037f317e`).

### 1. Prefill tok/s is now a tracked metric
All bench/laguna benches (batched `run_cell`, compiled-lane, alt-lane, and the AB
comparison line) now report **`prefill_tokps`** alongside decode — warmup-excluded (one
throwaway prefill compiles the forward_ar kernels so the timed pass is clean) — in both the
JSON receipts and the printed summaries.

### 2. P5 MoE-combine wired end-to-end — bit-exact, but not a robust win
The isolation "5.1× queued" was vs the **naive** combine. Wired into `alt_prefill_forward`
(both a D1 path and a new D1-free path so P5 isn't dragged by D1's prefill penalty) and A/B'd
across a context sweep against the shipped **reference** — which already installs
`kernel_moe_combine` on 47 layers — every cell **digest-exact**:

| ctx | ref | P5 (D1-free) | P5/ref |
|---|---|---|---|
| 1024 | 1564 | 1563 | 1.000× |
| 8192 | 1276 | 1298 | 1.018× |
| 16384 | 904 | 906 | 1.002× |
| 32768 | 492 | 511 | 1.039× |

**Verdict: bit-exact, never breaks (through ctx 65536, peak 84.5 GiB), but not a robust
prefill speedup.** Vs the already-fused reference the gain is ~1–4% at ctx ≥ 8k and within the
box's cross-run variance (ref@32k swung 582→492 between sweeps under contention). Same lesson
as D10 at decode: the combine is a small share and the reference already fuses it. There *is*
a real **size crossover** — P5 loses below ~8k (the `[M,top_k,hidden]` intermediate it removes
is cheap there) — so it's wired behind `p5_prefill_moe_tail` + a `prefill_min_tokens` gate
(recommend 8192) and left **default-off** as a validated, at-worst-neutral, digest-safe option.

**P1 (prefill qk-norm+rope)** is expected to follow the same pattern: its isolation win was vs
the `mx.fast` chain, but the runtime already installs `kernel_qk_rope` on all 48 layers, so vs
the reference it's very likely neutral too. Not wired (it needs an attention-path rewrite);
happy to confirm empirically if wanted, but the reference-already-fuses pattern is consistent.

Net: the ported prefill kernels don't beat the reference's already-fused prologue/epilogue
path — the transferable levers remain cross-op fusion (D1 decode) + scheduling (S1). The
concrete prefill deliverable here is the **tracking** + the honest measurement.
Receipts: `docs/laguna-mlxfast-port/benchmarks/laguna-p5-prefill-sweep-*.txt`.

PR #236, opened 2026-08-04
Release 2.5.2: settle the compiled growth handoff and raise the MLX floor

youssofal:
## Summary

Hotfix for the 2.5.1 long-response slowdown on Optimized Speed V2.

Long unbounded responses cross the compiled verifier's 512-token growth
reserve and hand verification to the eager path for the rest of the response.
The handoff restored the cache containers but did not settle the GPU work the
compiled path had scheduled, so every later verify step absorbed the old
dependency chain. On the reported 12k-token generation, decode opened at 59
tok/s and ended near 30, with 215 of 312 seconds spent in verify hidden-state
evaluation.

The fix settles all cache and recurrent state exactly once at the growth
ownership boundary, clears shadow and held references before eager takes
ownership, and exposes handoff telemetry in the compiled verifier stats.

## Evidence

- Field repro and root cause: request-log records show the collapse
  with acceptance at 96/89/80 and zero fallbacks; `growth_demotions=1` on
  every affected request.
- Controlled runs on the same V2 artifact, verified max fans, pinned seed,
  identical 12,288-token trajectories: 2.5.1 decodes its final 256-token
  window 17% slower than its early windows on a fresh server (live app
  servers reported roughly half the opening speed); the fixed build's
  closing windows run as fast as or faster than its opening ones. Output
  tokens, acceptance counters, and peak memory are identical across builds.
- Handoff engagement is proven live: exactly one materialization per long
  request, settled in single-digit milliseconds.

## Verification

- `tests/test_graphbank_compiled_verify.py`: 55 passed (includes a new
  synthetic hybrid growth-handoff regression and a pinned single-handoff
  exactness test).
- `tests/test_server_openai.py`: 281 passed.
- Compiled wiring + sustained integration: 41 passed, 1 skipped.
- Ruff clean, wheel/sdist build clean.

## Also in this release

The MLX floor moves to `>=0.32`. Fresh installs already resolve mlx 0.32.0
(exactness-gated 2026-07-11), but existing runtime venvs never left 0.31.2
because the bootstrapper upgrades with pip's only-if-needed strategy. The
floor raise converges every install to the stack new installs already run.

## Scope

The once-per-request growth handoff, the MLX floor, and release metadata.
Normal compiled ticks, short requests, model policies, catalogs, and sampler
are untouched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,40 @@ All notable user-facing changes to MTPLX. The format is based on
 [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); versions follow
 [Semantic Versioning](https://semver.org/).
 
+## [2.5.2] - 2026-08-04
+
+Hotfix for a long-response slowdown that shipped in 2.5.1 with Optimized
+Speed V2 as the recommended coding model.
+
+### Fixed
+
+- Long single responses no longer slow down and stutter partway through. The
+  compiled verifier hands generation to the eager path after its growth
+  reserve is exhausted, and that handoff carried unfinished GPU work into the
+  rest of the response. Every later token paid for the old work, so decode
+  throughput fell steadily on responses past a few thousand tokens. The
+  reported 12,000-token response opened at 59 tokens per second and
+  ended near 30. With the fix, the same seeded generation no longer
+  decays: the closing windows now run as fast as or faster than the
+  early ones.
+- The handoff now settles all cache and recurrent state exactly once at the
+  ownership boundary and releases the compiled references before eager
+  decoding continues. Handoff telemetry is exposed in the compiled verifier
+  stats.
+
+### Changed
+
+- The minimum MLX version is now 0.32. Fresh installs already resolved MLX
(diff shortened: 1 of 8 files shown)


reneleonhardt:
Are the amazing new Qwen3.6-27B weights
https://huggingface.co/prism-ml/Ternary-Bonsai-27B-mlx-2bit
https://huggingface.co/prism-ml/Bonsai-27B-mlx-1bit
already supported for a fraction of memory usage?

```shell
error: model cannot run with MTPLX
model: prism-ml/Ternary-Bonsai-27B-mlx-2bit
tier: architecture-compatible-but-unverified
runtime: missing-mtp-weights
reason: This model's config advertises MTP, but MTPLX did not find runnable MTP weights in the folder. mtplx_runtime.json is optional metadata; the blocker is missing MTP weights.
mtp: exists=false, tensors=0, gate=false
fix: use a complete model with matching MTP weights, or build and verify one from its original source with mtplx forge. MTPLX does not attach arbitrary sidecars because config/tensor checks cannot prove their trunk lineage.
```

PR #248, opened 2026-08-10
Add Muse-Glimmer-30B arch + dflash speculative backend

davidtai:
## Summary

Adds **Muse-Glimmer-30B** (a Gemma-family multimodal; the text tower is served) as a first-class MLX architecture in MTPLX, and a generic **dflash** block-diffusion **speculative-decoding backend** that drafts for it — wired into the existing MTP serving route so speculation is transparent to clients.

Two independent pieces, one PR:
1. the Muse-Glimmer text-tower arch + q4-affine quant, and
2. the dflash drafter + its integration into the MTP route.

---

## 1. Muse-Glimmer text tower

Arch reverse-engineered from llama.cpp `src/models/muse-glimmer.cpp` (there is no transformers modeling / no PyTorch reference). It is Gemma-3-like but custom:

- **sigmoid gated attention** (`self_attn.gate_proj`, applied before `o_proj`),
- **parameter-free QK-norm** with `qk_scale_factor` folded onto Q (k-norm identity),
- **NoPE on the global/full-attention layers** (RoPE θ=500000 only on sliding layers),
- Gemma `(1+w)` **sandwich norms** (post-norm eps 1e-8),
- **RMS-normed embeddings** (not ×√hidden), SiLU SwiGLU, `logit_scale` + tanh softcap.

52 layers, hidden 6656, GQA 32/2, head-dim 128, vocab 202048, sliding-window 2048 pattern-4, untied lm_head.

**Registration.** A new AR arch needs both an `ARCHITECTURE_CATALOG` entry and a `_passes_family_runtime_gate` case, otherwise inspect demotes it to `unsupported` and serve refuses ("no MTP head"). The model def is vendored and registered via a `sys.modules` shim (`muse_glimmer_patch.py`). A qwen3_vl-style multimodal wrapper drops the vision stack and accepts `input_embeddings` (`vendored_muse_glimmer.py`).

**q4-affine quant** mirrors unsloth UD-Q4_K_XL (4-bit gs32 body; 5-bit lm_head + attn-output layers 45–51; norms fp) → **5.06 bpw / 16 GB**. Weights (public): **[OpensourceWTF/Muse-Glimmer-30B-MLX-q4](https://huggingface.co/OpensourceWTF/Muse-Glimmer-30B-MLX-q4)**.

Quality (base-model completion, greedy, through the real MTPLX `/v1/completions` serve path — batch-decode, so indentation is correct):

| | pass@1 |
|---|---|
| HumanEval-164 | **47.6%** |
| HumanEval+ (stricter tests) | **43.3%** |
| MBPP-378 | **68.8%** |
| MBPP+ (stricter tests) | **56.9%** |

**Decode optimization (measure-first, ledger in `docs/`).** Decode is bandwidth-bound at ~82% of the M5 roofline. The one integrated win is a **bit-exact QKVG fusion** (q/k/v/gate → one `quantized_matmul`; +4.8%, max|Δ|=0, default on). An optional decode-only fused dense-SwiGLU kernel (+5.2%, quality-parity) is env-gated (`MG_MLP_KERNEL`) with a graceful fallback so a fresh checkout runs the bit-exact default. Shape-optimized hand kernels beat stock qmm at M=1 but their dispatch-fusion wins **saturate** against one fixed budget rather than compounding.

---

## 2. dflash speculative backend

A **generic, config-driven** block-diffusion drafter and its runtime. Nothing is Muse-Glimmer-specific — the tap layers, block size and mask token all come from the drafter's `config.json`, so a future dflash drafter is a **drop-in bundle**, no code.

- `models/dflash.py` — `DFlashDrafter`. `encode(taps)` fuses multi-layer target residual-stream taps → `fc` → norm; `propose_block` seeds `[last-committed, MASK×(k-1)]`, runs **one** non-autoregressive forward, and reads `argmax(target_lm_head(hidden[1:]))` → the block of drafts. The forward is **bit-exact vs the `bstnxbt/dflash-mlx` reference** (used as the correctness oracle — max|Δ|=0 on encode and block hidden). An **incremental context cache** injects only newly-committed positions each round (O(new), not O(context)).
- `backends/dflash.py` — `DFlashRuntime`. One round = propose → **single target forward** over `[primary, *drafts]` → **target-prefix accept** walk → cache-normalize rollback → context extend. Decode is **token-exact vs greedy AR** (up to fp near-tie non-determinism, which is inherent to batched-vs-sequential forwards, not a bug).

---

## 3. Wiring dflash *through* the MTP route

dflash is not a parallel decode path — it runs through the **same MTP speculative-decoding route** that drives native MTP heads, and reuses that route's core machinery. `generate_mtpk` — MTPLX's MTP-**K** (block-speculation) generation entry point — is exactly where native MTP-head speculation lives; dflash dispatches through it. Concretely, it slots into the three mechanisms the MTP route already provides:

1. **The MTP route's by-`backend_id` dispatch.** `generate_mtpk` already branches on `rt.backend_id` (native MTP heads take the built-in draft/verify loop; the external Gemma-4 assistant takes its own). dflash adds one branch — `if rt.backend_id == "dflash": return generate_dflash(...)` — so it *is* an MTP-route backend, selected the same way, with `speculative_depth` flowing straight through as the drafter's block size. No new serve/API surface; `mtplx serve` speculates with it unchanged.

2. **The MTP route's verify/accept contract.** Verification uses **target-prefix** semantics — the same exactness contract the MTP route uses for its exact block-speculation lanes: the target forwards `[primary, *drafts]` once, samples/argmaxes each of the `K+1` verifier rows, accepts the leading run, takes the first miss as the correction (all-match → bonus), and rolls the target KV cache back to the accepted length. Because the drafter is argmax (no per-position `q`-distribution for a p/q residual accept), target-prefix is the correct MTP verify strategy here — and it makes decode **provably the greedy-AR sequence**.

3. **The MTP route's target-hidden tap — generalized.** Native MTP heads read the target's residual stream through a single-layer hidden tap (`target_pre_norm_hidden`). dflash needs the residual stream at *several* layers, so this PR **generalizes that same tap** into a zero-overhead multi-layer hook (`model._tap_layers` / `model._taps`): set a layer-index set and those residuals are captured each forward; leave it unset and it costs nothing. dflash feeds those taps into its `fc` encoder → the drafter's context.

**Load-time binding** mirrors the MTP route's Gemma-4 assistant-pair hook: `runtime.load()` resolves a **pair bundle** (`target/` + `drafter/` + `dflash_pair.json`) via `resolve_dflash_pair_paths` and returns a `DFlashRuntime` (`backend_id = "dflash"`), installing the target's arch shim first (the drafter borrows the target's `tok_embd`/`lm_head`).

So the only genuinely new plumbing is the multi-layer tap and the drafter itself — **dispatch, verify/accept, KV rollback, and depth all reuse the MTP route.** Drop a pair bundle and `generate_mtpk` speculates with it; the speculation is invisible to the client.

---

## 4. Benchmarks — throughput × block size (K), by workload

All three workloads are **Python code continuation** (short / predictable-repetitive / genuinely-diverse). All on the final backend (fold + context cache + batched accept), 1024-token context where noted, greedy, **token-exact vs greedy AR** (up to fp near-tie non-determinism — the diverse run's occasional divergences are last-bit argmax flips between batched-verify and sequential-AR, not accept-logic errors). `tok/s` is end-to-end decode; `accept%` = mean accepted drafts / (K−1). Muse-Glimmer q4 target + bf16 dflash drafter on M5 Max.

**Short code prompt** (~6 tokens, `def fibonacci(n):`) — AR baseline 28.1 tok/s
| K | tok/s | ×AR | mean_accept | accept% |
|---|-------|-----|-------------|---------|
| 2 | 29.6 | 1.05 | 0.82 | 82% |
| 4 | 42.0 | 1.49 | 1.93 | 64% |
| 6 | 43.3 | 1.54 | 2.95 | 59% |
| 8 | 40.5 | 1.44 | 3.18 | 45% |
| 12 | 31.0 | 1.10 | 3.41 | 31% |
| 16 | 30.9 | 1.10 | 3.41 | 23% |

**Repetitive code** (1024-token context, boilerplate / repeated patterns) — AR baseline 27.2 tok/s
| K | tok/s | ×AR | mean_accept | accept% |
|---|-------|-----|-------------|---------|
| 2 | 26.7 | 0.98 | 1.00 | 100% |
| 4 | 39.2 | 1.44 | 3.00 | 100% |
| 6 | 42.6 | 1.57 | 4.96 | 99% |
| 8 | 48.2 | 1.77 | 7.00 | 100% |
| 12 | 49.2 | 1.81 | 10.93 | 99% |
| 16 | 59.3 | 2.18 | 15.00 | 100% |

**Diverse code** (1024-token context, a real non-repeating Python module) — AR baseline 27.4 tok/s
| K | tok/s | ×AR | mean_accept | accept% |
|---|-------|-----|-------------|---------|
| 2 | 25.2 | 0.92 | 0.71 | 71% |
| 4 | 28.6 | **1.04** | 1.29 | 43% |
| 6 | 24.4 | 0.89 | 1.55 | 31% |
| 8 | 22.9 | 0.84 | 1.73 | 25% |
| 12 | 19.4 | 0.71 | 2.01 | 18% |
| 16 | 20.8 | 0.76 | 2.39 | 16% |

**Reading the matrix.** Speedup tracks acceptance, which tracks predictability:
- On **repetitive** content the drafter accepts nearly the whole block, so larger K wins → up to **~2.2× at K=16** (the drafter's trained block).
- On **diverse** code acceptance is low, so large K just wastes verify cost and the optimum shifts to **small K** — the takeaway is that **K should adapt to observed acceptance** (a follow-up), so diverse workloads never drop below AR.
- **Short prompts** peak in the mid-K range (little context to draft from).

The verify (target forward) is the floor (~81% of a K=16 round; it's bandwidth-efficient and its cost plateaus for M=12–32). The bf16 drafter forward is weight-bandwidth-bound and already reads each weight **once across all m=k+1 lanes** (measured flat vs K), so neither `mx.compile` (−3.4%) nor quantizing the drafter (net-loss — acceptance craters faster than the forward speeds up) helps.

---

## Status

Draft / WIP. The text tower and dflash backend are validated end-to-end through the serve path (`load` → `DFlashRuntime` → `generate_mtpk` → coherent, exact output). Follow-ups: **adaptive-K**, the **vision encoder** port (mlx-vlm), and a stronger/retrained drafter for diverse workloads.

https://claude.ai/code/session_01AzsPk4X2dkhrNRUDuCE1K1


davidtai:
### Exact 1024-token benchmark prompts (repetitive & diverse)

For reproducibility, the two 1024-context rows in the benchmark matrix use these exact Python prompts — each is a source string **encoded and truncated to the first 1024 tokens**, then fed to `/v1/completions` (greedy) for a 160-token continuation.

<details>
<summary><b>Repetitive code</b> — a 7-line block (`import os` … `quicksort` … `pivot = …`) repeated to fill 1024 tokens. Near-100% acceptance because the model is largely copying.</summary>

```python
import os
from collections import deque

def quicksort(arr):
    if len(arr) <= 1:
        return arr
    pivot = arr[len(arr) // 2]
import os
from collections import deque

def quicksort(arr):
    if len(arr) <= 1:
        return arr
    pivot = arr[len(arr) // 2]
import os
from collections import deque

def quicksort(arr):
    if len(arr) <= 1:
        return arr
    pivot = arr[len(arr) // 2]
import os
from collections import deque

def quicksort(arr):
    if len(arr) <= 1:
        return arr
    pivot = arr[len(arr) // 2]
import os
from collections import deque

def quicksort(arr):
    if len(arr) <= 1:
        return arr
    pivot = arr[len(arr) // 2]
import os
from collections import deque

def quicksort(arr):
    if len(arr) <= 1:
        return arr
    pivot = arr[len(arr) // 2]
import os
from collections import deque

def quicksort(arr):
    if len(arr) <= 1:
        return arr
    pivot = arr[len(arr) // 2]
import os
from collections import deque

def quicksort(arr):
    if len(arr) <= 1:
        return arr
    pivot = arr[len(arr) // 2]
import os
from collections import deque

def quicksort(arr):
    if len(arr) <= 1:
        return arr
    pivot = arr[len(arr) // 2]
import os
from collections import deque

def quicksort(arr):
    if len(arr) <= 1:
        return arr
    pivot = arr[len(arr) // 2]
import os
from collections import deque

def quicksort(arr):
    if len(arr) <= 1:
        return arr
    pivot = arr[len(arr) // 2]
import os
from collections import deque

def quicksort(arr):
    if len(arr) <= 1:
        return arr
    pivot = arr[len(arr) // 2]
import os
from collections import deque

def quicksort(arr):
    if len(arr) <= 1:
        return arr
    pivot = arr[len(arr) // 2]
import os
from collections import deque

def quicksort(arr):
    if len(arr) <= 1:
        return arr
    pivot = arr[len(arr) // 2]
import os
from collections import deque

def quicksort(arr):
    if len(arr) <= 1:
        return arr
    pivot = arr[len(arr) // 2]
import os
from collections import deque

def quicksort(arr):
    if len(arr) <= 1:
        return arr
    pivot = arr[len(arr) // 2]
import os
from collections import deque

def quicksort(arr):
    if len(arr) <= 1:
        return arr
    pivot = arr[len(arr) // 2]
import os
from collections import deque

def quicksort(arr):
    if len(arr) <= 1:
        return arr
    pivot = arr[len(arr) // 2]
import os
from collections import deque

def quicksort(arr):
    if len(arr) <= 1:
        return arr
    pivot = arr[len(arr) // 2]
import os
from collections import deque

def quicksort(arr):
    if len(arr) <= 1:
        return arr
    pivot = arr[len(arr) // 2]
import os
from collections import deque

def quicksort(arr):
    if len(arr) <= 1:
        return arr
    pivot = arr[len(arr) // 2]
import os
from collections import deque

def quicksort(arr):
    if len(arr) <= 1:
        return arr
    pivot = arr[len(arr) // 2]
import os
from collections import deque

def quicksort(arr):
    if len(arr) <= 1:
        return arr
    pivot = arr[len(arr) // 2]
import os
from collections import deque

def quicksort(arr):
    if len(arr) <= 1:
        return arr
    pivot = arr[len(arr) // 2]
import os
from collections import deque

def quicksort(arr):
    if len(arr) <= 1:
        return arr
    pivot = arr[len(arr) // 2]
import os
from collections import deque

def quicksort(arr):
    if len(arr) <= 1:
        return arr
    pivot = arr[len(arr) // 2]
import os
from collections import deque

def quicksort(arr):
    if len(arr) <= 1:
        return arr
    pivot = arr[len(arr) // 2]
import os
from collections import deque

def quicksort(arr):
    if len(arr) <= 1:
        return arr
```
</details>

<details>
<summary><b>Diverse code</b> — a real non-repeating module (decorator / dataclass / token-bucket / DP / async / Trie / RLE / matrix / sliding-window), truncated mid-function at 1024 tokens. This is the honest "real rejection rate" case.</summary>

```python
import asyncio
import hashlib
import logging
from dataclasses import dataclass, field
from datetime import datetime, timedelta
from functools import lru_cache, wraps

logger = logging.getLogger(__name__)


def retry(times=3, delay=0.5):
    """Decorator that retries a coroutine on exception with a fixed backoff."""
    def decorator(fn):
        @wraps(fn)
        async def wrapper(*args, **kwargs):
            last = None
            for attempt in range(times):
                try:
                    return await fn(*args, **kwargs)
                except Exception as exc:
                    last = exc
                    logger.warning("attempt %d failed: %s", attempt, exc)
                    await asyncio.sleep(delay * (2 ** attempt))
            raise last
        return wrapper
    return decorator


@dataclass
class Session:
    user_id: int
    token: str
    created: datetime = field(default_factory=datetime.utcnow)
    ttl: timedelta = timedelta(hours=12)

    @property
    def expired(self) -> bool:
        return datetime.utcnow() - self.created > self.ttl

    def fingerprint(self) -> str:
        raw = f"{self.user_id}:{self.token}:{self.created.isoformat()}"
        return hashlib.sha256(raw.encode()).hexdigest()[:16]


class RateLimiter:
    """Token-bucket limiter shared across async workers."""

    def __init__(self, rate: float, capacity: int):
        self.rate = rate
        self.capacity = capacity
        self.tokens = capacity
        self.updated = datetime.utcnow()
        self._lock = asyncio.Lock()

    async def acquire(self, amount: int = 1) -> bool:
        async with self._lock:
            now = datetime.utcnow()
            elapsed = (now - self.updated).total_seconds()
            self.tokens = min(self.capacity, self.tokens + elapsed * self.rate)
            self.updated = now
            if self.tokens >= amount:
                self.tokens -= amount
                return True
            return False


@lru_cache(maxsize=1024)
def levenshtein(a: str, b: str) -> int:
    if not a:
        return len(b)
    if not b:
        return len(a)
    prev = list(range(len(b) + 1))
    for i, ca in enumerate(a, 1):
        cur = [i]
        for j, cb in enumerate(b, 1):
            cost = 0 if ca == cb else 1
            cur.append(min(prev[j] + 1, cur[j - 1] + 1, prev[j - 1] + cost))
        prev = cur
    return prev[-1]


async def fetch_all(urls, limiter: RateLimiter):
    """Fetch a list of urls respecting the rate limiter, gathering results."""
    results = []
    for url in urls:
        while not await limiter.acquire():
            await asyncio.sleep(0.05)
        results.append(await retry()(_download)(url))
    return results


class TrieNode:
    __slots__ = ("children", "is_word")

    def __init__(self):
        self.children = {}
        self.is_word = False


class Trie:
    """Prefix tree supporting insert and prefix queries."""

    def __init__(self):
        self.root = TrieNode()

    def insert(self, word: str) -> None:
        node = self.root
        for ch in word:
            node = node.children.setdefault(ch, TrieNode())
        node.is_word = True

    def starts_with(self, prefix: str) -> bool:
        node = self.root
        for ch in prefix:
            if ch not in node.children:
                return False
            node = node.children[ch]
        return True


def run_length_encode(data: bytes) -> list:
    """Compress a byte string into (value, count) pairs."""
    if not data:
        return []
    out = []
    prev = data[0]
    count = 1
    for b in data[1:]:
        if b == prev and count < 255:
            count += 1
        else:
            out.append((prev, count))
            prev, count = b, 1
    out.append((prev, count))
    return out


def transpose(matrix):
    """Return the transpose of a rectangular matrix of numbers."""
    if not matrix:
        return []
    rows, cols = len(matrix), len(matrix[0])
    result = [[0] * rows for _ in range(cols)]
    for i in range(rows):
        for j in range(cols):
            result[j][i] = matrix[i][j]
    return result


def moving_average(values, window: int):
    """Yield the simple moving average over a sliding window."""
    if window <= 0:
        raise ValueError("window must be positive")
    total = 0.
```
</details>

youssofal (changes requested):
Great work on this. The optimization ledger in `docs/` is the standard this repo wants (in-window A/Bs, bit-exactness classification), and you were right 4/4 on "you can't reject a fusion by a naive concat". Two of the three parts of this PR can merge with fixes. The dflash backend as-is cannot, for one structural reason.

**The decode lane is greedy-only.** `generate_dflash` accepts `sampler` and ignores it. With the `runtime.load()` hook, any dflash bundle served with defaults answers every request greedily regardless of the requested temperature, with no error and no stats marker. For future reference: greedy-only implementations will not be merged into MTPLX. Models are served at their native sampling settings here; a lane that only works at temperature 0 changes the output distribution behind the user's back, and greedy benchmark receipts are demo numbers.

**The fix is smaller than the PR description assumes.** The description says the argmax drafter has no q-distribution for a p/q accept, but `block_logits`/`propose_block_cached` already produce full target-projected logits for every draft position. Softmax of those is q. Sample each position from q (the NAR mask conditioning makes positions independent, so the sequential p/q walk with residual correction stays exact at any temperature, the same property the native lane relies on), and verify with the p you already have in `vlog`. It is wiring, not research.

Asks to merge dflash (line refs in the full review notes):

1. Wire the p/q acceptance above; greedy only when the request itself asks for temperature 0.
2. Enforce `max_context`. It is declared (sink 64 + window 1024) and never read; `extend_context` grows unboundedly, and the drafter runs off its training distribution exactly on long generations.
3. Re-bench accept rate and tok/s at the target's native sampling settings; the manifest's `mean_accept: 2.1` is a greedy number.
4. Tests: temperature-0 token-exactness vs AR, rollback for both cache classes (the `_normalize`/`_rollback` cache surgery pokes private mlx-lm internals), pair-loader + registry, and a concurrency guard (the runtime has mutable `_taps` and no serial pin).
5. Smaller items: `_QKVG_FUSED` is keyed by `id(self)` and never evicted, so after an unload and reload a recycled address can serve the previous model's fused weights; the shim-install `except: pass` in `load_dflash_runtime` will resurface as a baffling mlx-lm error two calls later; inspect reports `mtp_supported: yes` for a lane that cannot sample yet.

Proposal: split the PR. Muse-Glimmer-AR (vendored classes, registry, shim, the bit-exact QKVG fusion after the id() fix, plus a fixture test for the llama.cpp parity the docstring claims) can go in on its own merits now. dflash follows once sampling is wired and re-benched. At that point it is a genuinely valuable lane: MTP-free models getting exact-sampling speculation is a real differentiator, which is exactly why it cannot ship greedy.

PR #267, opened 2026-08-16
feat(kv-cache): add q6 paged KV quantization

TESTYEE-09:
## Summary

Adds `q6` as an intermediate paged KV-cache quantization mode between the existing `q8` and `q4` modes.

The implementation uses the existing symmetric KV quantization architecture with `qmax = 31` and packs four 6-bit values into three bytes.

For a 256-dimensional KV head:

- q8: 256 packed bytes per K/V tensor
- q6: 192 packed bytes
- q4: 128 packed bytes

This is an internal MTPLX q6 representation and is not llama.cpp Q6_K or another external quantization format.

### Motivation

For large-context models on memory-constrained Apple Silicon, there is a fairly large memory gap between q8 and q4 KV storage. q6 provides an intermediate memory/precision option.

For Qwen3.8-27B with 256-dimensional KV heads, calculated K+V storage per token/KV-head is:

- q8: 520 bytes
- q6: 392 bytes
- q4: 264 bytes

At a completely filled 262,144-token context with 16 full-attention layers and 4 KV heads, calculated KV storage is approximately:

- q8: 8.0625 GiB
- q6: 6.0625 GiB
- q4: 4.0625 GiB

These are calculated cache-storage figures, not measured process RAM.

### Implementation

- adds `q6` mode normalization/config support
- adds vectorized q6 quantization/dequantization
- packs four 6-bit values into three bytes
- adds q6 paged-cache allocation support
- wires q6 through CLI/config/backend descriptors
- adds q6 to the existing Swift app settings surfaces
- keeps defaults unchanged
- leaves q8 and q4 behavior unchanged

The audit also found two existing assumptions that effectively treated any non-q4 width as q8, or any non-q8 width as q4. Those are now explicit bit-width mappings so q6 cannot silently use the wrong quantization grid.

## Verification

Rebased onto current `main` (MTPLX 2.9.2) and executed on Apple Silicon with Python 3.14.7 / MLX 0.32.0.

The branch now includes an end-to-end serving regression test. It sends a real `POST /v1/chat/completions` request through the AR serving/generation path, then verifies that the cache created and written by that request is a `VllmMetalPagedKVCache` with q6 enabled, 6-bit configuration, uint8 packed storage, 8→6 packed width, fp32 scales, and a nonzero offset. This specifically guards against the q6 toggle becoming dead plumbing.

Checks run:

- q6-focused suite: 80 passed
- required pytest trio from `CONTRIBUTING.md`: 275 passed
- Ruff checks on changed Python files: passed
- `python -m build`: passed (sdist and wheel)
- `scripts/fresh_venv_smoke.sh`: passed

## Benchmark Evidence

None. q6 quantization/dequantization has not been benchmarked, and no runtime performance claim is made in this PR.

The implementation detail most worth review is the cross-byte packing cost: unlike q8 (one code per byte) and q4 (two per byte), q6 codes straddle byte boundaries, so pack/unpack each cost several mask/shift operations plus a stack and reshape per group of four values. Dequantization runs on every attention call over the active window, so that is the most likely place for q6 to cost decode throughput relative to q8.

## Notes

- q6 is opt-in
- defaults are unchanged
- model weights are not modified
- q6 is lossy KV-cache quantization, and no quality claim is made relative to q8
- the existing plain paged-KV path dequantizes before MLX SDPA, so no Metal kernel consumes the packed bytes and none needed a 6-bit case
- q6 performance has not yet been benchmarked

Rebased onto current `main` at `e07224a` (2.9.2); the branch is two commits on top of it.


youssofal:
The two latent bit-width fixes in kv_quant.py (the bits property and the clip-bound map) are correct defensive hardening and I will land them with credit in the next minor. The q6 surface itself stays out for now: it collides with the 2.8.0 kv-quant rework and the PR is self-declared unexecuted. If you rebase the q6 lane onto 2.8.0 and run the checks in CONTRIBUTING.md, I will look again.

youssofal:
Interested. KV quantization has a history of dead toggles here, so this needs end to end proof it engages. Does your branch include a test that the served path really runs q6?

TESTYEE-09:
Yes — the branch now includes an end-to-end served-path proof. `test_served_chat_completion_really_uses_q6_paged_kv` sends a real `POST /v1/chat/completions` request through AR generation and asserts the cache actually created and written by that request is q6: `VllmMetalPagedKVCache`, q6/6-bit config, uint8 packed storage, 8→6 packed width, fp32 scales, and nonzero offset. I also rebased onto current `main` (2.9.2) and ran the q6-focused suite (80 passed), the CONTRIBUTING pytest trio (275 passed), Ruff, package build, and fresh-venv smoke; all passed. The PR body now records the executed evidence.

Issue #269, opened 2026-08-16
Session commit never succeeds (retokenized_prefix_not_extending_session) — 3–4% prefix reuse on two different models via the built-in app chat


TobiasvonCarnap:
# Session commit never succeeds (`retokenized_prefix_not_extending_session`) — 3–4% prefix reuse on two different models via the built-in app chat
 
Every turn is a full cold re-prefill. `session_commit.committed` is `false` on every postcommit I have inspected, on two different model families, with two different `unsafe_reason` values but the same `reason`. The bank still writes the snapshot (`stored: true`), so each turn spends GPU time and RAM producing an entry that cannot extend the session — in the worst case here, 662 s and 5.93 GB for a single discarded commit.
 
I am not using an external agent harness. This is the **built-in MTPLX app chat** against the app-launched daemon.
 
## Environment
 
- MTPLX `2.7.1` — `/health` reports `MTPLX_COMPILED_VERIFY_MAX_CONTEXT: 32768`, so 2.7.0 or later
- macOS 26.6.1, Apple M3 Max (`Mac15,8`), 64 GiB unified memory
- MLX 0.32.0, `stock_pypi_layout: true`, Python 3.14 runtime venv
- Daemon auto-launched by the app; I cannot edit the launch line directly
- Client: built-in app chat
Launch line as generated by the app:
 
```
mtplx serve --host 0.0.0.0 --port 8000 \
  --model /Users/<user>/.mtplx/models/Youssofal--Qwen3.8-27B-MTPLX-Optimized-Quality \
  --profile turbo --scheduler-mode serial --batching-preset solo \
  --prefill-chunk-tokens 2048 --experimental-mtp-cohorts \
  --ssd-session-cache on --ssd-session-cache-max-size auto \
  --ssd-session-cache-min-prefix-tokens 512 \
  --api-key <key> --fan-mode smart --app-launch-id <uuid> \
  --unsafe-force-unverified --yes --reasoning auto \
  --temperature 0.6 --top-p 0.95 --top-k 20 \
  --draft-temperature 0.6 --draft-top-p 0.95 --draft-top-k 20 \
  --reasoning-effort medium --no-stats-footer
```
 
## Model A — Qwen 3.8 27B Optimized-Quality, vendor path, `preserve`, turbo
 
`mtplx inspect` confirms the artifact is clean:
 
```
architecture: Qwen3_5ForConditionalGeneration
tier: verified          recognized: true       can_run: true
[...]

Issue #255, opened 2026-08-14
MTPLX performance tanks after 32k context window


anthonyalayo:
When running MTPLX with the recommended models (either Qwen3.6 or the newly released Qwen3.8), I notice the performance degrade incredibly fast as the context window increases. While your "benchmarks" typically highlight an extremely fast speed, in practice in an agentic session, I watch the marketed 40~ tps turn into 8~ tps.

I'm assuming MTPLX doesn't have proper support for prefix caching of hybrid GDN models? `vllm-metal` recently implemented it here: https://github.com/vllm-project/vllm-metal/pull/584

I tested this against MTPLX, and I confirmed that `vllm-metal` maintains it's tps even as my context window grows. Could you confirm/deny if MTPLX supports this? If it indeed doesn't, could you consider implementing it? Without this support, MTPLX honestly has no edge in real agentic workflows.

youssofal:
Confirmed: MTPLX runs exact prefix caching and MTP speculative decoding
together on hybrid GDN, and has since 2.0.0 on July 6, five weeks before
vllm-metal #584 merged on August 10.

Worth reading the description of the PR you linked:

> Hybrid prefix caching cannot be combined with speculative decoding because
> draft-state rollback across mamba state blocks is not implemented.

That is #584 describing its own limitation. Draft-state rollback across
recurrent state blocks is the hard part of this problem, and it is the part
MTPLX implements. The session bank checkpoints KV for the attention layers
plus the recurrent and conv state for the GDN layers at commit boundaries; a
sub-prefix restore lands on a stored recurrent boundary and re-prefills only
the tail. RAM tier first, SSD tier behind it. Exact, not approximate. It is
also why `--no-mtp` disables the prefix cache here (#246): the cache lives in
the speculative path. The inverse of #584's constraint, not the same gap.

You can verify it in one request without trusting me. Every response carries
`usage.prompt_tokens_details.cached_tokens` and `/health` exposes a
`session_bank` block. Send an identical prompt twice and read the second one.

Release QA this week, Qwen3.8-27B Optimized Speed, real coding sessions:

| history tokens | cached_tokens | source |
|---|---:|---|
| 17,796 | 17,702 | ram |
| 21,218 | 20,914 | ram |
| 20,385 | 17,434 | ram |

On #584 as the comparison: every metric in it is TTFT or a `max_tokens=1`
engine probe, on Qwen3.5-0.8B, with a 1,088-token shared prefix. Median TTFT
[...]

PR #270, opened 2026-08-17
fix(server): preserve native tools for parallel requests

shiftedx:
## Summary

- Keep PI and Hermes agent traffic on the existing hybrid tool contract by default.
- Use the tokenizer-native tool template only when the request explicitly declares `parallel_tool_calls: true`.
- Keep `parallel_tool_calls: false` and an omitted field on the stable hybrid session policy, avoiding turn-to-turn prompt-mode fingerprint changes and cold re-prefills.
- Preserve existing precedence for backend-required modes, request-level tool-mode overrides, and OpenCode compact mode.
- Apply the same resolved mode to chat generation, postcommit encoding, and Anthropic token counting.

This is a request-routing fix only. It does not inspect prompt wording or alter model weights, sampling, speculative decoding, or tool-call parsing. The review update is rebased onto current `main` and routes through the 2.8.1 `RequestPolicy` resolver, keeping chat generation, postcommit encoding, and Anthropic token counting on one policy path.

## Verification

- `uv run --frozen --extra dev ruff check mtplx/server/openai.py mtplx/server/request_policy.py tests/test_server_openai.py` — passed
- `MTPLX_CONFIG=<isolated-empty-path> uv run --frozen --extra dev pytest -q tests/test_server_openai.py tests/test_request_observability_golden.py` — 350 passed
- `MTPLX_CONFIG=<isolated-empty-path> uv run --frozen --extra dev pytest -q tests/test_canon_policy_smalls.py tests/test_api_benchmark_contracts.py tests/test_no_mlx_imports.py tests/test_public_cli.py tests/test_runtime_kpis.py` — 320 passed
- `uv run --frozen --extra dev python -m build` — sdist and wheel built
- `scripts/hygiene_scan.sh` — passed

Endpoint regression coverage includes PI and Hermes, explicit parallel enablement, explicit opt-out, an omitted field, request-level prompt-mode override precedence, and OpenCode compact-mode precedence.

## Scope of the review update

The original revision also inferred parallel intent from natural-language phrases. That routing has been removed following maintainer review because prompt-dependent mode changes would alter the session policy fingerprint across turns. The implementation now uses only the declarative OpenAI request field.


youssofal (changes requested):
The parallel_tool_calls: true path is right and I want that half. The natural-language routing cannot ship as is: tool_prompt_mode is part of the session policy fingerprint, so a mid-session turn that happens to contain "in parallel" flips the fingerprint, invalidates the session cache for that session, and forces a full cold re-prefill, then flips back next turn. That is the exact defect class 2.8.0 spent a release killing (#269, #255). The negation regex also misfires: "Do not touch X. Read Y and Z in parallel" reads as negated. Drop the regex, keep the declarative field, and this merges.

shiftedx:
Addressed in 4259d3a. I removed both natural-language regexes and all prompt-text routing. Native mode now requires an explicit `parallel_tool_calls: true`; `false` and an omitted field remain hybrid, so ordinary turns cannot change the session policy fingerprint based on wording. I also updated the tests and PR description to remove the superseded natural-language benchmark claim. The focused server suite is green (326 tests) and Ruff passes.

shiftedx:
Rebased the corrected implementation onto current main (2.8.1) as c55dba6 and resolved the RequestPolicy extraction conflict. The branch now has one focused commit: no prompt-text routing, only explicit `parallel_tool_calls: true`, with false/unset remaining hybrid. Current policy/server/golden tests are green (350 + 320), along with build, hygiene, and fresh-venv smoke.

youssofal:
The rework addresses the review, explicit parallel_tool_calls only with false and unset staying hybrid is the right contract, no wording based routing. Re-reviewing the rebased branch for the next release.

youssofal:
2.9.2 changed tool handling to passthrough by default. Can you check whether your case still reproduces on 2.9.2? If yes, rebase and I will review.

shiftedx:
Confirmed on 2.9.2 and current main: with the default passthrough setting, MTPLX forwards the declared tool schemas, but PI and Hermes requests with `parallel_tool_calls: true` use `tool_prompt_mode=hybrid`. The native-routing issue remains.

I rebased the PR onto current main at `41b3b0e`. The targeted regression fails for PI and Hermes without the patch; with it, all seven cases pass. False and unset remain hybrid, while explicit request overrides retain precedence.

I ran the focused server and observability suite: all 365 tests passed. The GitHub wheel, no-MLX smoke, and repository hygiene checks passed.

Issue #271, opened 2026-08-17
Draft depth: static D3 beats the tuned D2 on every content type I measured, and D3 is the cap


rinaldofesta:
Static D3 beats my tuned D2 on every content type I measured, and the curve is still rising when it hits the D3 cap.

Setup: MTPLX 2.7.1, MacBook Pro M5 Max 128 GB, `Qwen3.8-27B-MTPLX-Optimized-Quality`, onboarding tune landed on D2 (candidates AR/D1/D2/D3). Numbers are e2e tok/s over 384-token generations at 2k context, temp 0.2, one run per cell, using the per-request `depth` override:

| content | D1 | D2 (tuned) | D3 |
|---|---|---|---|
| python | 29.8 | 30.3 | 36.8 |
| typescript | 28.1 | 32.7 | 34.5 |
| english prose | 27.9 | 33.7 | 36.0 |

At the official sampler (temp 1.0, python): D2 31.6, D3 35.5. The tuner's single static answer underfits content variance: +21% on python for D3 vs the tuned D2, and `draft_control.maximum = 3` means I cannot see where the curve turns over.

Two asks, in increasing order of work:

1. **Let `tune` explore deeper static depths (D4/D5) and raise `depth_max` where the on-device measurement wins.** Same contract as today: if nothing beats the baseline, nothing is saved.
2. **Per-round adaptive depth as a tune-verified mode.** Prior art: the current frontier policy in the Qwen 3.8 MTP challenge (Layr-Labs/qwen-3.8-mtp-challenge, `Sources/MLXFastModel/Qwen36MTPBlockSession.swift`): a greedy marginal-depth rule that converts the head's top-2 logit margin into an acceptance estimate (sigmoid of margin/2), keeps drafting while the predicted value beats a marginal cost (`headStepCostRatio` 0.18 at the current frontier), allows depth 0..8 per round, and gates the deepest rounds behind a streak of fully accepted rounds. The policy operates entirely on the drafting side, so the Leviathan and Chen verification and the exact-output guarantee stay untouched. Their effective depth on coding content lands around 2.7 to 4.2, which matches what the static sweep hints at: the right depth varies per round, not per machine. `tune` could calibrate the cost ratio on-device and keep adaptive only if it beats the best static depth, under the same honesty rule.

[...]

PR #276, opened 2026-08-17
feat(bench): add --compare-static fixed-depth baselines to mtp-adaptive

rinaldofesta:
### What is the current behavior?

`mtp-adaptive` can compare an adaptive run against AR (`--compare-ar`), but not against fixed depths on the same suite. Answering "does the adaptive policy beat the best static depth?" requires separate invocations with `--min-depth/--start-depth/--max-depth` pinned to the same value and a hand-merge of the outputs. That workaround is what produced the numbers in #271.

### What is the new behavior?

- `--compare-static 2,3` runs the same prompt suite at each listed fixed depth after the adaptive pass, reusing the existing generation path (`AdaptiveDepthPolicy` with `min == start == max`, fresh policy per case, same seeds, sampler and verify settings).
- Results gain `static_rows` (per-depth per-prompt rows: tok_s, acceptance_by_depth, drafted/accepted) and `static_summaries` (per-depth mean_tok_s and draft totals) next to the existing `ar_rows`.
- At `temperature <= 0` each static row records `exact_match_vs_adaptive`, mirroring the existing `exact_match` semantics vs AR.
- Default `()`: output is byte-identical unless the flag is passed.

### Validation

- `python -m py_compile` on both files.
- Live run on an M5 Max 128 GB against Qwen3.8-27B-MTPLX-Optimized-Quality with `--policy expected_value --max-depth 3 --limit 2 --compare-static 2,3`: `compare_static: [2, 3]` recorded, `static_summaries` populated (D2 21.98 tok/s, D3 22.22, adaptive 21.02 on the 2-prompt subset), `exact_match_vs_adaptive: null` at temp 0.6 as intended.
- Not run: full benchmark suites, non-default verify strategies/cores.

Motivated by the discussion in #271: a same-suite static baseline makes the adaptive bench self-contained.

*Implemented with the help of a coding agent (Claude Code) and reviewed locally by the author.*

### DIFF
--- a/mtplx/benchmarks/runners/mtp_adaptive.py
+++ b/mtplx/benchmarks/runners/mtp_adaptive.py
@@ -58,6 +58,7 @@ def run_mtp_adaptive(
     limit: int | None = None,
     enable_thinking: bool | None = None,
     compare_ar: bool = False,
+    compare_static: tuple[int, ...] = (),
     base_hidden_variant: str | None = None,
     mtp_hidden_variant: str = "post_norm",
     concat_order: str | None = None,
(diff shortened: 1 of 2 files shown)


youssofal:
Useful knob, fixed depth baselines belong in the harness. Reviewing.

rinaldofesta:
I rebased the single patch onto current `main` after the two base force-pushes: head is `2537527`, mergeable again. `uv run --frozen --extra dev pytest -q tests/test_adaptive.py tests/test_greedy_confidence_runner_telemetry.py` passes 13/13; `ruff check` and `py_compile` pass. I also removed the old AI co-author trailer because current `main` rejects AI attribution in commit history; not rerun: the live model benchmark.

youssofal:
Landed in 2.11.2 under your name as 25375272, merged as 2eeed4cd. Thank you.

PR #281, opened 2026-08-17
fix(forge): honor quantize:false module overrides

shiftedx:
## Summary

Forge's mixed-precision converter currently turns every matching `module_overrides` entry into a quantization-parameter dictionary. An entry such as:

```json
{"suffix": "linear_attn.in_proj_a", "quantize": false}
```

therefore ignores the exclusion and silently falls through to the default 8-bit override values.

This change makes `build_predicate` return the literal `False` value expected by `mlx_lm.utils.quantize_model` when a matching override explicitly sets `"quantize": false`. The matched module stays at source precision; ordinary mixed-precision overrides, layer filtering, first-match precedence, and body fallback behavior are unchanged.

The converter docstring now records the exclusion contract, and the regression test verifies both the excluded path and an unmatched body path.

## Why this matters

Hybrid recipes use source-precision exclusions for numerically sensitive projections while quantizing the rest of the trunk. Before this fix, those recipes completed without an error but produced q8 tensors for the modules that were intended to remain BF16, changing both model size and the requested quality/performance tradeoff.

## Verification

- `uv run --frozen --extra dev ruff check mtplx/commands/forge_mixed_convert.py tests/test_forge_mixed_convert.py` — passed
- `uv run --frozen --extra dev pytest -q tests/test_forge_mixed_convert.py` — 10 passed
- `MTPLX_CONFIG=<isolated-empty-path> uv run --frozen --extra dev pytest -q tests/test_no_mlx_imports.py tests/test_public_cli.py tests/test_runtime_kpis.py` — 268 passed
- `uv run --frozen --extra dev python -m build` — sdist and wheel built
- `scripts/hygiene_scan.sh` — passed

A real Qwen 3.8 27B conversion was also structurally inspected: all 96 requested recurrent-projection exclusions remained BF16, none gained quantization scale/bias tensors or quantization-config entries, and the other 234 mixed-precision overrides were still applied.

### DIFF
--- a/mtplx/commands/forge_mixed_convert.py
+++ b/mtplx/commands/forge_mixed_convert.py
@@ -10,7 +10,9 @@ params (``True`` from the predicate).
 
 The predicate is called by ``mlx_lm.utils.quantize_model`` with
 ``(path, module)``; returning a dict routes those params to ``to_quantized``
-and records the per-module entry in ``config["quantization"]``.
+and records the per-module entry in ``config["quantization"]``. An override
+with ``"quantize": false`` returns ``False`` so sensitive modules remain in
+their source precision.
 """
 
 from __future__ import annotations
@@ -18,7 +20,6 @@ from __future__ import annotations
 import argparse
 import json
 import re
-import sys
 from typing import Any, Callable
 
 _LAYER_INDEX_RE = re.compile(r"\.layers\.(\d+)\.")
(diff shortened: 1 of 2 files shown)


youssofal:
Confirmed bug. It pairs with the norm convention fix from #301 that lands in the next release, reviewing this on top of that so the forge fixes go together.

youssofal:
Merged into 2.9.2 as 0833c7bc with your authorship, credited in the release notes. Thanks.

PR #288, opened 2026-08-18
perf(mtp): combine greedy token and confidence reads in opt-in lanes

ArthurOstapenko:
## Summary

MTPLX 2.9.2 can keep the ordinary greedy draft chain on device. This patch
does not engage on that default lane. It targets only opt-in lanes that still
need host-visible confidence metrics: `draft_margin_threshold` and adaptive
policies that declare `wants_draft_metrics`.

For an eligible regular greedy selector, the patch queues argmax and the FP32
top-eight read before one synchronization, then reuses the selected token after
the shared materialization. Adaptive-width routing, target-prefix routing,
correction caches, adapter ensembles, reranking, and stochastic draft sampling
retain their existing readers. No default changes.

Two counters expose the engagement under `stats.draft_core`:

- `greedy_confidence_sync_calls` counts joint token/metric reads.
- `greedy_confidence_token_reuses` counts reads whose prepared token was
  consumed. A margin gate can read the metrics and stop before token selection,
  so the two counters are intentionally distinct.

The depth-sweep, adaptive-policy, and margin-grid benchmark artifacts preserve
the counters both per row and in their aggregate summaries.

The tests cover first-index tie handling, one-hot proposals, confidence values,
selector eligibility, ExpectedValue engagement and token reuse, both outcomes
of the margin gate, zero engagement on the default greedy lane, and counter
propagation through all three benchmark reports.

## Verification

- PR-focused confidence/runner suite: 38 passed.
- Required `CONTRIBUTING.md` smoke trio: 273 passed.
- Full Python suite: 4,376 passed, 26 skipped, 0 failed.
- Ruff on every changed Python file, `compileall`, and `git diff --check`:
  passed.
- Wheel and sdist build: passed; both artifacts passed `twine check`.
- `scripts/fresh_venv_smoke.sh`: passed.

## Benchmark Evidence

Measured on 2026-08-28 after rebasing onto current `main` at `e07224a`
(v2.9.2 plus its follow-up fixes). Prepared local head: `5551ee0`.

- Hardware: Apple M5 Max, 128 GB unified memory.
- Software: macOS 26.5.2, Python 3.13.13, MLX 0.32.0.
- Model: Qwen3.8-27B MTPLX Optimized Quality; affine Q8, group size 64,
  with one BF16 MTP layer.
- Measured lane: ExpectedValue D3, batched stock verification, persistent draft
  cache, cycle MTP history.
- Sampler: temperature 0, top-p 1, top-k 0; 64 generated tokens per run.

This benchmark measures the adaptive confidence lane. The 2.9.2 default greedy
chain is intentionally ineligible for this patch and is covered by a zero-counter
test rather than included in the performance claim. The margin lane is covered
for engagement semantics but was not used for the throughput headline.

The comparison used one loaded runtime and three prompt classes. Each arm had
one warmup per prompt, followed by 10 measured pairs per prompt. A/B order
alternated by block, prompt order rotated by block, and external wall time was
bracketed by MLX synchronization. The control forced separate token and metric
reads while leaving the model state and remaining decode path unchanged. All
66 runs produced identical token, policy, control, and work fingerprints across
the two arms.

- Paired geometric mean: **+0.559%** end-to-end throughput.
- Shared-block bootstrap 95% CI: **+0.401% to +0.704%**.
- Per-prompt paired deltas: code **+0.353%**, JSON/tool **+0.740%**, creative
  prose **+0.584%**.
- First and second halves: **+0.374%** and **+0.744%**.
- Control-first and candidate-first strata: **+0.588%** and **+0.530%**.
- Median draft-phase time fell by **7.22% to 8.33%**, depending on the prompt.
- Measured engagement: candidate **1,860 joint reads / 1,860 token reuses**;
  control **0 / 0**.

The post-benchmark follow-up only preserves those counters in benchmark JSON
summaries; the measured `mtplx/generation.py` hot path is byte-identical in the
prepared head.

### DIFF
--- a/mtplx/benchmarks/runners/mtp_adaptive.py
+++ b/mtplx/benchmarks/runners/mtp_adaptive.py
@@ -8,7 +8,11 @@ from dataclasses import asdict
 from pathlib import Path
 from typing import Any
 
-from mtplx.benchmarks.runners.mtp_depth_sweep import _rate_by_depth, _sum_lists
+from mtplx.benchmarks.runners.mtp_depth_sweep import (
+    _rate_by_depth,
+    _sum_draft_core,
+    _sum_lists,
+)
 from mtplx.benchmarks.schema import encode_prompt_case, load_prompt_suite
 from mtplx.benchmarks.validators.basic import (
     validate_json_text,
@@ -223,6 +227,7 @@ def run_mtp_adaptive(
                 "commit_time_s": out.stats.commit_time_s,
                 "capture_commit_time_s": out.stats.capture_commit_time_s,
                 "bonus_time_s": out.stats.bonus_time_s,
+                "draft_core": out.stats.draft_core,
                 "bonus_tokens": out.stats.bonus_tokens,
                 "correction_tokens": out.stats.correction_tokens,
                 "verify_calls": out.stats.verify_calls,
@@ -316,6 +321,7 @@ def run_mtp_adaptive(
             "commit_time_s": sum(row["commit_time_s"] for row in rows),
             "capture_commit_time_s": sum(row["capture_commit_time_s"] for row in rows),
             "bonus_time_s": sum(row["bonus_time_s"] for row in rows),
+            "draft_core": _sum_draft_core([row["draft_core"] for row in rows]),
             "bonus_tokens": sum(row["bonus_tokens"] for row in rows),
             "correction_tokens": sum(row["correction_tokens"] for row in rows),
             "verify_calls": sum(row["verify_calls"] for row in rows),
--- a/mtplx/benchmarks/runners/mtp_depth_grid.py
+++ b/mtplx/benchmarks/runners/mtp_depth_grid.py
@@ -8,6 +8,7 @@ from dataclasses import asdict
 from pathlib import Path
 from typing import Any
 
+from mtplx.benchmarks.runners.mtp_depth_sweep import _sum_draft_core
 from mtplx.benchmarks.schema import encode_prompt_case, load_prompt_suite
 from mtplx.benchmarks.validators.basic import (
     validate_json_text,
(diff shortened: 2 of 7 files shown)


youssofal:
Checked the engagement in source before merging this, and the picture changed: the default greedy path already reads token and top2 in one eval, the separate confidence read only fires under the margin and adaptive knobs. So the 0.65% applies to that opt in lane, not the default one. If you rescope the claim to the margin and adaptive lane I will take it, with a counter so the engagement is visible.

youssofal:
2.9.2 moved the greedy chain on device, which probably absorbs or conflicts with this. Can you rebase on 2.9.2 and remeasure? If the 0.65 percent survives, I will take it.

ArthurOstapenko:
Rebased onto current `main` at `e07224a` (v2.9.2 plus the follow-up fixes) and
rescoped as requested.

You were right about the default path: the 2.9.2 greedy device chain already
owns that case, so this patch now stays out of it. The joint read is eligible
only when the margin gate or an adaptive policy requests confidence metrics,
and only with the regular greedy selector.

I also added two `stats.draft_core` counters: one for joint confidence reads and
one for actual prepared-token reuse. The default greedy lane reports zero;
ExpectedValue and both margin outcomes are covered by tests. The dedicated
depth-sweep, adaptive-policy, and margin-grid reports retain them per row and in
their summaries.

Fresh order-balanced ExpectedValue D3 A/B on the rebased code:

- **+0.559%** E2E, bootstrap 95% CI **[+0.401%, +0.704%]**
- all three prompt classes, both temporal halves, and both order strata positive
- candidate engagement **1,860 reads / 1,860 reuses**; control **0 / 0**
- identical token, policy, control, and work fingerprints across both arms

Verification on the rebased branch: 4,376 passed, 26 skipped; build, `twine
check`, and fresh-venv smoke passed. The title and body now limit the claim to
the opt-in confidence lanes. Ready for another look.

youssofal:
Adapted and shipped in 2.10.0, credited in the notes. The margin-gate and adaptive-depth lanes read the greedy draft token and its confidence in one GPU sync instead of two, and an engagement counter proves the default lanes stay untouched. Closing since the adapted version is on main. Thanks for the careful paired receipt in the PR, it made the call easy.

PR #295, opened 2026-08-19
fix(smart-fan): stop a wedged request pinning fans at max

El-Patronum:
## The incident

A single MTPLX server ran for about 15 hours on an M-series Mac, serving
the 27-billion-parameter Qwen3.8 build in serial mode with `--depth 3
--profile turbo`. A local client opened 10 concurrent chat completions.
One executed; nine queued. The executing request then wedged:
`requests_completed` froze at 50 and `active_requests` froze at 10 across
many minutes of polling, while the GPU kept submitting Metal command
buffers (`mtplx-model-owner` inside `IOGPUCommandQueueSubmitCommandBuffers`,
then `mlx::core::eval_impl`).

Killing the client closed its sockets. The server still reported:

```
active_requests: 10   smart_fan_active_count: 20   fan_boost_active: true
foreground_active: 1  requests_completed: 50 (unmoving)
```

`thermalforge max` had pinned both fans at 5349 and 5777 RPM, and they
never returned to auto. Stopping the server restored them at once.

## Root cause

`ServerState._smart_fan_activity_probe` returns a bare `True` whenever
`has_foreground()` is true:

```python
if self.has_foreground():
    return True
```

`SmartFanController._do_probe_activity`, the stale-lease reconciler, drops
leaked leases only while the activity probe reports idle, and it resets
`_engine_idle_since = None` on every busy poll. A wedged request stays
registered forever, so it reads as legitimately busy and the reconciler
never reaches its lease-dropping branch. The safety net built for this
exact failure cannot see it: it catches leases held over an idle engine,
never over a wedged but busy one.

The repository already separates wedged from slow. `progress_heartbeat`
ticks inside `generation._eval` on every settled `mx.eval`, described
there as proof that the model owner is alive, and `_OwnerStallProbe`
already turns that into a frozen-for-N-seconds reading for the stream
watchdog. The fan probe never consulted it.

## What this changes

**Presence no longer counts as progress.** The fan probe consults a
dedicated `_OwnerStallProbe`. A registered request whose owner heartbeat
has frozen past `FOREGROUND_STALL_DEADLINE_S` (environment variable
`MTPLX_FOREGROUND_STALL_DEADLINE_S`, default 180 seconds) reports idle so
the existing restore path runs. The deadline sits below
`STREAM_STALL_DEADLINE_S` on purpose, because reporting idle only stops
claiming the fans and never cancels a request. The probe rearms on the
foreground zero-to-one edge alone, so a stale idle baseline cannot condemn
a fresh request and a queued arrival cannot forge liveness for a wedged
one.

**Heartbeat coverage now reaches the batch lanes.** `batched_decode.py`,
`a3b_mtp_batch.py` and the AR pump settled their evals raw, and
`record_batch_step` fires only when a decode step produced responses, so
long prefills ticked nothing and healthy width-8 prefill would have read
as a wedge. Settled evals on the owner thread now tick, and only on a
proven step, because an empty pump iteration must not forge progress.
Embedding and rerank evals stay raw deliberately: ticking from a request
thread would fake owner liveness and blind the stream watchdog.

**Admission now bounds the waiting line.** Serial mode runs one request at a
time by design, so concurrency is already 1, but `ModelScheduler` queues
into a bare `deque()` and `--max-active-requests` has no default, leaving
the waiting line unbounded with no backpressure.
`_GenerationAdmissionMiddleware` sheds past
`--max-inflight-generation-requests` (environment variable
`MTPLX_MAX_INFLIGHT_GENERATION_REQUESTS`, default 4) with 429 and a
`Retry-After` header. The default floors at scheduler capacity so
`mtp_batch` and `ar_batch` still seal full-width cohorts, serial stays at
exactly 4, and an explicit value wins verbatim, including `0` to disable.
It runs ahead of `_SmartFanArrivalMiddleware`, so a shed request never
ramps fans.

## Testing

`4209 passed, 21 skipped`, with one pre-existing unrelated failure
(`test_laguna_s_2_1_ar_route_skips_qwen_performance_hooks`, which needs
85.3 GiB of unified memory and fails identically on `main`). The contract
test `test_smart_fan_stale_lease_reconciler_never_fires_while_engine_busy`
still passes, so a long legitimate generation keeps its boost.

Every new test is mutation-verified. Each of these one-line mutations
makes a named test fail: removing the proven-step guard, reverting the
pump to an unconditional tick, changing the rearm guard to fire on every
request, deleting either admission exemption, swapping the two
`add_middleware` calls, and replacing the explicit `is not None` check
with a truthiness check in the child-argv builders.

## Notes for review

This bundles three separable concerns: the fan probe, batch heartbeat
coverage, and admission. I am happy to split them into separate pull
requests. They arrived together because the batch heartbeat work is what
makes the fan probe safe outside serial mode.

Admission changes default behaviour by introducing 429 responses. If you
would rather ship it opt-in, one line changes the default.

Review surfaced one issue that this branch deliberately leaves alone:
`_prepare_shared_prefix` submits the whole common prefix as a single
forward, bypassing the prefill chunking in `BatchGenerator`. It predates
this branch, and these commits only reduce its heartbeat blackout, adding
two ticks where there were none. Chunking it remains a real but independent
change.


youssofal:
The failure mode is real and holding fans at max off a wedged request is exactly the kind of thing that should never be possible. Reviewing this properly rather than fast, the thermal path is the one place I refuse to merge quickly.

youssofal:
Thank you, this is in 2.12.0. Your work is in main under your name, in c758dfee and 0b4b704e. A request now counts as fan activity only while the model makes progress, so a stuck request can no longer hold the fans at maximum. I added one follow-up that tests the stuck case through the real fan controller (a9bb6f33). The commits were applied directly, so GitHub will show this PR as closed rather than merged.

Issue #280, opened 2026-08-17
mtplx is unusable because it starts running something immediately that maxes GPU on start


taoeffect:
### mtplx doctor --json

Many problems. After upgrading `mtplx` from homebrew and running into (#279 ; see that also for `doctor` output), I had it download its model again to its own location as it wanted.

After it finished that it decided it wanted to do the performance tuning again so I let it do that.

After it finished it launched the chat and the dashboard, but the GPU usage became pinned to max as if it were processing stuff (the `python` process running mtplx was to blame):

<img width="330" height="297" alt="Image" src="https://github.com/user-attachments/assets/687a08af-d5ef-4ffa-a7f9-e825e14eb2f8" />

I noticed it was constantly accessing `~/.mtplx/session-bank/manifest.sqlite` too according to Activity Monitor.

When I did `^C` to quit it, it also spat this error out:

<details>

```
[mtplx] compiled-verify prewarm {"buckets": [], "skipped": ["no_paged_entries"], "already": [], "elapsed_s": 0.0, "complete": true}
[6/6] Extended warmup continues silently in the background
[6/6] Warmup complete in 1.7s (31.92 tok/s)

MTPLX is ready.
Chat UI: http://127.0.0.1:8000/
OpenAI API Base URL: http://127.0.0.1:8000/v1
Model: mtplx-qwen38-27b-optimized-speed
Reasoning history: preserve (policy auto)
API key: leave blank for localhost
Health check: http://127.0.0.1:8000/health
Keep this terminal open. Press Ctrl-C to stop MTPLX.
Opening chat UI in your browser...
Live Dashboard: http://127.0.0.1:8000/dashboard/
Opening live dashboard in your browser...
^CERROR:    Cancel 1 running task(s), timeout graceful shutdown exceeded
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/opt/homebrew/var/mtplx/venv-2.8.1/lib/python3.13/site-packages/anyio/streams/memory.py", line 118, in receive
    return self.receive_nowait()
           ~~~~~~~~~~~~~~~~~~~^^
  File "/opt/homebrew/var/mtplx/venv-2.8.1/lib/python3.13/site-packages/anyio/streams/memory.py", line 113, in receive_nowait
    raise WouldBlock
anyio.WouldBlock

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #300, opened 2026-08-20
Hold the background warmup ladder while a request is in flight

Blakeolson21:
Branch: `q6-lane/warmup-inflight-foreground-yield` (3 commits, base `main` @ 90d8c4b)
Files: `mtplx/server/openai.py`, `tests/test_background_warmup.py`

## What is wrong

`_BackgroundWarmup._foreground_quiet_for_s` decides whether a warm rung may be
admitted, and it reads only `state.last_request_at`. That field is stamped when a
request **completes** (five assignment sites, all on completion or cancellation
paths). A request that has arrived and is still generating therefore leaves it
untouched, and on a daemon that has not completed a request yet it is still
`0.0`, which the function reads as "no request has ever landed" and reports as
infinite quiet:

```python
def _foreground_quiet_for_s(self) -> float:
    last = float(getattr(self.state, "last_request_at", 0.0) or 0.0)
    if last <= 0.0:
        # No request has ever landed (fresh boot): warm immediately.
        return float("inf")
    return max(0.0, time.time() - last)
```

The comment says "landed". The field means "finished". Everything follows from
that gap.

## Why this window is the worst one

The failure is worst exactly where it is most visible. A control UI that
restarts the engine on a configuration change is typed into immediately
afterwards, so the very first request of a serve is both the one that has not
completed yet and the one most likely to be warmed over. Under the turbo
profile's eight-rung ladder (`512,1024,2048,2560,4096,8192,16384,32768`) a
single rung is tens of seconds of background prefill on a 27B, so the operator's
first request shares the GPU with a rung that the 90 second idle grace was
supposed to have held back.

Observed on a 27B (Qwen3.8-27B, 6-bit, M3 Max) from the server's own request
log, real non-warmup rows recorded while the ladder was walking:

```
{'prompt_tokens': 2622,  'completion_tokens': 28, 'decode_tok_s': 0.75}
{'prompt_tokens': 2622,  'completion_tokens': 0,  'decode_tok_s': 0.0}
{'prompt_tokens': 11459, 'completion_tokens': 0,  'decode_tok_s': 0.0}
```

The per-chunk `_ForegroundYield` abort still works and still bounds the damage
once a rung is running. It cannot help with a rung that should never have been
admitted.

## The change

Model work that is in flight or queued holds the ladder as its own admission
condition, whatever the grace is. Only when nothing is running does the
completion stamp decide.

Four details worth reviewing:

- The busy check (`_foreground_busy`) runs at **step admission**, in
  `_run_step_inner`, before the warming generation begins. At that moment
  `has_foreground()` counts only real requests, which is why this code can use
  it where the `_ForegroundYield` shim documents that it deliberately cannot
  (the shim runs inside the warming generation, which has incremented the
  counter itself via `_run_generation_dispatched`).
- The queued half goes through the same `_foreground_model_work_pending` helper
  the yield shim uses, so foreground work that is queued but not yet executing
  also holds the ladder.
- Busy is its own branch, not a zero folded into the grace comparison.
  `MTPLX_WARMUP_IDLE_GRACE_S=0` is how an operator says "do not wait between
  turns", and expressing busy as zero quiet would collapse the hold there
  (`0.0 < 0.0` is False) and hand a warm rung a request that is still
  generating.
- The completion stamp itself is now gated on the same
  `request_observability["warmup"]` filter `_dashboard_record_completion`
  already applies, so a warming generation no longer writes `last_request_at`
  or `requests_completed`. Before this, a rung's own completion made the next
  rung read quiet as roughly zero and park for the full 90 second grace, and
  `/health` reported `seconds_since_last_request` as if a user had just been
  served on an untouched daemon. `last_request_at` means "a user request
  completed"; warmup was writing it. The counter moves with the clock so the
  two `/health` fields cannot disagree.

`last_request_started_at` is deliberately **not** used as the anchor even though
`_smart_fan_activity_probe` pairs it with `last_request_at`: warming generations
go through `begin_foreground` too, so that field is set by the warmup itself and
using it would make every rung after the first defer forever.

## Verification

Four added to `tests/test_background_warmup.py`:

- `test_background_warmup_defers_while_a_request_is_in_flight` (fails before)
- `test_background_warmup_holds_a_live_request_even_at_zero_grace` (fails
  before)
- `test_background_warmup_defers_while_foreground_is_queued` (fails before)
- `test_background_warmup_still_warms_a_genuinely_idle_fresh_daemon` (passes
  before and after, so the guard is shown not to cost the case it exists for)

`make_state` gains the `foreground_active` counter and the `has_foreground`
accessor a real `ServerState` carries, and the shared `_deferral_probe`
helper replaces a verbatim copy of the same setup in the existing
foreground-recent test.

The warmup-stamp half of the change has no direct test: every test in this
file fakes `_run_generation` (the real one needs a model runtime), so no
harness can observe the stamp. The closest pinned evidence is the existing
`test_dashboard_record_completion_skips_warmup_rows`, which asserts the
identical warmup filter on the sibling call in the same function.

```
$ python -m pytest tests/test_background_warmup.py -q
................                                                         [100%]
16 passed

$ git checkout main -- mtplx/server/openai.py && python -m pytest tests/test_background_warmup.py -q
FAILED tests/test_background_warmup.py::test_background_warmup_defers_while_a_request_is_in_flight
FAILED tests/test_background_warmup.py::test_background_warmup_holds_a_live_request_even_at_zero_grace
FAILED tests/test_background_warmup.py::test_background_warmup_defers_while_foreground_is_queued
3 failed, 13 passed
```

## Residual, stated plainly

A request is only counted from `begin_foreground`, so the 100 to 200 ms of
handler Python between arrival and dispatch (tokenize, chat-template render) is
still a window in which a rung can be admitted. Closing that needs an arrival
stamp that warming cannot set, which is a larger change; the per-chunk yield
already bounds that case to one warming chunk, which is what it was designed
for. This change closes the unbounded case.

If a code path ever leaked `begin_foreground` without its matching
`end_foreground`, warming would stop rather than misfire. Every site pairs them
in a `finally`, so this is a degradation mode, not a new failure mode.

### DIFF
--- a/mtplx/server/openai.py
+++ b/mtplx/server/openai.py
@@ -20692,9 +20692,29 @@ class _BackgroundWarmup:
         return cls.IDLE_GRACE_S
 
     def _foreground_quiet_for_s(self) -> float:
-        last = float(getattr(self.state, "last_request_at", 0.0) or 0.0)
+        # last_request_at is stamped when a request COMPLETES, so a request
+        # that has arrived and is still generating leaves it untouched --
+        # 0.0 on a daemon that has not finished one yet, which read as
+        # "infinitely quiet" and admitted a warm rung against live traffic.
+        # That is the post-restart window an operator actually types into
+        # (a UI that restarts the engine on a config change is typed into
+        # immediately afterwards), so the very first request of a serve was
+        # the one most likely to be warmed over. Model work in flight or
+        # queued is zero quiet; only then does the completion stamp decide.
+        state = self.state
+        try:
+            if state.has_foreground():
+                return 0.0
+        except BaseException:
+            pass
+        try:
+            if _foreground_model_work_pending(state):
+                return 0.0
+        except BaseException:
+            pass
+        last = float(getattr(state, "last_request_at", 0.0) or 0.0)
         if last <= 0.0:
-            # No request has ever landed (fresh boot): warm immediately.
+            # Nothing has ever run and nothing is running: warm immediately.
             return float("inf")
         return max(0.0, time.time() - last)
 
--- a/mtplx/server/openai.py
+++ b/mtplx/server/openai.py
@@ -20462,7 +20462,14 @@ def _run_generation(
             else (getattr(out, "finish_reason", None) or "stop")
         )
         _record_request_metrics(state, dict(envelope))
-        state.last_request_at = time.time()
+        if not bool((request_observability or {}).get("warmup")):
+            # Warming generations are not user requests: the same filter the
+            # dashboard gauge applies. Stamping this clock from a warm rung
+            # made the ladder defer against its own output (rung 512 finishes,
+            # rung 2560 then waits out a full idle grace) and made /health
+            # report seconds_since_last_request as if a user had just been
+            # served on a daemon nobody has touched.
+            state.last_request_at = time.time()
         state.requests_completed += 1
         _dashboard_record_completion(state, envelope=envelope, stats=stats)
         last = {
--- a/mtplx/server/openai.py
+++ b/mtplx/server/openai.py
@@ -20468,9 +20468,12 @@ def _run_generation(
             # made the ladder defer against its own output (rung 512 finishes,
             # rung 2560 then waits out a full idle grace) and made /health
             # report seconds_since_last_request as if a user had just been
-            # served on a daemon nobody has touched.
+            # served on a daemon nobody has touched. The counter moves with
+            # the clock so the two /health fields cannot disagree, and so the
+            # max-idle watchdog does not read warming as user traffic and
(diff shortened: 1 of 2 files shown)


youssofal:
Right problem, #280 is the user facing version of it. Reviewing.

youssofal:
Adapted and shipped in 2.10.0, with credit in the release notes and changelog. The warm ladder yields to live and queued requests and never stamps the traffic clock, so the first message after starting the engine is not sharing the GPU with a warm-up rung. It also closed #280, where a user called the engine unusable because of exactly this. Closing since the adapted version is on main rather than this branch. Thank you, this was a real user-facing win.

PR #302, opened 2026-08-20
Make the 6-bit hexpack route at m == 4 opt-in: measured as a loss on a real 27B trunk

Blakeolson21:
Branch: `q6-lane/vk-qmm6-m4-route-gate` (7 commits, base `main` @ 90d8c4b)
Files: `mtplx/nax_verify.py`, `tests/test_nax_verify.py`,
`benchmarks/repro_vk_qmm6_m4_route.py`, `docs/turbo-verify.md`, `.gitignore`

## What is wrong

The 6-bit split-K hexpack routes in `nax_verify.py` carry their own provenance
in the comment above them:

> 6-bit routes (9B tier, 2026-07-07) ... Microbench on the 9B hot shapes:
> 1.1-2.4x vs stock

On a real 27B-class trunk with the whole model resident, the `m == 4` route
inverted in our measurement: 20 to 30 percent slower than stock
`mx.quantized_matmul` on `applegpu_g15s` (M3 Max). The `m5` and `m6` routes
still won 19 to 45 percent, and the 4-bit `m4` route won 17 percent. The
observed regression was specific to 6 bits at four rows.

## Production-trunk observation (the policy basis)

Whole-model weights-only sweep: the full Qwen3.8-27B linear stack (64 layers
plus `lm_head`, 19.37 GiB of 6-bit weights, group size 64, fp16 activations),
swept once per iteration so every layer's weights stream from DRAM the way a
real decode round streams them. Timing one weight tensor repeatedly in a loop
leaves that tensor hot and flatters whichever kernel has the better launch
geometry; this regime avoids that.

Provenance: Apple M3 Max (`applegpu_g15s`), 64 GB RAM, macOS 26.3, MLX
(Apple's array framework) 0.32.1, greedy weights-only timing (no sampler; the
sweep times the linear stack, not generation), fans on automatic, three
independent runs with rounds interleaved, measured 2026-08-19 against `main`
at 90d8c4b:

| bits | m | stock | hexpack | delta |
|---|---|---|---|---|
| 6 | 4 | 63.3-64.7 ms | 77.5-82.8 ms | **+20 to +30% (loss)** |
| 6 | 5 | 82.8-87.3 ms | 67.3-67.7 ms | -19 to -22% (win) |
| 6 | 6 | 131.0-134.5 ms | 74.0-86.2 ms | -35 to -45% (win) |
| 4 | 4 | 48.6-54.2 ms | 40.4-45.0 ms | -17% (win) |

The direction at `m=4`, 6 bits was the same in all three runs. The same run
measured the machine's roofline with a plain streaming reduce (375 to 379
GB/s): against the resulting bandwidth floor for 19.37 GiB of weights, stock
reached 85 to 87 percent at `m=4` while the hexpack path reached 67 to 71
percent, so stock was near what the memory system can deliver and the split-K
geometry was giving bandwidth up at four rows.

## The checked-in reproducer did not reproduce the loss

Honesty requires leading with this. The reproducer shipped in this PR
(`benchmarks/repro_vk_qmm6_m4_route.py`) re-measures every `(m, kernel)` cell
once per round in a Williams balanced order, because an earlier fixed-order
version of the benchmark produced self-inconsistent results on the same cell,
larger than the effect being measured. In its corrected balanced-order form,
run on the same machine (macOS 26.3.1, MLX 0.32.1, 2026-08-19, commit
f574e5d), the synthetic sweep showed the `m=4` 6-bit hexpack cell **winning**
by 15 to 33 percent, not losing.

Both measurements are recorded with full provenance in the reproducer's
header. The two regimes differ (production-trunk observation versus
synthetic standalone stack), and we do not currently have a mechanism that
reconciles them. The change therefore ships as a conservative dispatch gate
rather than a removal, and the reproducer is checked in precisely so other
hardware can contradict either number cheaply.

## The change

`vk_qmm6_m4_enabled()` gates the 6-bit `m == 4` route, default off, with
`MTPLX_VK_QMM6_M4=1` to restore it. Gated rather than deleted on purpose:
this was measured on one GPU family, the upstream 9B numbers were presumably
taken on another, and our own synthetic reproducer disagrees with our
production observation, so the route must stay reachable. Nothing else moves.
The `n >= 2048` floor, the prefill exclusion, the `m5`/`m6` 6-bit routes, and
the whole 4-bit and 8-bit branches are untouched. `docs/turbo-verify.md`
documents the routing contract, and `.gitignore` gains one editor-tooling
entry.

## Verification

`test_qlinear_patch_routes_6bit_verify_shapes` previously asserted that the
`m == 4` 6-bit route fires. That assertion encoded the behaviour this PR
changes, so it is replaced rather than deleted: the test now covers the full
split (m4 opt-in, m5/m6 on by default) and still covers the `n >= 2048` floor
and the prefill exclusion for the routes that remain on. Added alongside it:

- `test_vk_qmm6_m4_env_flag` for the flag parsing
- `test_4bit_m4_route_is_unaffected_by_the_6bit_gate`, which pins the claim
  that the observed regression is 6-bit specific
- `test_vk_qmm6_benchmark_order_balances_positions_and_carryover`, which pins
  the reproducer's balanced measurement order

## Scope of the claim

One GPU family (`applegpu_g15s`), one MLX version (0.32.1), one model shape
family (Qwen3.8-27B), and two measurement regimes that disagree with each
other, stated above. The flag makes acting on contradicting evidence cheap.

It is worth saying what this does not buy. On the profile where 6-bit models
actually run (`sustained`, which does not set `MTPLX_NAX_VERIFY`) these
routes are not installed at all, and at draft depth 2 the verify block is
`m == 3`, below the `4 <= m <= 6` window. The gate matters on `turbo` at
depth 3 and above. It is a correctness-of-dispatch change, not a headline
speedup.


youssofal:
Will verify the measurement on a 27B trunk here and take the opt in if it reproduces. Making a measured loss opt in is the right call either way.

PR #316, opened 2026-08-21
Opt-in load-time GDN projection fusion on the verify-kernel path

grzracz:
**Type:** opt-in perf feature, default off; bitwise exact inside a measured row window, falls back outside it

**Problem:** each GDN layer runs four projection launches over one activation (`in_proj_qkv` N=10240, `in_proj_z` 6144, `in_proj_b` 48, `in_proj_a` 48); the two N=48 shapes fill 12 threadgroups on a 38-core GPU, 0.49x stock efficiency. The existing `MTPLX_FUSE_GDN_PROJECTIONS` concatenates them lazily but keeps the originals alive, +2.5 GB resident and +38 ms, so it is not usable as a default.

**Evidence:** +0.80% tok/s, 2.36 GiB freed (peak RSS 18.62 to 17.36 GB), 192 member calls collapse to 48 matmuls. `max|dlogit| = 0.0` at M = 1, 4, 16, 64; serve-side L4 byte-identical.

**Fix:** fuse the group at load time, in place, into one `nn.QuantizedLinear` and drop the originals, so the last reference to the source weights disappears and MLX frees them. Members become zero-copy row views that fall back to the unfused computation outside the exact window. Enable with `MTPLX_FUSE_PROJ=gdn`; unset = no change.

## Details

The source attributes become `FusedProjectionMember` modules sharing a `_FusionHub` that runs the fused matmul once per distinct input array (identity-keyed) and hands each member its slice, so `gdn_capture._gdn_input_projections`, stock `GatedDeltaNet.__call__`, `attention_split.split_call` and stock attention all work unmodified. No call site is edited. Members hold `weight`/`scales`/`biases` as zero-copy row views (0 bytes) and defer to `nn.QuantizedLinear.__call__` over their own view outside the window. Decode is M=4 verify plus M=1 draft, so the window costs nothing; prefill keeps the original arithmetic and completion.

Keeping the fused module an `nn.QuantizedLinear` is the load-bearing design choice: the verify kernels are installed by patching `nn.QuantizedLinear.__call__`, and a module calling `mx.quantized_matmul` directly, as `packed_concats` does, drops off that path and regresses at M=4.

| arm | tok/s | verify ms/call |
|---|---:|---:|
| base (off) | 43.99 / 44.04 | 67.22 / 67.15 |
| `MTPLX_FUSE_PROJ=gdn` | 44.40 / 44.34 | 66.75 / 66.83 |

+0.35 tok/s (+0.80%), verify -0.40 ms/call; base repeats agree to 0.02 tok/s. Measured without the 8-bit shape table it is only +0.11 tok/s, so the two compose super-additively (43.78 to 44.33, +0.55, +1.26%): with `out_proj` on the fast verify lane the GDN block's serial chain is shorter, so removing three dispatches from it is worth more.

## Row ceiling

The default ceiling is a conservative 4, the window measured exact on the stock MLX lane. Operators can raise `MTPLX_FUSE_PROJ_MAX_ROWS` after measuring their own backend's exact window; at draft depth 4 a ceiling of 16 measured +1.1% (43.82 to 44.30), bitwise exact.

## Negative results

* `attn` and `mlp` are rejected, not unmeasured. Both fuse the MTP draft block, where at M=1 the extra Python (hub dispatch plus `mx.split`) costs draft/tok 1.847 to 1.891 ms x 3 depths = +0.13 ms/cycle, cancelling the verify gain: `all` 44.18 vs `gdn` 44.33. MLP gate+up measured -0.15 tok/s against a -2.8 ms/cycle prediction. They ship but stay off; reviewers may prefer to delete them.
* Dispatch-count arguments do not hold inside `mx.compile`. The pre-registered prediction of -1.5 to -4 ms/cycle was off by an order of magnitude: it priced 176 removed dispatches at an eager microbenchmark's 2.9 us each, but the verify forward runs inside `mx.compile`, which already batches launches into command buffers. The argument still applies to the eager draft chain.

## Protocol

M2 Max (38-core, 64 GB, macOS 26.2, mlx 0.32.0, mtplx 2.7.1 patched), 5 prompts x 400 tokens x 2 reps, n=10 per arm, ABBA-interleaved, same-window controls.
Model `Qwen3.8-27B-MTPLX-Optimized-Speed-FP16`, `--profile turbo`, depth 3.

## Validation

* Re-measure the row window on your hardware before enabling. It is a property of MLX kernel selection, not of the model; mis-set for a backend, the fusion silently changes numerics.
* In-process: load, forward, fuse in place, same forward, assert `max|dlogit| == 0.0` at M = 1, 4, and the intended ceiling; then a serve-side byte-identical greedy capture with `MTPLX_FUSE_PROJ=gdn` vs unset.
* Check the `[proj-fusion]` banner for 192 member calls to 48 matmuls and the intended `max_fused_rows`, and confirm peak RSS drops (~2.4 GiB here). If it does not, the originals were not dropped and the fusion is costing memory.

### DIFF
--- a/mtplx/proj_fusion.py
+++ b/mtplx/proj_fusion.py
@@ -108,7 +108,12 @@ class _FusionHub:
             self._key = None
             self._outs = None
             self._served = 0
-            _outs = tuple(mx.split(self.fused(x), self.split_points, axis=-1))
+            # Contiguous copies: a strided view can route downstream kernels
+            # onto a different reduction variant and flip a ULP.
+            _outs = tuple(
+                mx.contiguous(o)
+                for o in mx.split(self.fused(x), self.split_points, axis=-1)
+            )
             self._key = x
             self._outs = _outs
             _STATS["fused_dispatches"] += 1
@@ -360,6 +365,16 @@ def configure_fused_projections(model: Any | None = None) -> dict[str, Any]:
     reset_fused_projection_counters()
 
     if not groups or model is None:
+        return fused_projection_stats()
+
+    if str(os.environ.get("MTPLX_PACKED_PROJ_CONCATS", "")).strip().lower() in {
+        "1", "true", "yes", "on",
+    }:
+        # Both features fuse the same projection groups; installing both would
+        # double-fuse and confound any A/B.
+        _STATS["enabled"] = False
+        _STATS["skipped"] += 1
+        _STATS["skip_reasons"].append("refused: MTPLX_PACKED_PROJ_CONCATS is enabled")
         return fused_projection_stats()
 
     max_rows = _default_max_rows()
--- a/mtplx/proj_fusion.py
+++ b/mtplx/proj_fusion.py
@@ -1,0 +1,414 @@
+"""Load-time projection fusion for Qwen3.5 hybrid models.
+
+Concatenates projections that share one input along the output axis and replaces
+the source modules, so their weights are freed: the GDN ``in_proj`` group into one
+N=16480 matmul, attention q/k/v into N=14336, MLP gate/up into N=2*ffn. Affine
+quantization groups run along the input axis, so the concatenation needs no
+requantization. The fused module is an ``nn.QuantizedLinear``, so the verify-kernel
+patches on ``nn.QuantizedLinear.__call__`` still route it. Members fall back to the
+unfused computation outside the row window where fusion is bitwise exact.
+
+Default off. ``MTPLX_FUSE_PROJ`` selects families: ``gdn``, ``attn``, ``mlp``,
+``1``/``on``/``yes`` == ``gdn,attn``, ``all`` == ``gdn,attn,mlp``.
+``MTPLX_FUSE_PROJ_MAX_ROWS`` overrides the row window ceiling.
+"""
+
+from __future__ import annotations
+
+import os
+import sys
+from typing import Any
+
+import mlx.core as mx
+import mlx.nn as nn
+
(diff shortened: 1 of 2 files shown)


youssofal:
Reviewed and ported, off by default for now. Two changes made in the port that I would ask for here too: the hub's split outputs need mx.contiguous, a strided view can route downstream kernels onto a different reduction variant and flip a ULP, which breaks seeded sampled identity, and MTPLX_FUSE_PROJ needs mutual refusal with MTPLX_PACKED_PROJ_CONCATS since both fuse the same groups. In serve A/Bs on this hardware it did not win consistently and the 2k rung was negative, so it stays opt in until that is understood. The architecture is right though, staying on the QuantizedLinear surface composes with the verify lane instead of refusing like the concat path has to.

grzracz:
Both changes are in a follow-up commit. The hub now makes its split outputs contiguous before serving them, since a strided view can send downstream kernels to a different reduction variant and shift the result by a ULP. configure_fused_projections now refuses to install when MTPLX_PACKED_PROJ_CONCATS is enabled, and that refusal shows up in skip_reasons in the [proj-fusion] report. The check is in this module; packed_concats is unchanged here. On the serve A/Bs: we measured +0.80% with the verify-kernel patch claiming the fused module's shapes. On hardware where that path doesn't trigger, the gain may not reproduce, which could explain the inconsistency you saw.

youssofal:
Landed in 2.9.2 as the opt-in MTPLX_FUSE_PROJ with some fixes on top, credited in the notes. Closing as merged.

PR #317, opened 2026-08-21
Opt-in FR-Spec reduced-vocabulary draft LM head

grzracz:
**Type:** opt-in perf feature, default off; output distribution provably unchanged (speculative sampling is exact for any proposal `q`).
**Problem:** every draft step streams the full 248,320 x 5,120 quantized draft head, ~0.70 GB of weights + scales + biases, about 2.1 ms of pure bandwidth per step at 340 GB/s. That is roughly a third of the measured 6.1 ms draft step.
**Evidence:** +5.2% tok/s (36.9 to 38.8), draft 4.3 to 2.2 ms/tok, ~0.7 GB freed, L4 byte-identical greedy capture.
**Fix:** FR-Spec (arXiv:2502.14856): an exact row subset of the draft-only LM head over a frequency-ranked 64k subset. `full` mode is a drop-in, `compact` mode is optional. Enabled only by `MTPLX_DRAFT_VOCAB_IDS`; unset means no change.

## Details

Exact row subset, no re-quantization: `mx.quantize` packs along the last axis, so `nn.QuantizedLinear` stores weight `(out_features, in_features * bits // 32)` uint32 with `scales`/`biases` of `(out_features, in_features // group_size)`. Output feature `i` lives entirely in row `i` of all three arrays, so `weight[S], scales[S], biases[S]` is an exact row subset and logits on `S` are bit-identical to the full head's. Rejection sampling with the residual correction `normalize(max(p - q, 0))` is exact for any proposal `q` provided the same `q` that produced the draft token is used in the acceptance test, and the draft head feeds only the draft logits, never the verify/target logits, so `p` and the output distribution are untouched.

`full` (default) returns a full-vocabulary-shaped row with a very negative sentinel outside `S`, so top-k/top-p sampling, argmax, `SparseDistribution` vocab_size and the compiled device draft core keep working untouched. `MTPLX_DRAFT_VOCAB_KEEP_BASE=1` keeps the full head resident for A/B work; the default frees it (~0.7 GB). Nothing in 2.9.0 does this: no reference to a reduced or compact draft vocabulary anywhere in the tree, and `draft_lm_head.py` is byte-identical between 2.7.1 and 2.9.0 (`draft_sampling.resolve_draft_temperature()` is adjacent but different: draft temperature, not vocabulary).

`compact` returns an `|S|`-shaped row plus sampler shims that map the subset index back to the vocabulary id, and hard-disables the paths that read raw indices. On its own it is worth +0.10 tok/s (+0.23%), draft/tok 2.247 to 2.201 ms, `accepted_by_depth` unchanged; `full` mode alone carries the entire +5.2%.

`scripts/build_vocab_subset.py` counts tokenizer ids over a mixed corpus (English web/edu, English Wikipedia, math/reasoning, multi-language code, Chinese Wikipedia, chat-templated dialogue), unions them with an always-keep set (all 33 specials, all 320 control ids >= 248000, digits, punctuation, whitespace/indent runs, all 128 single-byte tokens: 6,014 ids), and emits sorted int32 `.npy` at 16k/32k/48k/64k. It has hard-coded corpus and model paths: it is a reference implementation of the method, not shippable code. Coverage is measured on a per-domain held-out 10% document split, so the numbers are not self-fulfilling:

| domain | \|S\|=16k | \|S\|=32k | \|S\|=48k | \|S\|=64k |
|---|---:|---:|---:|---:|
| web_en | 85.13% | 93.37% | 96.87% | 98.57% |
| wiki_en | 85.76% | 94.39% | 97.54% | 98.96% |
| math | 92.61% | 97.44% | 98.96% | 99.61% |
| code | 88.29% | 94.57% | 97.10% | 98.30% |
| zh | 75.64% | 86.91% | 91.86% | 94.98% |
| chat | 94.98% | 98.52% | 99.50% | 99.88% |
| pooled | 84.16% | 92.41% | 95.81% | 97.62% |

We shipped 64k, 26% of the vocabulary.

## Reviewer concerns

* Artifact and tokenizer binding. The `.npy` must match the model's tokenizer; a stale or mismatched subset degrades acceptance silently, since verify still corrects every token and only speed drops. No checksum or tokenizer-id binding is implemented.
* Multilingual coverage. 94.98% held-out coverage on Chinese at 64k vs 98.57% on English web, so acceptance on non-Latin-script traffic drops more. A shipped default should be larger, per-locale, or opt-in.
* `compact` silent failure. A consumer that bypasses the installed sampler shims would emit raw subset indices, which verify silently corrects at the cost of acceptance collapsing. `full` mode has no such path.
* Silent hook failure. The hook in `draft_lm_head.py` sits in a bare `except Exception` that never fails the load. Right for an experiment, probably wrong upstream: a misconfigured `MTPLX_DRAFT_VOCAB_IDS` should be loud.

## Protocol

M2 Max (38-core, 64 GB, macOS 26.2, mlx 0.32.0, mtplx 2.7.1 patched), 5 prompts x 400 tokens x 2 reps, n=10 per arm, ABBA-interleaved, same-window controls.
Arms are the full 248k head vs the 64k subset, greedy decoding; metric is single-stream tok/s, with draft time per token from the same runs.

## Validation

* L4 byte-identical greedy capture with the head on and off, plus `scripts/test_reduced_vocab.py`, which checks the row-subset identity, the sentinel behaviour and the compact shim mapping.
* `accepted_by_depth` before and after: expect within noise, not necessarily identical, since drafts may legitimately differ where the subset bites. A collapse there is the signal that the subset is dropping tokens the target wanted.
* Still worth running: a non-English arm, and a deliberate tokenizer-mismatch arm to see what the failure looks like.

### DIFF
--- a/mtplx/draft_lm_head.py
+++ b/mtplx/draft_lm_head.py
@@ -303,6 +303,15 @@ def _install_draft_lm_head(rt: Any, *, bits: int, group_size: int, mode: str) ->
                 group_size=group_size,
                 mode=mode,
             )
+            # opt-in reduced-vocab draft head, active only via MTPLX_DRAFT_VOCAB_IDS
+            try:
+                from .reduced_vocab_draft import (
+                    maybe_wrap_draft_head as _mtplx_ext_wrap,
+                )
+
+                draft_head = _mtplx_ext_wrap(text, draft_head)
+            except Exception:  # pragma: no cover - never fail the load
+                pass
             layer.shared_head_head = draft_head
             reports.append({"layer": idx, **report})
         text._mtplx_step_mtp_draft_shared_heads = {
@@ -345,5 +354,25 @@ def _install_draft_lm_head(rt: Any, *, bits: int, group_size: int, mode: str) ->
         )
     else:
         raise AttributeError("model has no lm_head and does not tie output projection to embeddings")
+    try:
+        from .reduced_vocab_draft import maybe_wrap_draft_head as _mtplx_ext_wrap
+
+        _mtplx_ext_head = _mtplx_ext_wrap(text, draft_head)
+    except Exception as _mtplx_ext_exc:  # pragma: no cover - never fail the load
+        import sys as _mtplx_ext_sys
+
+        print(
+            f"[reduced-vocab] hook failed: {_mtplx_ext_exc!r}",
+            file=_mtplx_ext_sys.stderr,
+            flush=True,
+        )
+        _mtplx_ext_head = draft_head
+    if _mtplx_ext_head is not draft_head:
+        draft_head = _mtplx_ext_head
+        if isinstance(report, dict):
+            report = dict(report)
+            report["reduced_vocab"] = dict(
+                getattr(text, "_mtplx_reduced_vocab", {}) or {}
+            )
     text._mtplx_draft_lm_head = draft_head
     return report
--- a/mtplx/reduced_vocab_draft.py
+++ b/mtplx/reduced_vocab_draft.py
@@ -1,0 +1,450 @@
+"""FR-Spec (arXiv:2502.14856) reduced-vocabulary draft LM head.
+
+Restricts the draft-only LM head to a frequency-ranked token subset S.
+Speculative sampling is exact for any proposal q and the draft head never
+feeds the verify/target logits, so the output distribution is unchanged.
+The subset is an exact row slice of weight/scales/biases with no
+re-quantization, so logits on S are bit-identical to the full head's.
+
+``MTPLX_DRAFT_VOCAB_IDS`` (path to an ``.npy`` integer array of ids) is what
+activates the feature.  ``MTPLX_DRAFT_VOCAB_MODE`` selects ``full``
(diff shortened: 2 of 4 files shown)


youssofal:
This one is bigger than a normal review cycle. A reduced vocab draft head needs a per pack vocab artifact and acceptance gates at temperature before it can ship, the house acceptance rule keeps exactness so the risk is speed only, but that still has to be measured properly. It is being treated as part of larger draft head work rather than a straight merge. Leaving open.

youssofal:
Closing with a straight explanation. 2.10.0 ships an FR-Spec style pruned draft-vocabulary prototype behind MTPLX_FRSPEC_DRAFT, default off, but it is a separate implementation we ported from our own experiment rig, not a merge of this branch: it swaps a side-stamped head only around the device draft core so ratio acceptance stays exact for any proposal, which was the property we could not get from the branch as written. Same idea, independent code, so it would be wrong to leave this open implying a pending merge. If the prototype graduates after the long-context revisit, the technique discussion in this PR is part of how it got there. Thank you for pushing on it.

PR #338, opened 2026-08-24
Dense batched MTP: continuous batching, up to n=32, and prefix caching

grantoverton:
Tested on Qwen3.8-27B-MTPLX-Optimized-Speed (4-bit) on a Mac Studio M3 Ultra 80c 256 GB, and on Qwen3.5-4B-MTPLX-Optimized-Speed (4-bit).

Aggregate throughput on the 27B, thinking medium, MTP depth 3:

| streams | temp 1.0 (top-p 0.95, top-k 20) | greedy (temp 0) |
|---|---|---|
| 1 | 57.85 tok/s | 79.97 tok/s |
| 2 | 70.36 tok/s | 80.09 tok/s |
| 4 | 83.12 tok/s | 95.18 tok/s |
| 8 | 142.75 tok/s | 166.56 tok/s |
| 16 | 148.30 tok/s | 183.14 tok/s |
| 24 | 161.14 tok/s | 198.88 tok/s |
| 32 | 164.03 tok/s | 217.80 tok/s |

MTP speculation stays on the whole way.

## The problem

The dense path decoded one request at a time. A second caller waited for the first to finish. There was no way to put several requests through a single model pass while keeping MTP speculation, so concurrency and speculation were an either/or choice: you could have a fast single stream, or you could serve several people slowly.

This adds the missing path. N requests decode together in one model pass, each keeping its own prompt length, sampling settings and seed, with speculation running for all of them at once.

## What this adds

**Continuous batching.** A request that arrives while a batch is already running joins that batch instead of waiting for it to finish. In plain terms, a caller no longer waits in line behind a group that already started. Width follows demand in steps of 1, up to n=32. The batch is never padded, so three requests run as three rows, not as eight rows with five doing nothing.

**Prefix caching.** When a conversation continues, the server reuses the part of the prompt it has already read instead of reading it again. Measured on the running server with multi-turn chat traffic on the 27B: **82.7 percent of prompt tokens were never re-read**, across 8 concurrent conversations of 4 turns each. Off by default.

| turn | prompt tokens | reused | reuse | requests that hit |
|---|---|---|---|---|
| 1 | 628 | 567 | 79.0% | 7 of 8 |
| 2 | 786 | 628 | 80.1% | 8 of 8 |
| 3 | 939 | 786 | 83.4% | 8 of 8 |
| 4 | 1090 | 939 | 86.1% | 8 of 8 |

Reuse rises with conversation length, which is the point: the longer the history, the more there is to skip. The first turn already hits 7 of 8 because a shared system preamble is reused across conversations, not only within one.

A restored turn keeps the conversation. Built the same final prompt two ways, once turn by turn so the cache restores and once with a nonce that makes every stored prefix un-matchable, the replies agree to 90.9 percent word overlap and the restored one correctly references the earlier turns. Bit-identity is not the bar and the contract says why: float reduction order changes with batch geometry, so a restored row differs from a cold one exactly as any batched row differs from a solo run.

Two reporting defects were fixed to get these numbers, and both were the same shape. The lane restored prefixes correctly and told nobody: `cached_tokens` and `cache_hit` were hardcoded to `0` and `false`, so the usage block, `/metrics`, the restore mode and a benchmark all agreed the feature did nothing while the driver logged hits. Reuse is recorded per cohort row and a response needs it per request, and the two were never connected. Fixing the initial-cohort path alone still under-reported, because under continuous batching most requests JOIN, and the joiner path was not recording either: 12 driver hits showed as 3 on the wire. Both paths now attribute per request, into the dict the usage block is actually built from.

**A correctness fix that was failing silently.** The usual way to admit a row into a running batch is to left-pad it to the batch's length. On hybrid trunks that quietly corrupts output, because GDN (gated delta net) recurrent state is a running summary rather than an addressable cache, and it cannot be rewound past the padding the way a KV cache can. Rows are now admitted at their own length and joined along the row axis. Nothing pads.

## Throughput

Two numbers matter and they pull in opposite directions. **Aggregate tok/s** is the whole server's output, which is what counts when serving many people at once. **Per-stream tok/s** is how fast one reply appears, which is what a single user feels. Adding width raises the first and lowers the second.

**tau (τ)** is the number of tokens each stream commits per decode cycle. Its ceiling is 4 at MTP depth 3. It is the number that shows whether speculation still works as width grows.

Measured with the bundled `long_code` suite, 223 token prompts, `--vary-suffix` so every stream is distinct, 512 tokens per stream, MTP depth 3, thinking medium, prefix caching off. KV is the cache memory that scales with width; model weights are separate and unchanged by n.

### Temperature 1.0 (top-p 0.95, top-k 20)

Median of **24 sampling seeds** per width, with the 95% confidence interval on the mean.

| n | agg tok/s | per-stream tok/s | tok/cycle (batch) | τ (per stream) | accept d1/d2/d3 | KV reserved | KV used est | 95% CI |
|---|---|---|---|---|---|---|---|---|
| 1 | 57.85 | 57.85 | 2.995 | 2.995 | 82.5 / 66.8 / 50.2 | 0.06 GB | 0.04 GB | 53.9-60.0 |
| 2 | 70.36 | 35.18 | 5.627 | 2.813 | 84.0 / 64.5 / 48.2 | 0.12 GB | 0.09 GB | 67.5-72.8 |
| 4 | 83.12 | 20.78 | 10.807 | 2.702 | 82.4 / 63.4 / 49.0 | 0.25 GB | 0.18 GB | 80.0-85.8 |
| 8 | 142.75 | 17.84 | 20.481 | 2.560 | 82.6 / 65.9 / 50.4 | 0.50 GB | 0.36 GB | 136.8-145.8 |
| 16 | 148.30 | 9.27 | 37.410 | 2.338 | 81.8 / 63.4 / 48.1 | 1.25 GB | 0.72 GB | 136.8-151.7 |
| 24 | 161.14 | 6.71 | 56.892 | 2.370 | 82.6 / 63.7 / 48.1 | 1.88 GB | 1.08 GB | 140.9-160.3 |
| 32 | 164.03 | 5.13 | 74.264 | 2.321 | 81.7 / 62.2 / 47.1 | 2.50 GB | 1.44 GB | 149.8-165.6 |

### Greedy (temp 0)

| n | agg tok/s | per-stream tok/s | tok/cycle (batch) | τ (per stream) | accept d1/d2/d3 | KV reserved | KV used est |
|---|---|---|---|---|---|---|---|
| 1 | 79.97 | 79.97 | 3.879 | 3.879 | 99.2 / 95.5 / 94.7 | 0.05 GB | 0.04 GB |
| 2 | 80.09 | 40.04 | 6.132 | 3.066 | 91.3 / 80.9 / 71.6 | 0.12 GB | 0.09 GB |
| 4 | 95.18 | 23.80 | 11.703 | 2.926 | 88.1 / 74.8 / 63.0 | 0.25 GB | 0.18 GB |
| 8 | 166.56 | 20.82 | 22.261 | 2.783 | 84.6 / 66.6 / 52.3 | 0.50 GB | 0.36 GB |
| 16 | 183.14 | 11.45 | 42.890 | 2.681 | 84.6 / 66.0 / 50.7 | 1.00 GB | 0.72 GB |
| 24 | 198.88 | 8.29 | 65.016 | 2.709 | 86.1 / 69.3 / 56.2 | 1.50 GB | 1.08 GB |
| 32 | 217.80 | 6.81 | 90.519 | 2.829 | 86.8 / 70.4 / 56.0 | 2.00 GB | 1.44 GB |

**On precision.** The greedy figures are deterministic: argmax consumes no randomness, verified by running seeds 0, 7 and 12345 and getting byte-identical output, and three timing reps agree to within 0.06 to 0.93 percent. The temperature 1.0 figures are the median of 24 sampling seeds per width; sampling variance is large at these token counts, with a coefficient of variation of 8 to 16 percent, so single-seed numbers are not meaningful and the confidence interval is given alongside each median.

Greedy rises monotonically and peaks at n=32. Temperature 1.0 also peaks at n=32 once averaged over seeds. No width was ever blocked by the memory bound on this machine.

## What speculation is worth

Plain autoregressive decode over the same prompts and the same grid, so the only difference is speculation.

| n | AR agg tok/s | MTP agg tok/s | MTP τ | MTP vs AR |
|---|---|---|---|---|
| 1 | 30.93 | 79.97 | 3.879 | **+158.5%** |
| 2 | 57.79 | 80.09 | 3.066 | **+38.6%** |
| 4 | 97.22 | 95.18 | 2.926 | **-2.1%** |
| 8 | 119.25 | 166.56 | 2.783 | **+39.7%** |
| 16 | 143.68 | 183.14 | 2.681 | **+27.5%** |
| 24 | 203.78 | 198.88 | 2.709 | **-2.4%** |
| 32 | 271.43 | 217.80 | 2.829 | **-19.8%** |

Both sides are greedy because the AR path is argmax-only and takes no temperature, so a matched comparison under sampling would need a sampler added to it first.

Speculation is worth the most where there is least batching to be had, and the advantage narrows as width does the work instead. The τ column is the mechanism: every row where MTP loses is a row where τ has fallen well below its ceiling of 4. At n=32 greedy τ is 2.829, about 71 percent of maximum, and MTP would need roughly 3.5 to match AR there. n=1 already reaches 3.879 on this same prompt, so a better-matched MTP head raises the whole column and moves the crossings to the right.

## Reproducing

```
python -m mtplx.benchmarks.runners.dense_batch_bench \
  --model <model> --batches 1,2,4,8,16,24,32 --depths 3 --max-new 512 \
  --capture-backend linear-gdn-from-conv-stream --head-history committed \
  --loop-mode pipelined --history-window 24576 --thinking medium \
  --vary-suffix --warmup-tokens 16 --prompt-suite long_code \
  --temperature 1.0 --top-p 0.95 --top-k 20
```

Add `--decode ar` for the autoregressive baseline, and `--session-bank` to run with prefix caching live.

## What is in each commit

1. `kernels` KV-shared two-pass SDPA and ragged batched decode, the attention support for many streams in one pass.
2. `dense` the batched MTP driver, the no-padding row admission fix, per-row sampling and seeds, and optional prefix reuse.
3. `serve` the serving lane and continuous batching, memory-bounded cohort growth, and the counters an operator needs.
4. `bench` the runners that reproduce the numbers above.
5. `docs` the caller-facing contract, the operator guide, and the measured results.

## Tests

The full suite passes apart from nine failures that are present on unmodified `main` as well (`test_graphbank_compiled_verify`, `test_laguna_model`, `test_onboarding`). This branch adds no new failures.


youssofal:
Reviewing. n=32 continuous batching plus prefix caching is a big surface, so it will take some time. One question before I start, what did you measure for single stream regression when batching is idle?

grantoverton:
Just remeasured: appears to be no regression, single stream through the batch driver actually looks faster. Temp 0 shows 52 tok/s on the solo path vs 63 tok/s through the batch driver at n=1, so about +22%. At temp 1 it's about +12% over six matched seeds, but noisier.

Pushed the full numbers as `docs/benchmarks/single-stream.md`

grantoverton:
One tiny follow-up commit, flagging it so it doesn't widen the surface you're reviewing: `53501e3` makes `continuous` settable via `MTPLX_DENSE_BATCH_CONTINUOUS`. It's 11 lines.

PR #340, opened 2026-08-24
perf(generation): materialize greedy target IDs inside the verify evaluation sync (+10% greedy decode)

LucaLow:
## What

Eliminates per-row host/device synchronization in greedy verification.

Today, when the sampler is pure greedy, the accept loop calls `int(mx.argmax(row).item())` once per verifier row. Each of those scalar reads is a full MLX graph materialization: the CPU blocks while the device drains, once per proposed token, every verification window.

With this change, the target argmax IDs for all needed verifier rows are built as a single lazy array and attached to the **existing** verify-output evaluation in `_eval_verify_outputs`, so they materialize in the synchronization the runtime was already paying for. The accept loop and the bonus-token path then read plain host ints.

## Why it's safe

- **Token-neutral by construction**: the batched IDs are the same per-row argmaxes over the same unmodified logits - only the evaluation boundary moves. On a matched 1,024-token screen (Qwen3.8-27B NVFP4 target, depth-3 MTP, capture_commit, M5 Pro) the acceptance vector was identical (`[205, 152, 102]`), all 1,024 tokens emitted, output text byte-identical to control.
- **Narrow engagement**: the batched path activates only when temperature ≤ 0, no presence/frequency penalties, no steering/guard overlay, and not in `target_prefix` verification. Anything else keeps the stock per-row logic untouched. Grammar-constrained decoding is unaffected: the constraint clamp operates on the committed window, not on the per-row argmax, in both paths.
- **Purely additive**: when the engagement conditions don't hold, no existing code path is removed or reordered.
- Tests cover per-row/batched argmax equivalence and that the auxiliary array materializes through `_eval_verify_outputs` in lazy-logits mode.

## Measured effect

Same-day matched A/B, one warm process, identical prompt and settings:

| | decode tok/s |
|---|---|
| per-row reads (before) | 30.245 |
| batched into verify sync (after) | **33.277** (+10.0%) |

Detailed timing attribution showed host accept time falling from roughly one second per 1,024 tokens to milliseconds; the rest of the gain is consistent with not stalling the device pipeline once per row. The win should generalize to any model/host, since it scales with sync latency × verifier rows rather than with anything model-specific.

Extracted from a larger local optimization effort on Qwen3.8-27B; this is the piece that is model-independent and quality-neutral, so it seemed worth contributing upstream.


youssofal:
I ran this in a gated A/B on an M5 Max. The improvement did not separate from boot to boot variance, which was 3.3 tok/s on its own, larger than the deltas between arms. 2.9.2 also moved the greedy chain on device, which changes this path. If you measure a win on top of 2.9.2 with interleaved arms, reopen with numbers.

PR #344, opened 2026-08-25
Do not read the early tool-call cancel as a foreign cancel

### LINKED ISSUE #343: Streaming turns die with "request cancelled via POST /v1/mtplx/cancel" when nothing cancelled them (early tool-call cancel races the 250 ms queue poll)
## Summary

Streaming `/v1/chat/completions` requests are killed mid-flight with

```
request cancelled via POST /v1/mtplx/cancel after 1510 streamed tokens
```

when **nothing ever called that endpoint**. The server cancels its own
generation after a complete tool call, and then, on the very next iteration of
the stream loop, mistakes the flag it just set for a foreign cancel.

My client is an OpenAI-compatible desktop agent. It never POSTs to
`/v1/mtplx/cancel/{id}` — the only callers of that endpoint are MTPLX's own
dashboard (`postCancel`) and `benchmarks/runners/aime.py`. Nobody clicked cancel
anywhere — this fires unattended, mid-turn. The turn dies and the user loses the
whole assistant message.

## Environment

- MTPLX 2.9.1 (verified the same code path is unchanged in **2.9.2**)
- Apple Silicon (M-series, 64 GB), macOS 26.3.1
- `--backend-id qwen3_next --generation-mode mtp --depth 3 --profile turbo`
  `--scheduler-mode serial --batching-preset latency --stream-interval 1`
- Agentic workload: 23 tools, `tool_choice: auto`, ~58k-token prompts,
  50+ tool calls per session

## It is not OOM and not a crash

- The server process was still alive and serving when I investigated — no
  restart, no crash, no MLX allocation failure.
- It emitted a well-formed SSE error frame + `[DONE]` and kept serving: the
  requests right after completed normally at 1813 / 1667 / 1200 tokens.
- `/metrics` for the affected window:

  | field | value |
  |---|---|
  | `peak_memory_bytes` | 48.4 GB |
  | `producer_gap_ms_p95` | 224 ms |
  | `producer_gap_ms_max` | 795 ms |
  | `producer_gaps_over_200ms` | 177 (single request) |

Those gaps are the whole story — see below.

## Root cause

All line numbers are from **2.9.2**, `mtplx/server/openai.py`.

1. `STREAM_TOOL_CALL_FINISH_GRACE_S = 0.05` (L377). 50 ms after a complete tool
   call is detected, the stream loop cancels its own generation (L29415-L29424):

   ```python
   if (
       pending_tool_cancel_started_s is not None
       and not generation_future.done()
       and now_s - pending_tool_cancel_started_s
       >= STREAM_TOOL_CALL_FINISH_GRACE_S
   ):
       early_tool_cancel_used = True
       pending_tool_cancel_started_s = None
       _cancel_stream_generation(cancel_event, generation_future)  # cancel_event.set()
       continue
   ```

2. `continue` goes straight back to `kind, item = await queue.get(0.25)` (L29363).

3. The generation thread only observes `cancel_event` inside `on_tokens`
   (L27772/L27783), i.e. **once per committed token batch**. So the latency of
   the `("cancelled", …)` acknowledgement is exactly the inter-batch gap.

4. If that gap exceeds the 250 ms poll, `queue.get` raises `Empty` and the first
   check in the `except Empty:` branch is (L29377):

   ```python
   if cancel_event.is_set() or client_disconnected_now:
   ```

   `cancel_event` is set — by step 1, by the server itself. The client is still
   connected, so the `not client_disconnected_now` arm fires and the stream ends
   with `RuntimeError("request cancelled via POST /v1/mtplx/cancel after N
   streamed tokens")`, blaming an endpoint nobody called.

When the worker happens to acknowledge within 250 ms, `kind == "cancelled"`
(L30305) is handled correctly by `if early_tool_cancel_used and
streamed_assistant_tool_calls:` (L30340), which emits the proper terminal
tool-call frame. That is why this looks random: **it is a race between the
cancel acknowledgement and the 250 ms queue poll.** On a machine whose p95
producer gap is 224 ms and whose max is 795 ms, the loop loses that race
constantly.

Worth noting: the sibling loop for legacy completions already guards against
exactly this, excluding its own internal cancel (L31099):

```python
if (cancel_event.is_set() and not stop_hit) or completion_client_disconnected:
```

The chat-completions loop is missing the equivalent exclusion.

## Proposed fix

Do not treat a self-inflicted early tool-call cancel as a foreign cancel, the
same way `stop_hit` is already excluded in the completions loop:

```diff
-                            if cancel_event.is_set() or client_disconnected_now:
+                            if (
+                                cancel_event.is_set() and not early_tool_cancel_used
+                            ) or client_disconnected_now:
```

With that, an early tool cancel keeps draining until the worker acknowledges,
`kind == "cancelled"` lands on the existing `early_tool_cancel_used and
streamed_assistant_tool_calls` path, and the client receives the terminal
tool-call frame it is owed. A genuine `POST /v1/mtplx/cancel` still ends the
stream (the worker acknowledges and the `cancelled` handler runs), and a real
client disconnect is still caught by `client_disconnected_now`.

Two smaller things that would help independent of the fix:

- The error text hardcodes `POST /v1/mtplx/cancel` for *any* cancel with a live
  transport. `_stream_cancelled_queue_item` already carries a reason string —
  surfacing it would have made this a one-minute diagnosis instead of an
  afternoon.
- 50 ms of grace against a decode loop whose p95 batch gap is 224 ms is very
  tight; the cancel is essentially guaranteed to land mid-batch.

Happy to send a PR with the one-line change if that shape looks right to you.

Tavernari:
Fixes #343.

## The bug

The streaming chat-completions loop kills healthy tool-calling turns with

```
request cancelled via POST /v1/mtplx/cancel after 1510 streamed tokens
```

although no client ever calls that endpoint. It is a race the server runs
against itself:

1. 50 ms after a complete tool call, the loop cancels its own generation
   (`STREAM_TOOL_CALL_FINISH_GRACE_S`, L29415-L29424) — `cancel_event.set()` —
   and `continue`s.
2. Back at `await queue.get(0.25)`.
3. The generation thread only observes `cancel_event` inside `on_tokens`
   (L27772/L27783), i.e. once per committed token batch, so the acknowledgement
   latency *is* the inter-batch gap.
4. When that gap outlives the 250 ms poll, `except Empty` fires and its first
   check is `if cancel_event.is_set() or client_disconnected_now:` — the event
   the loop itself set one iteration earlier. Client still connected, so it
   emits the error frame and returns.

When the worker does acknowledge inside 250 ms, `kind == "cancelled"` (L30305)
handles it correctly via `early_tool_cancel_used and
streamed_assistant_tool_calls` (L30340) and the client gets its terminal
tool-call frame. Hence the apparent randomness.

On the box where I hit this (`qwen3_next`, depth 3, ~58k-token prompts, 23
tools) `/metrics` reported `producer_gap_ms_p95: 224`, `producer_gap_ms_max:
795`, and `producer_gaps_over_200ms: 177` in a single request — the 250 ms poll
loses that race constantly.

## The change

One condition, mirroring the guard the completions loop already applies to its
own `stop_hit` cancel (L31100):

```diff
-if cancel_event.is_set() or client_disconnected_now:
+if (
+    cancel_event.is_set()
+    and not early_tool_cancel_used
+) or client_disconnected_now:
```

The loop now keeps draining until the worker acknowledges, landing on the
existing `kind == "cancelled"` path that emits the proper terminal frame.

Deliberately **not** modelled on the `stop_monitor.stopped` guard a few lines
above, which `continue`s past the stall watchdog and the heartbeat: if a worker
never acknowledges an early cancel, this version still breaches
`MTPLX_STREAM_STALL_DEADLINE_S` and fails with a truthful message instead of
hanging.

Behaviour preserved:

- a genuine `POST /v1/mtplx/cancel/{id}` still ends the stream — the worker
  acknowledges and the `cancelled` handler runs;
- a real client disconnect is still caught by `client_disconnected_now`;
- the stall watchdog still bounds the wait.

## Verification

```
pytest tests/test_no_mlx_imports.py tests/test_public_cli.py \
       tests/test_runtime_kpis.py tests/test_stream_stall_watchdog.py \
       tests/test_server_obs_stream_termination.py tests/test_stream_guard_env.py
```

281 passed, before and after. `ruff check mtplx/server/openai.py` reports the
same 289 pre-existing findings on both sides of the diff — the change adds none.

I did not add a regression test: the branch lives inside a ~3k-line async
generator with no seam to drive it from. If you would like one, I am happy to
extract the predicate into a module-level helper (in the spirit of
`_OwnerStallProbe`) and unit-test it — say the word and I will push it here.

Two side notes from #343, not addressed in this PR because they are yours to
decide:

- the error text hardcodes `POST /v1/mtplx/cancel` for *any* cancel with a live
  transport, while `_stream_cancelled_queue_item` already carries a reason
  string that would have made this a one-minute diagnosis;
- 50 ms of grace against a decode loop whose p95 batch gap is 224 ms means the
  cancel essentially always lands mid-batch.

### DIFF
--- a/mtplx/server/openai.py
+++ b/mtplx/server/openai.py
@@ -29374,7 +29374,26 @@ def create_app(state: ServerState) -> FastAPI:
                             client_disconnected_now = (
                                 await raw_request.is_disconnected()
                             )
-                            if cancel_event.is_set() or client_disconnected_now:
+                            if (
+                                cancel_event.is_set()
+                                and not early_tool_cancel_used
+                            ) or client_disconnected_now:
+                                # `and not early_tool_cancel_used` because the
+                                # early tool-call cancel below sets the very
+                                # same event and then `continue`s straight back
+                                # into this branch. The worker only observes
+                                # the event once per committed token batch
+                                # (`on_tokens`), so whenever that batch gap
+                                # outlives the 0.25s poll above, the loop read
+                                # its own cancel as a foreign one and killed a
+                                # healthy tool-calling turn. Excluding it keeps
+                                # draining until the worker acknowledges, which
+                                # lands on the `kind == "cancelled"` handler
+                                # and its `early_tool_cancel_used and
+                                # streamed_assistant_tool_calls` terminal
+                                # frame. The completions loop already guards
+                                # its own `stop_hit` cancel this way.
+                                #
                                 # Truthful cancel accounting (#F36): only a
                                 # genuinely dead transport is a client
                                 # disconnect; an explicit server-side cancel


youssofal:
Merged as 2e0b2e1f with your authorship. Your diagnosis was exact. I reproduced the failure deterministically first, a fake worker that pumps a complete tool call and then blocks 1.2 seconds before returning, and on the unfixed tree it produced the same phantom cancel frame from your report. With your guard the stream drains until the worker acknowledges and the turn ends with the proper tool_calls frame. The full server suite passes, and the hung worker case stays bounded because the stall watchdog sits right below the guard. The repro test is on main as e07224aa. This goes out in the next patch release.

PR #345, opened 2026-08-25
perf: int4 simdgroup-mma lanes (small-M vocab + prefill retile), optional native MPP path

Manojbhat09:
## What

New module `mtplx/kernels/int4_simd_mma.py` — exactness-gated INT4 matmul lanes that engage behind `MTPLX_INT4_MMA` (off by default), plus repro scripts. No existing files are touched. Rebased on 2.9.2 (bbc6742).

- **vocab lane** (M 8..16): BM8/16-padded tile, BN=32, BK=64, one 128-thread TG per 32 output columns. Stock steel dispatch collapses on wide-N shapes at M>=8 (44.5/28.4 GB/s at m=8/16 vs 98.6 GB/s at m=1); this lane keeps the weight stream at bandwidth.
- **prefill lane** (M >= 128, M%128==0): BM128/BN32/BK64, WM4/WN1, double-buffered weight tiles with next-tile dequant overlapped into current MMAs, one barrier per inner iteration, GROUP_M=8 swizzle.
- **native MPP lane** (opt-in `mpp`): feeds MLX-packed uint4 directly into `tensor_ops::matmul2d` as a `uint4b_format` tensor on G17-class GPUs (macOS >= 26.4), with per-group affine decomposition (`C += s_g*(A_g*W_g) + b_g*sum(A_g)`). One-time eager compile probe caches availability (MLX compiles custom kernels lazily); graceful fallback elsewhere.
- `install_int4_mma_qlinear_patch()` mirrors `nax_verify.install_nax_qlinear_patch`: idempotent, `lane_disabled` kill switches, dispatch counters + snapshot.

## Results (base M4 / 16 GB dev box)

| lane | check | result |
|---|---|---|
| prefill | exactness | **bit-exact** on all trunk shapes incl. real W4/G64 weights, m 128..8192 |
| prefill | speed vs stock | flat **1.06–1.09x** M=128→8192 |
| vocab | exactness | bf16-rounding level (= stock's own fp32 error) |
| vocab | speed | **1.47x @ m=8**, 1.25x @ m=16 (real W4/G64); 1.32x/1.22x @ N=248320 synthetic |

## Reproduce

Requirements: Apple Silicon Mac, Python ≥3.11, `mlx>=0.32,<0.33`, `mlx-lm>=0.31,<0.32` (repo pins). Kernels run on any Apple GPU; the native MPP lane additionally needs G17-class + macOS ≥26.4 and is opt-in.

```bash
# kernels: exactness gates + speedups (~1 min; add --real-weights for
# production tensors — auto-downloads the 239 MB drafter checkpoint)
PYTHONPATH=. python benchmarks/repro_int4_mma_lanes.py [--real-weights]

# long-context serving of a 27B 2-bit model on 16 GB:
# >=10 tok/s decode from 4K to 128K, <=12 GB peak
PYTHONPATH=. python benchmarks/repro_long_ctx_16gb.py     --model auto --download-model          # or --model <local kexjos dir>
```

Expected outputs are documented in each script's docstring; full receipts (incl. the MLX grid-semantics gotcha, needle-recall controls, negative results) on branch `notes/m4-16gb-qwen38-worklog` → `docs/perf/receipts/m4-base-16gb-worklog-qwen38.md`.

## Validation

Random-weight gates across gs{32,64} x M{8..16} (vocab) and all four real trunk shapes at M∈{128,512} plus m∈{128,256} on real drafter weights (prefill, bit-exact); installer counters verified; uninstall restores stock; repro scripts executed green before push.


youssofal:
This is serious work and the exactness gates are exactly how kernel PRs should arrive. I will run your repro on an M5 Max and come back with numbers. The native MPP lane is the part I am most interested in.

Manojbhat09:
Nevermind, I worked on it this weekend but it seems like it can't handle more than 1024 ctx after which decode is garbage but if you make it work it would be a great addition!

PR #336, opened 2026-08-24
feat(runtime): add native semantic, replay, and adaptive systems

PhilipJohnBasile:
## Maintainer action required

The reviewer-facing README pass is complete on exact head:

`9e067af5f65e26f4ff3d056f55c5dd8b80428e8a`

Because this external-fork PR modifies workflow files, GitHub created new exact-head runs with no jobs and marked them `action_required`. Please approve:

- [build](https://github.com/youssofal/MTPLX/actions/runs/32744840351)
- [ci](https://github.com/youssofal/MTPLX/actions/runs/32744840252)
- [hygiene](https://github.com/youssofal/MTPLX/actions/runs/32744840224)
- [native-adaptive matrix](https://github.com/youssofal/MTPLX/actions/runs/32744840146)

The preceding code-and-matrix head, `53ba0cc1bc96b135da32fba25615dfff96b81bf6`, passed all four authoritative upstream workflows. The only change after that green head is `README.md`. Maintainer edits remain enabled, and `@youssofal` is the requested reviewer.

## Reviewer-facing README update

The top-level [README before-and-after section](https://github.com/PhilipJohnBasile/MTPLX/blob/9e067af5f65e26f4ff3d056f55c5dd8b80428e8a/README.md#before-and-after) now documents the complete contribution in product terms rather than requiring the reviewer to reconstruct it from source files. It includes:

- a seven-row **before / after** comparison covering SessionBank reuse, MoE working sets, memory governance, production diagnosis, candidate evaluation, observability/policy, and operations
- every new subsystem: semantic anchors, expert locality, safe-point memory governance, expert warm-set control, unified-memory coordination, privacy-default capture, deterministic replay, trace parity, OTLP/HTTP export, lifecycle policy hooks, and capture-to-replay orchestration
- the actual environment-variable names and a representative opt-in configuration
- `/v1/mtplx/systems`, the Systems dashboard, the read-only Native command surface, and the relevant CLI commands
- the privacy, safe-point, router, KV-resize, fail-open, and non-automatic-promotion boundaries
- a collapsible maintainer review order and links to the operator docs and validation receipt

## What this adds

This is a vendor-neutral MTPLX implementation. It does **not** import, embed, launch, or require FreeToken, Future AGI, or another external runtime.

### Semantic runtime intelligence

- exact SessionBank anchors from verified complete-message chat-template prefixes
- safe-point SessionBank memory governance under the model lock
- opt-in MoE expert-locality telemetry without router mutation

### Deterministic replay and evaluation

[...]

PR #350, opened 2026-08-26
feat(cache): add exact semantic session anchors

PhilipJohnBasile:
## Scope

Adds opt-in semantic SessionBank anchors behind `MTPLX_SEMANTIC_ANCHORS=1`: complete-message prefix planning with strict token-prefix validation, bounded anchor selection and prefill-boundary propagation across cold, exact and near-prefix paths. The feature is extracted from #336.

## Completed real-model campaign: acceptance fails

The [completed report and all receipts](https://github.com/PhilipJohnBasile/MTPLX/blob/6d0db7f43a879646f54c107fd2d9afd3e86c7b64/docs/validation/semantic-anchor-20260907/README.md) are attached in documentation-only commit `6d0db7f43a879646f54c107fd2d9afd3e86c7b64`. The measured runtime and replay harness remain at `dfe28bf94fb62d35ba305d3b764c044c880f0905`; neither was changed between arms or to accommodate a failure.

**Three fresh OFF/ON pairs per workload are complete: twelve isolated server arms and sixty measured requests.** The workloads cover a 4K conversation and a 126,982–127,166-token long conversation, each with actual model-generated tool calls, deterministic tool replies, a plain append and repeated tool use. Frozen complete request bodies are replayed without feeding new outputs into later requests.

- Output-parity failures: **long pair 1, long pair 2, long pair 3**. Full output hashes, completion counts and finish reasons are retained. The initial long plain append returns `Marker: amber-17` with OFF versus `amber-17` with ON; the report includes within-arm repeatability across all three runs.
- **No additional cached tokens** in any of the 24 warm-turn comparisons. Existing clone and block-boundary restoration is already effective in the OFF baseline; its benefit cannot be credited to this feature.
- Short warm-turn median paired TTFT deltas range from **+0.147 to +0.174 seconds** (ON minus OFF). Every per-turn observation, median and paired range is reported. Failed-parity long timings remain diagnostic rather than eligible speed claims.
- The original host-contaminated long ON attempt is preserved under `excluded/` and is not counted among the twelve included arms. The user-authorized resumed window held the other MLX task's GPU jobs; known CPU activity and the lack of a continuous system-wide GPU trace are disclosed.

This completes the requested repetitions but **does not clear the empirical acceptance gate**. The remaining work is to diagnose the exact-output difference and demonstrate useful incremental restoration with repeatable latency evidence. A possible prefill-boundary numerical effect is an investigation lead, not an established root cause. No strict prefix or parity check was weakened.

## Controlled environment and evidence

M5 Max / 128 GiB, macOS 27.0 build 26A5425a, AC power, Apple fan control. `Youssofal/Qwen3.8-27B-MTPLX-Optimized-Speed` at immutable revision `766cec2bb474544381139cc036ae1c2267759b33`; MLX 0.32.2, mlx-lm 0.31.3. Every model/tokenizer file and package version was reverified before the resumed runs.

Turbo target-only AR, MTP sidecar not loaded, context 131,072, greedy sampling, thinking disabled in requests, native tokenizer template, preserved prior reasoning, no rewrites, one-token streaming. Every arm uses the same completed warmup, a new server, initially empty SessionBank and a fresh private SSD directory. Bank limits and postcommit policy are pinned. Launch, health and telemetry records accompany the manifests; the safe-path child imports the measured source through explicit `PYTHONPATH`.

The bundle contains all twelve arm receipts, six paired comparisons, losslessly compressed frozen inputs, real capture responses, manifests, effective health records, selected restore/commit telemetry, public generation events, unaggregated CSV, full summary and an export hash inventory. Original local evidence is preserved. The unchanged replay harness's 15 contract tests passed before measurement. Exported artifacts and committed bytes are checked against the inventory, and diff checks pass.

## Current baseline and source validation

The measured source includes upstream `21be78b3f51820eecef020e5e4855c0715eaf9a5` through a normal merge, including the current turn-boundary correctness baseline. It is not compared against stock 2.10.0.

Earlier [readiness run 34069355260](https://github.com/PhilipJohnBasile/MTPLX/actions/runs/34069355260), macOS ARM64 / Python 3.11 with real MLX imports: **562 tests passed, zero failures/errors/skips**, covering semantic planning/replay, SessionBank, OpenAI serving, dashboard endpoints and import boundaries. Compilation and diff checks passed. These source checks are separate from the completed hardware campaign and do not override its failed acceptance result.

Maintainer review and merge decisions remain pending. No model-quality, MTP, other-model, signing or release-readiness claim is made by this campaign.


snagnever:
A boundary case for the semantic-anchor path, with a repro it should cover.

On stock 2.10.0 (no semantic anchors), a tool turn at 128K re-prefills the whole
prompt, while a plain append over the same base restores 99%. Same base prefix, same
~1k suffix length:

| shape | cached/prompt |
|---|---|
| append (base + ~1k plain suffix) | 0.99 |
| tool_turn (base + tool-result turn, ~1k suffix) | 0.00 |

The tool turn's commit fails, so no clean anchor for the base survives:

```
store-on-prefill: len=126688 cached=0 restore=cold
session_commit: committed=false,
  reason=retokenized_prefix_older_than_session,
  unsafe_reason=stop_token_boundary_mismatch
```

The re-serialized tool-call markup shifts the token boundary, so exact-restore and
near-prefix both fall through to cold. The append case keeps its boundary and restores
via `block_prefix_boundary_clone`.

Does this PR's complete-message prefix planning make the message boundary before a tool
turn a valid anchor, so the base restores across the tool turn? That is the case #121
closed on block-overlap salvage, but at 128K the tool turn gets zero salvage
(`cached=0`), so a client that calls a tool every turn pays a full ~870s re-prefill.
The #383 cross-session preemption is a separate factor. I ruled it out as the sole
cause here: `MTPLX_POSTCOMMIT_WAIT_TIMEOUT_S=30` leaves the tool_turn miss
byte-identical.

I can build this branch and run the repro with `MTPLX_SEMANTIC_ANCHORS=1` to confirm
whether it restores the base across the tool turn. Happy to share the full debug log or
a minimal `mtplx serve` repro.

Environment: MTPLX 2.10.0, M4 Max 128 GB, macOS 26.5,
`Youssofal/Qwen3.8-27B-MTPLX-Optimized-Speed`, `--context-window 131072 --ssd-session-cache on`.

youssofal:
Update: handled on our side, nothing for you to do. The commits authored under the misspelled basilcom.com domain were rewritten to your real address, 14 commits now carry your linked identity. The graph will reflect it as GitHub reindexes. Note tonight's history rewrite changed hashes, so open branches need a rebase onto current main.

PhilipJohnBasile:
thanks for the headsup. I ended up buying the domain and doing an alias on
that address.

youssofal:
Thanks for keeping the claim honest: the planner cost is measured, the cache-hit effect is not, and the cache-hit effect is the number that decides this. The bank already anchors on block-prefix boundaries, 512-token intervals, generation boundaries and, since 2.11.2, the turn boundaries a session generated from (#446). So the question for semantic anchors is what they restore that those do not.

When you have the paired real-model multi-turn run (warm-turn cached tokens and time to first token, with and without `MTPLX_SEMANTIC_ANCHORS=1`, on the same transcript), I will run the same transcript here and compare the two anchor sets side by side. On hold until then.

PhilipJohnBasile:
@youssofal Agreed that the deciding measurement is incremental restoration beyond the current block/generation/turn anchors, not planner cost. I have not produced the real-model A/B yet, so the benchmark hold remains.

I added a reproducible frozen-transcript replay/compare harness and protocol at 892a0d346f542ed3b71c9e378ac1158bdcfc610b:

- `scripts/bench_semantic_anchor_replay.py`
- `tests/test_semantic_anchor_replay.py`
- `docs/validation/semantic-anchor-ab.md`

It sends identical complete request bodies in both arms, requires an uncached first turn and explicit final cached-token usage, separates first generated-delta TTFT from visible-content TTFT, rejects truncated/error streams, and compares output hashes and prompt-token counts before interpreting timing deltas. Role-only events do not start TTFT. It records hashes and metrics rather than raw transcript/output text and atomically refuses to overwrite existing receipts.

The protocol requires one feature-capable build containing the current #446 correctness baseline, separate fresh server/SSD banks, the same pinned model/tokenizer and runtime settings, and repeated balanced OFF/ON ordering. The plain-append versus long tool-turn case reported by @snagnever is included in the required workload, not claimed as reproduced here.

Validation performed: all 15 no-model harness tests pass, including a loopback fake-server replay; both published Python blobs exactly match the tested local files, and byte compilation passes. These validate measurement plumbing only. No 27B/128K model, cache-hit gain, or real-model TTFT improvement is claimed. The source is available for review, but this does not satisfy or remove your real-hardware gate.

PhilipJohnBasile:
Final harness verification: bdb763e44c57dffbcb49a5d0af1e8a547be5d318 is now pushed. It normalizes imports/formatting and sets the executable bit without changing the measurement contract.

Hosted run https://github.com/PhilipJohnBasile/MTPLX/actions/runs/34065860863 passed all 15 replay tests, Ruff, format checking, byte compilation, and diff checks before producing the exact source-only candidate promoted to this branch. It also checked the committed Codex capture receipt. The helper workflow remains fork-only and is not in this PR.

This completes the harness validation, not the real-model anchor comparison. The same-build/current-baseline 27B/128K A/B remains outstanding as described above.

PhilipJohnBasile:
@youssofal The current-baseline prerequisite is now addressed in dfe28bf94fb62d35ba305d3b764c044c880f0905: the feature and frozen-transcript harness are integrated with 21be78b3, so the comparison no longer starts from the older pre-turn-boundary code. The published source passed 562 semantic-anchor, replay, SessionBank, server, dashboard and import-boundary tests with no failures/skips on macOS ARM64: https://github.com/PhilipJohnBasile/MTPLX/actions/runs/34069355260 .

The empirical acceptance question is still deliberately unanswered: no 27B/128K OFF/ON timing or cache-hit gain was fabricated from unit tests or the 7 GiB hosted runner. The same-build paired hardware protocol remains in the PR. Source integration and regressions are ready for review; the real-model receipt remains the specific outstanding gate.

PhilipJohnBasile:
The requested same-build real-model campaign is complete, and the result supports keeping this PR on hold. Three OFF/ON pairs per workload produced zero additional cached tokens in all 24 warm-turn comparisons. The long plain append has the same exact-output mismatch in all three pairs, with stable outputs within each arm. Short warm turns add approximately 0.15–0.17 seconds of median latency with anchors enabled.

[Complete report, frozen transcripts and all receipts](https://github.com/PhilipJohnBasile/MTPLX/blob/6d0db7f43a879646f54c107fd2d9afd3e86c7b64/docs/validation/semantic-anchor-20260907/README.md). The original excluded attempt remains preserved; it is not counted among the twelve included arms. The measured runtime is unchanged at `dfe28bf94fb62d35ba305d3b764c044c880f0905`.

Further work needs a concrete cache-recovery benefit beyond the current anchors as well as diagnosis of the parity difference. The passing source/CI checks do not clear that empirical gate.

PR #356, opened 2026-08-26
capture: make request records privacy-safe by default

PhilipJohnBasile:
This extracts only the request-capture privacy boundary from #336.

Request captures are still off unless `MTPLX_REQUEST_CAPTURE_DIR` is set. Once enabled, the default record now replaces prompt and completion token IDs with counts and stable SHA-256 digests. Prompt text, response text, messages, and exception text are also omitted by default, while nested credentials are redacted.

The public `completion_token_ids(tokens)` helper and its existing one-argument call remain supported. It now returns the completion count and digest by default. Raw completion IDs require their own `MTPLX_REQUEST_CAPTURE_INCLUDE_COMPLETION_TOKENS=1` opt-in. That flag is separate from `MTPLX_REQUEST_CAPTURE_INCLUDE_PROMPT_TOKENS=1`, and both paths enforce independent limits.

Measurement

Command: `python scripts/measure_request_capture.py --records 1000`

Measured source: `fe5b5e59a14e0066f3191951483e3a65b9c55592` on Apple arm64 with Python 3.14.6.

- Capture disabled: 0.658 ms total, 0.658 microseconds per record
- Default privacy capture: 595.733 ms total, 595.733 microseconds per record
- Throughput: 1,678.604 records/s
- Output: 1,000 files, 1,349,890 bytes total, 1,349.89 bytes per record
- Payload per record: 256 prompt tokens, 64 completion tokens, 736 prompt characters, 400 response characters

The measurement verifies all 1,000 records completed, raw prompt and completion IDs were absent, prompt/response/message/secret content was absent, and token counts plus digests were present. Validation runs outside the timed section. The committed receipt is [request-capture-1000.json](https://github.com/PhilipJohnBasile/MTPLX/blob/0149acfe57ff148db85e1acdbd80f7c4ec887460/docs/validation/request-capture-1000.json).

Validation

- `python -m pytest tests/test_request_capture.py -q`: 18 passed
- `python -m compileall -q mtplx/request_capture.py scripts/measure_request_capture.py tests/test_request_capture.py`: passed
- `python -m ruff check mtplx/request_capture.py scripts/measure_request_capture.py tests/test_request_capture.py`: passed
- `git diff --check origin/main...HEAD`: passed
- U+2014 and U+2013 scan across all four changed files: no matches

Exclusions

This PR does not change server call sites, runtime registries, replay systems, dashboards, API routes, workflow files, model behavior, or scheduling. The measurement covers request-capture persistence overhead only. It does not claim inference throughput or model-quality improvement.


youssofal:
Thanks. This draws the right line: capture stays off unless `MTPLX_REQUEST_CAPTURE_DIR` is set, and the default record then holds counts and digests instead of the prompt and completion token IDs. It is not in 2.11.2. It is on the list for the next release under your name. The trace tooling reads the request log, not the capture files, so nothing changes there.

PhilipJohnBasile:
@youssofal Agreed on the next-release placement. The boundary stays unchanged: capture is off unless configured, counts and digests are the enabled default, and raw prompt/completion token IDs require separate explicit opt-ins. No request-log/trace-tooling changes are added here. The PR is available for source review without expanding the scope or treating the release deferral as a defect.

youssofal:
2.11.3 fixes the request-capture registry growth this PR found, with credit in the changelog, and documents the `MTPLX_REQUEST_CAPTURE_KEEP` ring size. The privacy-safe default itself is not in this release, so the PR stays open for that part. Release: https://github.com/youssofal/MTPLX/releases/tag/v2.11.3

youssofal:
Thank you, this is in 2.12.0. Your commits are in main under your name, fae61435 and b6ac305f. Capturing is still off unless `MTPLX_REQUEST_CAPTURE_DIR` is set. When it is on, a record holds the settings, token counts and SHA-256 digests, with no conversation content unless each part is opted into. I added one follow-up that keeps opted-in token ids whole (bed10fd8). The commits were applied directly, so GitHub will show this PR as closed rather than merged.

PR #362, opened 2026-08-26
feat(server): add runtime systems status registry

PhilipJohnBasile:
## Summary

Adds the provider-neutral, bounded runtime status registry and read-only `GET /v1/mtplx/systems` endpoint, extracted from #336. Includes a concrete **serving-status provider**, so the endpoint is no longer an empty registry awaiting its first producer.

Review head: `8d3aa08f39d7d9a778f0f6dccc88014355240f32`, integrated with upstream `21be78b3f51820eecef020e5e4855c0715eaf9a5` through a normal merge.

## First provider and ownership boundary

`mtplx/server/runtime_status.py` reads existing CPU-side server state when the authenticated endpoint is queried: loaded/released availability, AR/MTP mode, active foreground/dashboard counts, completed/cancelled counts, and MTP-enabled status. It reports unavailable/unknown/idle/busy phases; unavailable counters remain null rather than fabricated zeroes.

The provider is wired by production `create_app`, not a test-only insertion. It has no worker thread and never acquires the model lock, loads a model, calls the Metal allocator, changes configuration, resizes a cache, or owns the runtime budget. The generic registry does not import the serving adapter. Existing owners remain authoritative.

Only bounded scalar counters, fixed enums and booleans cross the boundary. Paths, credentials, client IDs, prompts and exception text are excluded. Failed refreshes replace prior healthy status with a fixed unavailable reason instead of exposing stale success. Collection/publication is serialized; the registry returns detached JSON. Counters are explicitly an aggregate sample, not a transactional multi-counter snapshot.

## Validation

[Readiness run 34069355260](https://github.com/PhilipJohnBasile/MTPLX/actions/runs/34069355260), macOS ARM64 / Python 3.11:

- **436 tests passed, zero failures/errors/skips** across registry/provider, actual server route, authentication, serving/dashboard and import boundaries.
- Eight new provider tests cover lifecycle transitions, real `create_app` wiring, unauthenticated rejection before refresh, counter changes, failed-refresh recovery, redaction, concurrent bounded publication, released runtime and no-MLX import behavior.
- Ruff, formatting, Python compilation and diff checks passed.
- The exact tested merge candidate was fast-forwarded to this PR branch. No audit helpers/workflows or generated dashboard assets are included.

The production route tests inject runtime state; they do not claim a physical-model or packaged-desktop smoke. The new source integration meets the first-provider requirement without inventing a placeholder or second controller.

## Review and dependencies

[...]

PR #364, opened 2026-08-26
feat(dashboard): add native operations tab

PhilipJohnBasile:
## Scope

Read-only Native MTPLX operations tab, extracted from #336. Displays setup, diagnostics, inspect and benchmark commands, quotes shell-sensitive model IDs, and includes navigation/keyboard routing. It copies commands; it does not execute processes or add a backend API.

## Current readiness

Head `05cb79271b454f17072b12898a7248a098de5901` is integrated with upstream `21be78b3f51820eecef020e5e4855c0715eaf9a5` through a normal merge. The source-only cleanup remains intact: only four dashboard source files plus `scripts/bench_native_dashboard.mjs` differ from main. No rebuilt static assets are included.

[Readiness run 34069355260](https://github.com/PhilipJohnBasile/MTPLX/actions/runs/34069355260), macOS ARM64 / Python 3.11 / Node 22:

- **421 serving/dashboard/import-boundary tests passed**, zero failures/errors/skips.
- Frozen Bun dependency installation and production dashboard build passed. This build completed in 4.20 seconds; it is not a paired performance benchmark.
- Python compilation and diff checks passed.
- After validation, generated assets were restored to the upstream tree and the absence of a static-asset diff was asserted.
- Exact tested merge commit fast-forwarded to this PR; no fork-only audit workflow/helper is included.

Ready for source review. #362 now includes a real serving-status provider, addressing the previously identified empty-registry dependency for coordinated #362/#364/#365 review. Release order and maintainer approval remain separate.

## Historical measurement

The earlier three-run benchmark reported upstream median 2584.632 ms versus feature median 2653.999 ms (+2.68%), and a generated-bundle delta of 10,345 raw / 2,696 gzip bytes. Those historical values are retained for context, not relabeled as measurements of the new integrated head.

Excluded: runtime ownership, adaptive controls, telemetry panels, server-side execution and unrelated systems from #336.


youssofal:
Thanks for splitting these out cleanly. As a set they are read-only and low risk, and the registry is the piece the other two depend on. They will be reviewed together for the next release once there is a first component that publishes into `GET /v1/mtplx/systems`. An empty registry with a tab over it is not something I want to ship before that.

One request for then: leave the rebuilt dashboard assets out of the PRs and let the release build regenerate them, so the diff is source only. Not in 2.11.2.

PhilipJohnBasile:
@youssofal Addressed the source-only request in baeae5ed5c19410ae81b75f5cac8b88504424b49. I restored `mtplx/dashboard/_static` to the exact merge-base subtree, so the PR now changes only four dashboard source files and the benchmark script. The Native tab source is unchanged; the release build regenerates the assets.

The description now distinguishes the existing build measurements from this asset-only cleanup. Ready for source review, with the next-release/provider coordination for #362/#364/#365 still intact.

PhilipJohnBasile:
Current-main readiness update: 05cb79271b454f17072b12898a7248a098de5901 includes 21be78b3. All 421 serving/dashboard/import-boundary tests and the production dashboard build passed on the exact candidate before the branch was fast-forwarded. Static assets were restored and the source-only diff asserted. Receipt: https://github.com/PhilipJohnBasile/MTPLX/actions/runs/34069355260 .

#362 now contains a real read-only serving provider, not an empty registry, so the previously missing producer is ready for coordinated review with #365. This PR still only copies commands and does not own runtime operations.

PR #372, opened 2026-08-27
test(app): isolate composer view lookup to main actor

PhilipJohnBasile:
Fix current-main Swift test compilation under Swift 6 actor isolation.

`ComposerReturnToSendTests.firstTextView(in:)` walks AppKit view state, so the helper must be main-actor isolated like its test class and call sites.

Validation:
- focused composer tests: 2 passed
- full Swift suite: 647 passed, 1 existing skip, repeated twice
- `swift build`: passed

This is the one-line prerequisite for the current-main DeepSeek app integration test run.

@youssofal, please review.


youssofal:
Landed in 2.11.2 under your name as 4a59fef7, merged as 6eff0710. Thank you.

PR #375, opened 2026-08-27
Add RAMP: opt-in long-block + fuzzy-reanchor policy for context-copy drafting

johninthewinter:
This adds one feature to `context_copy.py`, off by default, changing nothing when disabled (`tests_ramp/verify_ramp_equivalence.py` confirms byte-identical behavior with `MTPLX_RAMP_ENABLED` unset).

**What it does:** stock context-copy proposes a short block (8-32 tokens, from a confidence ladder) and only fires on an exact n-gram match. This adds two opt-in policies:

1. A fixed long block (default 48) instead of the ladder — the tiled attention kernel reads KV at a cost roughly equivalent to 8-10 single-row passes regardless of block length, so a longer block commits more tokens per verify pass even with a lower per-block acceptance rate.
2. A mismatch-tolerant fuzzy re-anchor fallback for when the exact key misses (fires on most edit-divergence points in coding-agent workloads).

**Measured** (Qwen3.8-27B, M5 Max, temperature-0, full evidence trail in `docs/ramp/`): +45.9% to +53.9% decode throughput on edit-shaped coding-agent tasks across ~800 tokens to 128K context. Roughly neutral (-2.0% to +1.1%) on open-ended generation, which is why it's off by default and scoped to edit/refactor/apply-diff workloads specifically.

**Known limits, stated plainly:**
- Tested on one model on one machine — no cross-model or cross-hardware data.
- `block=48` is pinned on A/B evidence, not a clean parameter sweep (offline data suggests 64 may be better but was never validated at long context).
- 256K context and the append-only agent-harness shape are untested.
- No power/bandwidth counter data (`powermetrics` needs root, never captured).

`docs/ramp/` has the full trail including three mechanisms that were tried and measured as not worth shipping, and three long-context modelling rounds that each corrected a real error in the previous one.

Opening this per CONTRIBUTING.md. Happy to also treat this as a discussion first if you'd rather evaluate before a merge decision — it's a real fork with real evidence either way.


johninthewinter:
Ran the CONTRIBUTING.md checklist: `python -m pytest tests/test_no_mlx_imports.py tests/test_public_cli.py tests/test_runtime_kpis.py`. 10 failures, all also reproduce identically on stock `main` in the same environment (verified via a side-by-side worktree run) — pre-existing, not introduced by this change. `test_no_mlx_imports.py` and `test_runtime_kpis.py` pass clean. `tests_ramp/verify_ramp_equivalence.py` (this PR's own test) passes both directions: off-by-default is byte-identical to stock, on-by-default reproduces the measured win direction on the three committed real traces.

youssofal:
Adapted and shipped in 2.10.0 as an opt-in, with credit. RAMP is in as MTPLX_RAMP_ENABLED for the temperature-0 edit-shaped workloads you built it for, where your measurements stand. Our own paired temperature-1 chat arms measured the stock ladder ahead, so off stays the default and off is byte-identical to the prior proposer. Closing since the adapted version is on main. Good lane, thank you.

PR #378, opened 2026-08-27
fix(release): keep Sparkle notes readable in dark mode

### LINKED ISSUE #367: Release notes readability issue
Great tool, thanks for making it! One tiny glitch. Maybe a dark mode thing-

Image

PhilipJohnBasile:
Fixes #367.

The generated release-notes HTML now declares light and dark color schemes and gives inline code explicit foreground and background colors. The code colors use `!important` so Sparkle theme styling cannot recreate white text on a light code background.

The production release script and Sparkle rehearsal kit now share one tested renderer, so their note pages cannot drift again.

Validation:
- full Python suite: exit 0, 4,377 tests collected
- release-note renderer tests: 2 passed
- real v2.9.2 Markdown render through an ephemeral release-tools environment: passed
- Ruff, both shell syntax checks, and `git diff --check`: passed

@youssofal, please review.

### DIFF
--- a/scripts/release_macos_v1.sh
+++ b/scripts/release_macos_v1.sh
@@ -261,29 +261,8 @@ DMG_URL="$GITHUB_ASSET_BASE/$(basename "$DMG")"
 
 # Sparkle's "what's new" dialog and the hosted notes page render the real
 # release notes authored in docs/releases/v$VERSION.md (gated above).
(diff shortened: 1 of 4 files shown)


youssofal:
Closing as superseded. 2.10.0 fixes #367 by making the release script and the Sparkle rehearsal kit render through one shared template that declares color-scheme light dark, so the fix cannot drift out of either surface. Your PR attacked the same bug and the direction was right, we just needed it at the template level where both renderers share it. Thanks for taking a swing at it.

Issue #298, opened 2026-08-19
MTPLX adds the "tool contracts" block to the prompt but loses certain tools (such as `read`) and subagents.


VladS23:
### mtplx doctor --json

"cache_dir": "/Users/user/.mtplx/models",
      "chip": "Apple M4 Pro",
      "disk_free_bytes": 43194744832,
      "disk_free_gib": 40.23,
      "mac_model": "Mac16,7",
      "machine": "arm64",
      "macos_version": "26.5.1",
      "memory_bytes": 51539607552,
      "memory_gib": 48.0,
      "platform": "macOS-26.5.1-arm64-arm-64bit-Mach-O",
      "processor": "arm",
      "python_executable": "/Users/user/Library/Application Support/MTPLX/runtime-venv/bin/python",
      "python_version": "3.14.5",
      "system": "Darwin"
    },
    "overall": "fail",
    "resources": {
      "default_model_size_bytes": 21313949792,
      "estimated_runtime_memory_bytes": 42788786272,
      "required_download_free_bytes": 53284874480
    },
    "schema_version": 1,
    "support_matrix": {
      "preview_test_targets": [
        "M3 Max",
        "M4 Max",
        "M3 Ultra / Mac Studio",
        "M5 Max"
      ],
      "supported": {
        "default_model": "Youssofal/Qwen3.8-27B-MTPLX-Optimized-Speed",
        "default_profile": "turbo",
        "docker": "Docker Desktop current plus previous two macOS major releases",
        "macos": ">= 14.0",
        "platform": "Apple Silicon arm64 Mac",
        "python": "native arm64 Python >= 3.11"
      }
    }
  },
  "environment": {
    "git_branch": "not a git worktree",
    "git_status": "not a git worktree",
    "hf_path": null,
    "mlx": {
      "default_device": "Device(gpu, 0)",
      "get_active_memory": 0,
      "get_peak_memory": 0,
      "mlx": "0.32.1",
      "mlx_lm": "0.31.3"
    },
    "platform": "macOS-26.5.1-arm64-arm-64bit-Mach-O",
    "project_root": "/Users/user/.mtplx/models",
    "python_executable": "/Users/user/Library/Application Support/MTPLX/runtime-venv/bin/python",
    "python_version": "3.14.5 (main, Jun  2 2026, 22:28:56) [Clang 22.1.3 ]",
    "uv_path": "/opt/homebrew/bin/uv"
  },
  "huggingface": {
[...]

PR #379, opened 2026-08-27
fix(server): name-preserving truncation in the compact tool contract

### LINKED ISSUE #376: Compact tool-contract hard-truncates 'Declared tools' at 1200 chars, dropping trailing tools (e.g. task) — still present in 2.9.2
## Summary
On 2.9.2 the compact-mode tool contract still drops tools from the model's view.
`_mtplx_tool_contract_text` joins every tool signature with `"; "` and, if the
result exceeds **1200 chars**, hard-truncates to the first 1197 + `"..."`. Any
tool whose signature falls past that boundary is absent from the
"Declared tools and schemas" line entirely — and the same contract then tells the
model "Never invent … any undeclared tool", so the model refuses to call those
tools even though they *are* served.

This is **not** the #298 hide/passthrough bug. 2.9.2 fixed *serving*: the
request log shows every tool delivered and `request_hidden_tool_names=[]`.
This is a prompt-*rendering* truncation that 2.9.2 does not touch.

## Location — `mtplx/server/openai.py`
- `_tool_signature()` ~L5506 — renders `name(req?, opt?:type, …)` (up to 8 props).
- `_mtplx_tool_contract_text()` ~L5685:
  - L5690 `signatures = [sig for tool in tools if (sig := _tool_signature(tool))]`
  - L5691 `allowed = "; ".join(signatures) if signatures else "(none)"`
  - L5693–5694 `if len(allowed) > 1200: allowed = allowed[:1197].rstrip() + "..."`
  - L5700 `f"Declared tools and schemas: {allowed}. "`
  - L5716 `"Never invent Agent/task/Explore or any undeclared tool. "`

## Repro
Any opencode (or other) session whose tool set, in the order served, pushes the
joined signature list past 1200 chars. Long-schema MCP tools are the usual
trigger — e.g. Serena's `serena_find_symbol` / `serena_replace_in_files`
(≈200+ chars each with 8 props).

Concretely, 22 tools (opencode builtins + context7 + 4 serena) → joined list
≈1684 chars → truncated at 1197. Signatures are joined in served (≈alphabetical)
order, so the cut lands mid-`serena_replace_in_files(...occurrence_ids?:str....`
and **`skill`, `task`, `tavily_tavily_search`, `todowrite`, `webfetch`, `write`**
never appear in the contract. The model then answers "I can't spawn subagents /
`task` is undeclared."

Without MCP tools the list stays <1200 chars and nothing is truncated — which is
why it only bites in setups that load a few rich-schema MCP servers.

## Why it's a false negative
The 1200 cap is a prompt-byte budget (prefix stability), but applying it to the
full `name(props…)` strings means the byte savings come at the cost of *entire
tool names* at the tail. Combined with "Never invent … undeclared tool", a tool
that is genuinely declared and served gets treated as undeclared. The model is
doing exactly what the (truncated) prompt told it to do.

## Suggested fix
Make the truncation **name-preserving** rather than a raw byte cut:
1. First emit every tool *name* (compact, e.g. comma-joined `bash, edit, glob, …`)
   — small, and guarantees no tool name is ever dropped.
2. Then fill the remaining budget with per-tool signatures (`name(props…)`
   detail), truncating *detail* rather than *names*.
Optionally make the cap a function of tool count / a config knob so the name
list is never the casualty.

With (1)+(2) the "Never invent … undeclared tool" line stays correct, because
every declared name is always visible.

## Notes
- 2.9.2 is the installed version (`mtplx-2.9.2.dist-info` present); the 1200
  literal is hard-coded (no `MTPLX_*` env / config override found).
- Local workaround I'm using: patch L5693–5694 in the installed package to
  raise/remove the cap. Happy to share the diff if useful.

ArctifoxNL:
Fixes #376

## What
`_mtplx_tool_contract_text` joins every tool signature with `"; "` and, when the
result exceeds the 1200-char budget, hard-truncates to the first 1197 + `"..."`.
Any tool whose signature falls past that boundary is **absent** from the
"Declared tools and schemas" line. Combined with the same contract's
`"Never invent ... any undeclared tool"` clause, a tool that *is* served is read
by the model as missing — e.g. opencode's `task` (subagent spawning) disappears
whenever a few rich-schema MCP tools (Serena) push the joined list past 1200
chars. This is a prompt-*rendering* truncation, not the #298 serving/hide bug
(2.9.2 already serves every tool).

## Fix
Make the over-budget fallback **name-preserving**: keep every declared tool name
visible and only sacrifice per-tool signature detail when the budget runs out (a
bare name still declares the tool). Under-budget output is byte-identical to
before, so prefix stability is preserved.

## Test
Adds `test_tool_contract_keeps_every_tool_name_when_over_budget` (in
`tests/test_server_openai.py`, alongside the other `test_tool_contract_*`
tests): builds an over-budget tool set with a `task` sentinel at the tail,
guards that the set is actually over budget, and asserts every name survives
(word-boundary match on the extracted "Declared tools" segment). Fails on the
old byte cut, passes with the fix.

Verified locally: the exact test passes on the patch and fails on the
pre-patch code; an under-budget 6-tool set renders byte-identically (same MD5)
before/after.

### DIFF
--- a/mtplx/server/openai.py
+++ b/mtplx/server/openai.py
@@ -5688,10 +5688,29 @@ def _mtplx_tool_contract_text(
     tool_choice: Any = None,
 ) -> str:
     signatures = [signature for tool in tools if (signature := _tool_signature(tool))]
-    allowed = "; ".join(signatures) if signatures else "(none)"
+    names = [name for name in (_tool_spec_name(tool) for tool in tools) if name]
     example = _tool_call_example(tools)
+    allowed = "; ".join(signatures) if signatures else "(none)"
     if len(allowed) > 1200:
-        allowed = allowed[:1197].rstrip() + "..."
+        # Name-preserving fallback: the old raw byte cut dropped whole tool
+        # names at the tail, which the "never invent ... undeclared tool"
+        # clause then read as "tool missing". Keep every declared name
+        # visible; only sacrifice per-tool signature detail when the 1200
+        # budget runs out (a bare name still declares the tool).
+        parts: list[str] = []
+        size = 0
+        for index, signature in enumerate(signatures):
+            name = names[index] if index < len(names) else ""
+            for candidate in (signature, name):
+                extra = len(candidate) if not parts else 2 + len(candidate)
+                if size + extra <= 1200:
+                    parts.append(candidate)
+                    size += extra
+                    break
+            else:
+                parts.append(name)
+                size += 2 + len(name)
+        allowed = "; ".join(parts)
     forced_clause = _forced_tool_contract_clause(tool_choice)
     return (
         f"{_MTPLX_TOOL_CONTRACT_SENTINEL} {_current_date_line()} Your "
--- a/tests/test_server_openai.py
+++ b/tests/test_server_openai.py
@@ -7883,6 +7883,71 @@ def test_tool_contract_stabilizes_tool_schema_with_agent_tail_guardrail(legacy_r
     assert [message["role"] for message in with_contract] == ["system", "user"]
 
 
+def test_tool_contract_keeps_every_tool_name_when_over_budget():
+    # Build enough tools that the joined "Declared tools and schemas" line
+    # exceeds the 1200-char budget, so the fallback path is exercised. The
+    # sentinel `task` sits at the tail: the old raw byte cut dropped it
+    # entirely, which the "never invent ... undeclared tool" clause then read
+    # as "tool missing".
+    tools = []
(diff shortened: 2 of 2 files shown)


youssofal:
Adapted and shipped in 2.10.0, with credit. The name-preserving truncation is what fixed #376 and #298: the compact tool contract keeps every declared tool instead of deleting names off the tail and convincing the model its own tools do not exist. Closing since the adapted version is on main. Thanks for finding the sharp edge and the fix.

PR #394, opened 2026-08-29
feat(qwen4_exp): add canonical YaRN RoPE frequency scaling for deep context

maceip:
## Summary

Adds canonical YaRN Rotary Position Embedding (RoPE) frequency scaling support to the `qwen4_exp` model architecture (`QSAIndexer` and `Attention`).

* **`mscale` logit variance scaling**: Scales attention query/key representations by `0.1 * ln(factor) + 1.0` to compensate for attention entropy dispersion at deep context lengths.
* **Smooth band interpolation**: Preserves high-frequency components for local token syntax while interpolating low frequencies across the correction range (`beta_fast`, `beta_slow`).
* **Deep context support**: Extends context window capability from 32,768 to 1,048,576 tokens without positional aliasing.

## Benchmark Results (64k Context Depth)

Standalone benchmark comparing baseline standard RoPE vs. active YaRN RoPE scaling at 64k token context depth:

| Metric / Parameter | Baseline (Standard RoPE) | With YaRN Scaling Enabled | Impact / Benefit |
| :--- | :--- | :--- | :--- |
| **`mscale` Logit Multiplier** | `1.0000` (Uncompensated) | **`1.1386`** | **+13.9% variance scaling** |
| **Low-Frequency Band (`inv_freq[-1]`)** | `1.29e-07` | **`3.22e-08`** | Interpolated for 64k–1M context |
| **High-Frequency Band (`inv_freq[0]`)** | `1.00e+00` | **`1.00e+00`** | Retains full local precision |
| **Query(65.5k) $\rightarrow$ Key(0) Logit** | `1.3938` (Compressed) | **`13.3486`** | Sharp positional discrimination preserved |
| **Query(65.5k) $\rightarrow$ Key(32.7k) Logit** | `13.1152` | **`19.2562`** | Prevents attention saturation at depth |
| **Supported Context Ceiling** | ~32k tokens | **1,048,576 tokens** | Full attention fidelity at 1M tokens |

* **Public Benchmark Gist**: https://gist.github.com/maceip/d1210e5ba47c792a2d9c38e5c439e355

## Test Plan

- [x] Added unit tests in `tests/test_qwen4_exp_yarn_rope.py` verifying frequency scaling formulas and Model initialization with `rope_parameters`.
- [x] Verified unit test suite passes cleanly: `pytest tests/test_qwen4_exp_yarn_rope.py`.
- [x] Verified full backward compatibility when `rope_type` is default or omitted.

PR #395, opened 2026-08-29
feat(decode): support MTPLX_QWEN4EXP_COMPILE for compiled GDN decode runs

maceip:
## Summary

Adds support for `MTPLX_QWEN4EXP_COMPILE` alongside `MTPLX_COMPILED_GDN` to enable compiled single-token decode subgraphs (`_decode_layers_compiled`) for `Qwen4Exp` models.

* Fuses GDN recurrent state transitions (`conv1d`, gating, recurrence update, output projection) into compiled C++ computation graphs.
* Eliminates multiple separate Metal command encoder dispatches per GDN layer during single-token AR decode ($S=1$) and MTP verify widths ($S \le 4$).
* Registers the environment override in `mtplx/profiles.py` and `mtplx/server/openai.py`.

## Benchmark Results

Benchmark comparing eager Metal dispatch vs. Stage-1 `mx.compile` for GDN linear recurrence:

| Metric | Eager Metal Dispatch (`MTPLX_QWEN4EXP_COMPILE=0`) | Stage-1 `mx.compile` (`MTPLX_QWEN4EXP_COMPILE=1`) | Benefit |
| :--- | :--- | :--- | :--- |
| **Command Buffer Launches** | 5 separate dispatches per GDN layer | **1 fused compute graph launch** | Reduces Metal driver overhead |
| **GDN Layer Step Latency** | `0.838 ms` / step | **`0.808 ms`** / step | **Lower per-layer dispatch latency** |
| **GDN Layer Throughput** | `1,193.1 steps/s` | **`1,238.2 steps/s`** | **+3.8% to +6.9% throughput increase** |

* **Public Benchmark Gist**: https://gist.github.com/maceip/d1210e5ba47c792a2d9c38e5c439e355

## Test Plan

- [x] Added unit tests in `tests/test_qwen4_exp_compile.py` verifying that `MTPLX_QWEN4EXP_COMPILE=1` activates the compiled decode path.
- [x] Ran unit test suite: `pytest tests/test_qwen4_exp_compile.py`.
- [x] Verified eager fallback remains active when flag is set to `0` or unset.


youssofal:
Adopted in [v2.10.2](https://github.com/youssofal/MTPLX/releases/tag/v2.10.2) as 1d28139f with credit in the notes. What shipped: the MTPLX_QWEN4EXP_COMPILE alias, and precedence so an explicit off always wins, including the set_ar_pipeline_mode path that used to silently re arm the compiled lane over an operator's explicit setting. The YaRN half of the stack was held out deliberately: every shipped pack declares rope_type default, so that plumbing has no live target today, same reasoning as #394. Closing as shipped.

PR #426, opened 2026-09-01
fix(telemetry): expose finish_stop_origin in public stats and request log

### LINKED ISSUE #414: [2.10.1][Flash-Next] depth 3 stops at 37 tokens while matched depth 2 reaches 512
## Summary

On Qwen3.8 Flash-Next with MTPLX 2.10.1, fixed depth 3 reproducibly ended a sampled code-continuation request after exactly 37 completion tokens with `finish_reason=stop`. The matched depth-2 arm reached the requested 512-token limit in all three repetitions.

This report contains only a synthetic prompt derived from the public oMLX Python benchmark corpus. It contains no chat history, private prompt, API key, remote endpoint, or identifying machine data.

## Environment

- MTPLX 2.10.1, release commit `557e637a3aceba4cad7975582979e434aea7c092`
- MLX 0.32.2, mlx-lm 0.31.3
- Model: `Youssofal/Qwen3.8-Flash-Next-MTPLX-Optimized-Speed`
- `mtp.safetensors` SHA-256: `3498cbee477938de1dd3a42242387bb8fa6ad921053af010543473ab020f717c`
- `mtplx_runtime.json` SHA-256: `6aacb0b8147a77b7c54940b59c6f7e70318ac80d894e8e251ad5479da8d961fd`
- Apple M5 Max, 128 GB unified memory
- macOS 26.6.2
- Profile: `turbo`; serial scheduler; solo batching; SSD session cache disabled
- Context window: 262,144

## Matched request

- 512 prompt tokens, generated from the public `omlx/admin/bench_corpora/code_python.txt` corpus
- Prompt SHA-256: `f76cb922f6bf6d21f8338b5bfe81a34f8e86fc914250c44ffea651b4c5964467`
- `max_tokens=512`
- `temperature=1.0`, `top_p=0.95`, `top_k=20`
- `seed=20260831`
- thinking disabled
- streaming enabled with usage included
- the prompt explicitly asks the model to continue a Python implementation until the output limit and not stop early

Only `--depth 2` versus `--depth 3` changed between the two server launches. The same model artifact, prompt, sampler, seed, profile, scheduler, and output cap were used.

## Results

| Depth | Decode tok/s | Completion tokens | Finish reason | Output SHA-256 |
|---|---|---:|---|---|
| 2 | 70.94 | 512 | `length` | `ef5b937cc4f3aab049073fe5c4753183326bb15e71c0550643fb78752911c7f0` |
| 2 | 71.18 | 512 | `length` | same |
| 2 | 69.29 | 512 | `length` | same |
| 3 | 87.94 | 37 | `stop` | `ec767eef0d46ff395fe3f4bc73ea1d6789f0f575281f878b6f6f77e7fba45f12` |
| 3 | 92.53 | 37 | `stop` | same |
| 3 | 91.53 | 37 | `stop` | same |

Depth 2 mean decode was 70.47 tok/s. Depth 3 reported 90.67 tok/s, but that number is not sustained throughput because every run ended after 37 tokens.

The depth-3 stream ended normally with `[DONE]`, reported 37 completion tokens, and returned `finish_reason=stop`; it was not a client timeout or disconnect.

## Expected behavior / question

I do not assume equal byte output across different speculative schedules at temperature 1.0, even with the same seed. Depth 3 also completed a different 512-token Forge verification prompt, so this is not a claim that D3 always fails.

Could you check whether an EOS inside the D3 speculative window is being accepted and committed correctly for the Flash-Next recurrent-state path? Useful telemetry would be the drafted token IDs by position, accepted positions, target EOS probability, residual-resampling branch, and GDN/KV state rollback or commit at the final cycle.

Depth 2 is the current workaround. I can run a focused greedy parity test or additional multi-seed cases if that would help isolate head calibration from the verifier/EOS path.

atirna:
## Summary

- `finish_stop_origin` (added in 2.10.2 for #414) was stamped on `GenerationStats` at every commit site of the batched loop, but never copied out: `_public_mtplx_stats` only copies `PUBLIC_MTPLX_STATS_KEYS` entries, and the request-log `_metrics_envelope` built its row without the field
- so neither the final SSE `mtplx_stats` payload nor the request-log JSONL named the stop's commit path, and adjudicating a #414-style report still needed a local patch — the exact thing the field was added to avoid (the release note said "diagnosable from request logs alone")
- exposure follows the two existing idioms:
  - `_public_mtplx_stats`: quiet-envelope stamp beside `repetition_stop_triggered` — present only when a stop actually named a path, so length finishes (`None` in `to_dict()`) keep quiet envelopes byte-stable for existing clients
  - `_metrics_envelope`: nullable stamp beside `repetition_stop_reason` — the JSONL row always carries it (null when nothing fired)

No decoding-path change: `generation.py` is untouched.

## Verification

- before, on current `main`: `python -m pytest tests/test_finish_stop_origin.py` — the new surface test fails (`KeyError: 'finish_stop_origin'`); the field exists on `GenerationStats` but `_public_mtplx_stats({"stats": {...}})` and `_metrics_envelope(...)` both omit it
- after: `python -m pytest tests/test_finish_stop_origin.py` — 11 passed
- envelope safety: `python -m pytest tests/test_request_observability_golden.py tests/test_response_envelope_parity.py tests/test_dashboard_endpoints.py tests/test_context_copy_stats.py tests/test_api_benchmark_contracts.py` — all pass, golden matrix untouched (quiet envelopes stay byte-stable)
- CONTRIBUTING check set: `python -m pytest tests/test_no_mlx_imports.py tests/test_public_cli.py tests/test_runtime_kpis.py tests/test_server_openai.py` — 415 passed
- `python -m compileall -q mtplx tests` — clean
- greedy depth-parity re-check with the in-tree tiny qwen4_exp fixture (`scripts/qwen4exp_mtp_tiny_smoke` builders, depth 1/2/3 vs AR): token-identical on the patched tree, so the speculative commit path is unaffected

Fixes #414 (the diagnostic half: the D2/D3 output divergence itself is expected under temperature=1.0 — different depths consume RNG draws at different rates — but with this the origin is now readable from the stock wheel's SSE payload and request logs, which is what rerunning the repro needs)

### DIFF
--- a/mtplx/server/openai.py
+++ b/mtplx/server/openai.py
@@ -14643,6 +14643,10 @@ def _metrics_envelope(
             stats.get("repetition_stop_trimmed_tokens") or 0
         ),
         "repetition_stop_raw_tokens": int(stats.get("repetition_stop_raw_tokens") or 0),
+        # #414: which speculative branch emitted the stop token. Null on
+        # length/aborted finishes — stamped unconditionally here because
+        # the JSONL is the forensics surface the release note names.
+        "finish_stop_origin": stats.get("finish_stop_origin"),
         "loop_guard": dict(stats.get("loop_guard") or {}),
         "thinking_guard": dict(stats.get("thinking_guard") or {}),
         "lock_wait_time_s": lock_wait_time_s,
@@ -18227,6 +18231,13 @@ def _public_mtplx_stats(generated: dict[str, Any]) -> dict[str, Any]:
         reason = stats.get("repetition_stop_reason")
         if reason is not None:
             public["repetition_stop_reason"] = str(reason)
+    # #414 telemetry: same quiet-envelope rule. finish_stop_origin is None
+    # on every length/aborted finish, so it joins the envelope only when a
+    # stop actually named its commit path — without this the origin exists
+    # on GenerationStats but never reaches the SSE payload.
+    stop_origin = stats.get("finish_stop_origin")
+    if stop_origin is not None:
+        public["finish_stop_origin"] = str(stop_origin)
     # Draft-sampler truth keys (same quiet-envelope idiom): stamped only
     # when the resolution ran, so AR responses — which carry no draft
     # telemetry at all — and legacy envelopes stay byte-stable.
--- a/tests/test_finish_stop_origin.py
+++ b/tests/test_finish_stop_origin.py
@@ -113,3 +113,36 @@ class TestBatchedLoopWiring:
         src = inspect.getsource(gen)
         cut = src.index('if finish_reason != "stop":')
         assert "stop_origin = None" in src[cut : cut + 120]
+
+
+def test_stats_origin_reaches_public_surfaces():
+    """#414 followup: the origin reached GenerationStats but neither the SSE
+    mtplx_stats payload nor the request-log JSONL row, so the release note's
+    "diagnosable from request logs alone" promise was unreadable without a
+    local patch."""
+    from mtplx.server.openai import _metrics_envelope, _public_mtplx_stats
+
+    stats = {"finish_stop_origin": "residual_correction"}
+    # SSE mtplx_stats: quiet-envelope rule — present when a stop named its
+    # commit path, absent on length finishes (None in to_dict()).
+    assert (
+        _public_mtplx_stats({"stats": stats})["finish_stop_origin"]
+        == "residual_correction"
+    )
+    # Request-log JSONL row: nullable like repetition_stop_reason.
+    envelope = _metrics_envelope(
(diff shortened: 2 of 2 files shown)


youssofal:
Merged for 2.11.1 with your authorship (commit ad366c45): `finish_stop_origin` reaches `mtplx_stats` and the request log. Thank you, and thank you for the reading on #414 and #435. Release: https://github.com/youssofal/MTPLX/releases/tag/v2.11.1

atirna:
thx, and glad the #414 reading was useful. nice to see it in 2.11.1.

PR #433, opened 2026-09-02
fix(openai): accept OpenAI's off-ish reasoning_effort values

sypsyp97:
## Summary

A client that hardcodes OpenAI's `reasoning_effort: "minimal"` or `"none"` gets a 400 out of the chat request path. Cherry Studio's translate pane sends one on every call and exposes no effort picker, so the entire pane is unusable against MTPLX — reproduced against 2.11.0:

```
$ curl -s localhost:8080/v1/chat/completions -H 'Content-Type: application/json' \
    -d '{"model":"...","messages":[{"role":"user","content":"hi"}],"reasoning_effort":"none"}'
{"message": "reasoning_effort must be one of: auto, low, medium, high, xhigh",
 "type": "invalid_request_error", ...}
```

This maps `minimal`/`none`/`off`/`disable`/`disabled` onto the lowest level in `normalize_reasoning_effort`, the one place the request path and the live-settings POST share.

None of those values can mean "no reasoning" on this server — that is `reasoning: off` or `chat_template_kwargs.enable_thinking` — and `reasoning_effort` can only answer with a level, so the lowest one is the closest available answer rather than a rejected request. Deliberately **not** wired to reasoning-off, because a client asking for an effort level should not have thinking silently disabled underneath it.

Scope is narrow by construction:

- `REASONING_EFFORT_CHOICES` is untouched, so every family's `effort_levels` contract and the app's picker are unchanged.
- Junk values still raise (`"banana"` → 400).
- The aliases only widen what arrives over the wire: `--reasoning-effort` is gated by argparse `choices=REASONING_EFFORT_CHOICES` and `mtplx config set` compares against the same tuple, so an alias cannot reach either surface.

I left CHANGELOG.md alone since entries there look maintainer-written at release time; happy to add one in whatever wording you prefer.

## Verification

macOS 15.6 (arm64), Python 3.12.13, this branch installed with `-e ".[dev,server]"`.

```
python -m pytest tests/test_no_mlx_imports.py tests/test_public_cli.py \
                 tests/test_runtime_kpis.py tests/test_qwen38_family.py \
                 tests/test_openai_bridge.py
453 passed

python -m build              -> mtplx-2.10.2-py3-none-any.whl
scripts/fresh_venv_smoke.sh  -> passed no-MLX CLI survival checks
scripts/check_ai_attribution.py --range upstream/main..HEAD -> clean
```

New test `test_normalize_reasoning_effort_maps_openai_off_values` covers the five aliases plus case and whitespace variants, asserts every canonical choice still round-trips, and asserts junk still raises.

End to end on a live server (Qwen3.8-27B-UD-Q4_K_XL, one M5 Pro, `--profile turbo --reasoning on --reasoning-effort xhigh`): `minimal`, `none`, `off`, `disable` all return 200 after the change and `banana` still returns 400.

## Benchmark Evidence

No runtime performance change: the added lookup is one dict `get` on the request-validation path, before generation.

### DIFF
--- a/mtplx/reasoning_effort.py
+++ b/mtplx/reasoning_effort.py
@@ -12,6 +12,11 @@ level. That is how 2.7.0 shipped with Qwen 3.8's ``xhigh`` bouncing back to
 ``medium``: the request path knew the level, the live-settings coercer did
 not, and because that POST is all-or-nothing the app's whole settings payload
 was rejected.
+
+``_OPENAI_EFFORT_ALIASES`` widens only what arrives over the wire. The CLI
+gates ``--reasoning-effort`` on ``REASONING_EFFORT_CHOICES`` through argparse
+``choices`` and ``mtplx config set`` compares against the same tuple, so an
+alias never reaches either; the canonical vocabulary is unchanged.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


youssofal:
Merged for 2.11.1 with your authorship (commit f47199f5): the off-ish `reasoning_effort` values are accepted. Thank you. Release: https://github.com/youssofal/MTPLX/releases/tag/v2.11.1

sypsyp97:
Thanks for the merge!

PR #442, opened 2026-09-03
forge: extract MTP heads that are not stored under an mtp. prefix

stooit:
## Summary

is_mtp_key matches only "mtp." and "language_model.mtp.". Heads stored any other way extract zero keys, so forge writes no sidecar -- while _inspection_has_mtp_weight_evidence still reports a head is present.

Three layouts exist:

    mtp.* / language_model.mtp.*   Qwen, Gemma                handled
    model.layers.{N}.*             GLM-4 MoE, DeepSeek-V3.2   extracted nothing
    model.mtp_layers.{i}.*         MiMo                       extracted nothing

Prefix is probed first, and the order matters: a model can declare `num_nextn_predict_layers=1` and still keep its head under a prefix, so an appended-first probe would look for a layer that does not exist.

Taking MiMo through to a working pack found three more blockers. Each one stopped forge after a successful convert, extract and calibrate:

1. lm_head was never bound. _MiMOMTP owns an lm_head that only the trunk-shard branch fills. A forged artifact keeps the head in mtp.safetensors, and that branch returns the sidecar alone, so lm_head stayed random and load_weights(strict=False) bound the rest silently. Calibration then scored 0.0 on all 64 candidates and blamed the sidecar, which was complete and mapped 16/16.

2. model_family_from_inspection had no mimo branch, so MiMo resolved to "unknown" and tune refused it -- though mimo-mtp ships a native backend with can_run_verified=True.

3. forge asked for a fixed --depths 1,2,3 and required rows 0,1,2,3. MiMo drafts one token, so tune failed with "tune depths must be one of 1". Depths now come from the tune policy. Non-D candidates (Gemma 4 blocks) and lookup failures fall back to 1,2,3, so no other backend changes.

Scope. MiMo is the layout carried end to end. Appended-layer extraction is implemented and tested, but no appended-layer model completes forge build here: tune has no family branch for glm or deepseek, so those still stop at the family gate. That gate is a separate change.

## Verification

Key counts from each published model.safetensors.index.json. All extract 0 keys today:

    DeepSeek-V3.1 / R1 / V3-0324    appended    1564 each
    DeepSeek-V3.2-Exp               appended    1571
    GLM-4.5 / 4.6 / 4.5-Air         appended    502 / 500 / 404
    GLM-4.7-Flash                   appended    212
    MiMo-7B-RL / SFT                namespace   16 each

MiMo-7B-RL end to end: forge build exit 0, artifact branded, 16/16 tensors mapped, calibration exact_agreement at 1.0 for depths 1-3.

Tests: 13 added, all passing. The suite has pre-existing failures on this machine -- Metal kernels, the mlx pin floor, and numerical tolerance -- and the same set fails on unmodified main, so this branch adds none. The extraction tests use real layer indices and key counts. The key case is cross-claiming: MiMo's trunk has 36 layers, so model.layers.36. would be its appended form. The head is not there, and neither matcher may take the other's keys.

## Benchmark Evidence

    Date          2026-09-02
    Commit        d6677b9, on upstream main e652d55
    Hardware      Apple M3 Max (Mac15,9), 12P+4E, 128 GiB, macOS 26.6.2 (25G83)
    Fan mode      fans pinned
    Profile       sustained
    Model         XiaomiMiMo/MiMo-7B-RL, forged locally
    Quantization  body 4-bit affine, group_size 64; MTP sidecar bf16
    Sampler       temperature 0.6, top_p 0.95, top_k 20, seed 0, thinking off
    Runtime       mlx 0.32.0, mlx-lm 0.31.3, MTPLX 2.10.2
    Contract      concat_order hidden_embedding, hidden_variant post_norm,
                  base_hidden_variant post_norm, position local,
                  cache persistent, history committed

forge build, long-code-uncapped, 2048 tokens:

    AR    69.69 tok/s
    D1    91.38 tok/s    1.311x    acceptance 69.2%    mtp_depth_wins

Both rows passed the quality gate.

mtplx tune, cold-long-code-192, 256 tokens, limit 8, three runs:

    run   AR tok/s   D1 tok/s   ratio   acceptance
    1        72.58      96.29   1.327        68.4%
    2        72.69      96.29   1.325        68.4%
    3        72.61      96.27   1.326        68.4%

AR baselines span 0.15%. mlx-lm 0.31.3 is the pinned version, so these reproduce as-is.

Not benchmarked:

- Appended-layer models (DeepSeek, GLM) are 100B+ and would not fit on the hardware available. Their key routing is verified against real weight indices and in tests, not by a forged pack.
- The tied-lm_head branch never runs for MiMo-7B-RL, which is untied.

### DIFF
--- a/mtplx/artifacts.py
+++ b/mtplx/artifacts.py
@@ -151,6 +151,53 @@ def _num_mtp_layers(config: dict[str, Any]) -> int:
     )
 
 
+def appended_mtp_layer_range(config: dict[str, Any]) -> range:
+    """Layer indices holding an appended-layer MTP head.
+
+    GLM MoE checkpoints ship the MTP head as extra decoder layers starting
+    at ``num_hidden_layers`` (``model.layers.47.*`` for GLM-4.7-Flash),
+    rather than under an ``mtp.`` prefix.  Returns an empty range when the
+    config does not describe that layout.
+    """
+    tcfg = text_config(config)
+    start = int(tcfg.get("num_hidden_layers") or config.get("num_hidden_layers") or 0)
+    count = _num_mtp_layers(config)
+    if start <= 0 or count <= 0:
+        return range(0)
+    return range(start, start + count)
+
+
+def is_mtp_layers_namespace_key(key: str, config: dict[str, Any]) -> bool:
+    """Match an MTP head kept in its own ``model.mtp_layers.N.`` namespace.
+
+    MiMo stores the head neither under an ``mtp.`` prefix nor as an appended
+    decoder layer, but in a separate namespace beside ``model.layers.*``.
+    ``mimo_mtp_patch`` already reads that form, so extraction only has to
+    select the keys; no rewrite is needed.
+    """
+    text = str(key)
+    count = _num_mtp_layers(config)
+    return count > 0 and any(
+        text.startswith(f"model.mtp_layers.{index}.") for index in range(count)
+    )
+
+
+def uses_mtp_layers_namespace(config: dict[str, Any]) -> bool:
+    return _num_mtp_layers(config) > 0
+
+
+def uses_appended_layer_mtp(config: dict[str, Any]) -> bool:
+    return len(appended_mtp_layer_range(config)) > 0
+
+
+def is_appended_layer_mtp_key(key: str, config: dict[str, Any]) -> bool:
+    text = str(key)
+    return any(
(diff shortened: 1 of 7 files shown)


youssofal:
Landed in 2.11.2 under your name: 3d2f74da, 14a190dd, ec1c00e6, 044d2afd and 3056ad1d, merged as 45d45aeb, with #457 alongside it. Thank you, Stuart.

PR #444, opened 2026-09-03
Fix SSD prefix restores hydrating full long snapshots

softpudding:
## Summary

- Decode only the KV blocks required through the selected prefix/GDN boundary for SSD cold prefix restores.
- Slice committed MTP history to the boundary-aligned `N - 1` prefix instead of hydrating and immediately trimming its full tail.
- Preserve the legacy full-decode path for tiny hybrid gaps without a boundary and cache layouts with coupled/ring metadata.
- Carry the materialized snapshot spans into SessionBank so restore does not trim data that was deliberately never read.

## Validation

- Added codec and end-to-end cold-tier regressions covering partial KV, GDN boundary state, committed MTP history, and boundary metadata restoration.
- `63 passed, 1 deselected` across the related cache-bank/cold-tier test set; syntax and F/E9 static checks pass.
- Replayed a persisted 74,308-token Qwen entry locally at its 2,048-token GDN boundary: 2,281 blobs / 0.277 GiB were decoded in 0.314 s, instead of hydrating the multi-GiB entry.

### DIFF
--- a/mtplx/cache_bank/codec.py
+++ b/mtplx/cache_bank/codec.py
@@ -59,6 +59,12 @@ class DecodedPayload:
     # v3: (token_count, recurrent-only CacheSnapshot, hidden_last|None)
     gdn_boundaries: tuple = ()
     has_recurrent: bool = False
+    # When a cold prefix restore decodes only the attention-KV blocks that
+    # precede a safe boundary, these fields describe the materialized prefix
+    # lengths.  The SessionBank keeps the entry's original token identity;
+    # these are strictly the lengths represented by the decoded snapshots.
+    cache_snapshot_prefix_len: int | None = None
+    mtp_history_snapshot_prefix_len: int | None = None
 
 
 class ColdEncodeInterrupted(RuntimeError):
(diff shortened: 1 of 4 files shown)


youssofal:
The direction is right: 0.277 GiB decoded instead of a multi-GiB entry at a 2,048-token boundary. On current main it merges without textual conflicts, but two session-bank invariants fail on the merged tree: `test_session_bank_live_reference_can_restore_block_prefix_boundary` (MTP offset 1022, expected 1023) and `test_session_bank_near_prefix_trims_mtp_history_by_gap_not_absolute_offset` (trimmed [2], expected [1]). The committed-history slice lands one row short of the gap-based trim that shipped in 2.10.2. 2.11.1 is on main now; please rebase onto it and make those two pass, and I will merge after that.

softpudding:
Rebased onto current main at dd4031a (v2.11.1) and fixed the MTP off-by-one by keeping legacy/live snapshots on gap-based trimming while using the recorded physical span only for prefix-decoded SSD snapshots.

Verified:
- tests/test_session_bank.py::test_session_bank_live_reference_can_restore_block_prefix_boundary
- tests/test_session_bank.py::test_session_bank_near_prefix_trims_mtp_history_by_gap_not_absolute_offset
- full tests/test_session_bank.py + tests/test_cache_bank.py: 45 passed
- Ruff on the touched Python files

PR #452, opened 2026-09-04
Add optional aria2c download acceleration

### LINKED ISSUE #451: Add optional aria2c backend for resumable Hugging Face model downloads
Large multi-shard MTPLX model downloads currently stream one file at a time through requests. Add an optional aria2c backend for parallel segmented transfers. Requirements: preserve the existing .incomplete resume behavior; pin every file to the resolved revision; retain size and SHA-256 verification before atomic rename; keep structured progress events working for the CLI and macOS app; never expose Hugging Face tokens in process arguments; and fall back to the current downloader when aria2c is unavailable. Include focused tests for backend selection, resume, authentication handling, progress, failure cleanup, and fallback.

zeeshanhaque21:
## Summary

- add an optional aria2c backend for parallel Hugging Face model pulls
- resume existing .incomplete files with 16 segmented connections and four concurrent files
- report accurate structured progress through aria2 JSON-RPC instead of sparse-file size
- verify Hugging Face SHA-256 checksums before atomic rename
- keep Hugging Face credentials off process arguments and fall back to Python when aria2c is unavailable

## Usage

Install aria2 with Homebrew, then use the default auto selection or pass --download-backend aria2 to require it. Pass --download-backend python to retain the built-in path.

## Verification

- repository CI test set plus aria2, Hugging Face loader, and artifact suites
- real loopback aria2 resume from a 512 KiB partial with final SHA-256 verification
- Python bytecode compilation

Closes #451

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,5 +18,5 @@ jobs:
           python-version: "3.11"
       - run: python -m pip install -U pip
       - run: python -m pip install -e ".[dev,server]"
-      - run: python -m pytest tests/test_no_mlx_imports.py tests/test_public_cli.py tests/test_runtime_kpis.py tests/test_server_openai.py tests/test_openai_bridge.py tests/test_thermal.py tests/test_cache_state.py
+      - run: python -m pytest tests/test_no_mlx_imports.py tests/test_public_cli.py tests/test_runtime_kpis.py tests/test_server_openai.py tests/test_openai_bridge.py tests/test_thermal.py tests/test_cache_state.py tests/test_aria2_downloader.py
       - run: python -m compileall -q mtplx tests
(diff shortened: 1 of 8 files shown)


youssofal:
Thanks, the design is clean: credentials over stdin, SHA-256 before the atomic rename, and a Python fallback. One thing before it lands: `--download-backend auto` selects aria2c whenever it is installed, which changes the download path for anyone who already has aria2 on their Mac. I would ship the first release with `python` as the default and `aria2` as the opt-in, and flip the default once a few pulls have been measured. It is not in 2.11.2. It is planned for the next release after a real pull here, under your name.

youssofal:
Merged for 2.11.3 as commit 2fd60025 with your authorship kept, thank you. aria2 is an opt-in backend and the built-in downloader stays the default (commit 9b594bec). Closing here because the merge went in by patch rather than through this button. Release: https://github.com/youssofal/MTPLX/releases/tag/v2.11.3

Issue #447, opened 2026-09-04
prefill_admission_shed refuses a warm +23K turn (507) while 12.6 GB of evictable bank is resident, and cache_cleared makes retries unrecoverable


nomishbhardwaj:
Split out from #432 as requested there. Same config as the sibling issue: MTPLX 2.11.1 (brew), M5 Max 128 GB, Optimized-Speed pack, depth 3, draft temp 0.3, turbo, window 262144, temp 0.3 / top-p 0.9 / top-k 40, reasoning high, SSD session cache on, dense-decode + marathon-protect envs.

First, the good news for the record: **the pressure-deferral wedge I reported on #432 does not reproduce on 2.11.1.** The canonical repro (second deep 5×23K session started while the first session's ~7 GB RAM bank is resident) now runs clean — walls 21–80 s, one graceful `pressure_trim` bank eviction, no deferred-loop hang, no client timeout. Whatever changed in this release, that scenario is fixed.

What remains is the admission guard's behavior when a request genuinely projects over the threshold: **it fails the request while holding evictable bank, and clears the session's own reusable prefix on the way out, which makes every retry strictly worse.**

**Receipt 1 — 2.11.1, warm turn refused while 12.6 GB of bank sits resident.** A 115K session (the #432-fix scenario: deep turns, one cross-session vision call — the new marathon protect worked exactly as described, the postcommit survived and stored). The next incremental turn:

```
prefill_admission_shed prompt_tokens=137966 reusable_prefix_tokens=114880
  reusable_prefix_mode=exact miss_tokens=23086
  active_bytes=99724543246 projected_bytes=107024774967 threshold_bytes=99986838650
  bank_bytes_before=12575677808 lru_entries_evicted=0 bank_bytes_after=12575677808
  cache_cleared=true
→ HTTP 507, then allocation_failure_shed
```

The request needed ~7 GB over threshold; 12.6 GB of bank was resident and none of it was evicted (`lru_entries_evicted: 0`) before the 507. And `cache_cleared: true` threw away the exact 114,880-token prefix that admission had just measured — so the retry is a 138K full-cold miss that can never be admitted on this machine.

[...]

PR #453, opened 2026-09-04
server+session bank: the admission shed asks the live sessions for reuse and walks the bank before letting a request die (#447)

nomishbhardwaj:
## Summary

Fixes the two admission-shed defects from #447, each behind an explicit-export kill switch and with its receipt extended:

1. **The reuse estimate now asks the live sessions, not just the bank** (`MTPLX_PREFILL_ADMISSION_LIVE_PREFIX`). The engine's live sessions serve a committed prefix directly — that is what `cached_tokens` reads on every warm turn — and a live frontier can be unbanked (a refused snapshot on retokenized-history mismatch banks nothing while the live KV still serves). Estimating from the bank alone read a warm 212k session as a full miss, cleared its snapshots as "superseded", and every retry was then a 211,807-token cold miss until a server restart. The shed now walks the same ladder session resolution walks (`longest_prefix_session` → `pending_near_prefix_session` → `best_common_prefix_session`, tolerant answers rewound to the block floor the bank estimate already uses) and reports `reusable_prefix_mode: "live_session"`.

2. **When the projection crosses the line, the shed escalates through the bank instead of letting the request die** (`MTPLX_PREFILL_ADMISSION_CHAIN_SHED`). New `SessionBank.shrink_for_admission`: phase 1 walks non-terminal chain entries (every session keeps its highest-prefix entry — the one the protected-terminal order guards); phase 2, only if the deficit stands, takes remaining entries in the take-anything order real memory pressure uses (active sessions last). Both phases spare the entry the imminent prompt restores from and never touch a live-referenced entry (the same bar `_supersede_contained_prefixes` sets). Evictions are RAM-only — the SSD cold tier still restores a walked entry, so the worst case is a disk read on some session's next turn, not this request's abort. `shrink_to_bytes` and its dynamic-ceiling / macOS-pressure callers are unchanged. The receipt gains `chain_entries_evicted` and `terminal_entries_evicted`.

## Verification

- `tests/test_prefill_admission_shed.py`: all pre-existing tests unchanged and passing; new cases for the live ladder (inert under the miss floor, pin-instead-of-clear, exact beats tolerant rungs, exceptions never cost the request, export yields), and for the escalation (siblings walked when LRU is protected, terminals only in phase 2, export yields, banks without the method tolerated).
- New `tests/test_session_bank_chain_shed.py`: phase ordering, protect_tokens shielding, live-ref entries never victims, active-last ordering in phase 2, LRU among siblings, reasons recorded.
- Full run: the admission + session-bank suites (incl. `test_session_bank_restore_aliasing.py`, `test_session_bank_snapshot_settle.py`, `test_session_bank_protected_terminal.py`), plus the CONTRIBUTING trio (`test_no_mlx_imports.py`, `test_public_cli.py`, `test_runtime_kpis.py`), `python -m build`, and `scripts/fresh_venv_smoke.sh` — all green.
- `scripts/agent_session_gate.py --context-tokens 40000` against a serving daemon: `pass: true, failures: []` on this branch (and on stock 2.11.1 for the baseline).

**Incident replay (the #447 receipt-1 scenario)** — 115k incremental session, cross-session vision call 0.5 s after the deep turn, then the next warm turn, M5 Max 128 GB, Optimized-Speed pack, SSD session cache on:

- stock 2.11.1: next turn **HTTP 507** (`reusable exact 114,880, miss 23,086, projected 107.0 G > threshold 100.0 G, bank 12.6 GiB resident, lru_entries_evicted: 0, cache_cleared`), then `allocation_failure_shed`.
- this branch: next turn **completes in 33 s, `cached=114,941`** — receipt: `reusable_prefix_mode: exact 114,941, chain_entries_evicted: 0, terminal_entries_evicted: 2, bank 8.35 → 4.35 GiB`. Reproduced twice (33.0 s / 33.5 s).

**Estimate A/B (receipt-2 shape)** — incremental session with snapshots too large to bank (`MTPLX_SESSION_BANK_MAX_BYTES=256MiB`) under a reduced `MTPLX_MEMORY_LIMIT_BYTES`: on the same turn, stock's shed read `reusable_prefix_tokens: 0, mode: none` while the completed turn itself reported `cached=37,963`; this branch read `mode: live_session, reusable 38,188`.

## Benchmark Evidence

No decode or prefill path is touched; the guard runs pre-prefill only, and with no shed firing the code is inert (verified: zero guard events in the parity runs below).

- Hardware: MacBook Pro M5 Max, 128 GB. Model: `Youssofal--Qwen3.8-Flash-Next-MTPLX-Optimized-Speed` (trunk 4-bit dynamic + 8-bit attention). Sampler: temp 0.3 / top-p 0.9 / top-k 40, reasoning high, MTP depth 3, draft temp 0.3, turbo profile, context window 262144, 700-token code answers. Fans: system default (no fan control). Date: 2026-09-04. Baseline: brew 2.11.1 (`dd4031a`); branch: this PR.
- Short context (23k prompt, one round each, zero guard events both arms): decode 65.3 vs 67.1 tok/s median, prefill 883 vs 976 tok/s — within this workload's run-to-run spread.
- Deep context (5 × 23k incremental to ~115k, full-cache decode, 120 s settle before each arm, rotated order, 5 branch / 6 stock samples across three sessions): medians ~64 vs ~64 tok/s. Both arms occasionally show a single ~16 tok/s generation at ~115k during an allocator-deferral window (observed on stock and branch alike; pre-existing, unrelated to this change).

Harness scripts and full serve logs available on request.

### DIFF
--- a/mtplx/server/openai.py
+++ b/mtplx/server/openai.py
@@ -17435,6 +17435,18 @@ def _prefill_admission_min_miss_tokens() -> int:
         return 4096
 
 
+def _prefill_admission_live_prefix_enabled() -> bool:
+    return os.environ.get(
+        "MTPLX_PREFILL_ADMISSION_LIVE_PREFIX", "1"
+    ).strip().lower() not in {"0", "off", "false", "no"}
+
+
+def _prefill_admission_chain_shed_enabled() -> bool:
+    return os.environ.get(
+        "MTPLX_PREFILL_ADMISSION_CHAIN_SHED", "1"
+    ).strip().lower() not in {"0", "off", "false", "no"}
+
+
 # Same line as _allocator_pressure_level's WARNING edge: at >=97% of the
 # Metal limit the next growth step swaps.
 _PREFILL_ADMISSION_PRESSURE_FRACTION = 0.97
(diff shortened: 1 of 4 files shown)


youssofal:
Integrated through #458, now merged into main at 406b5f768e984e036d16aca1edaddaa29fe8519e. The live-prefix lookup and protected session-bank reclamation contributions retain nomishbhardwaj’s human authorship, with an additional allocator-first step to preserve reusable snapshots whenever releasing unused buffers is sufficient. The recovery completed the 206K session that failed admission at 191K on stock 2.11.1, with unchanged memory limits. Closing this duplicate integration path; thank you for the underlying fix.

PR #471, opened 2026-09-06
Allow removing stale custom models from picker

nRanzo:
## Summary
* Allow stale persisted custom models to be removed from the model picker.
* Reuse the existing `removeCustomModel(id:)` configuration API.
* Keep installed, official, and currently selected models non-removable.
* Removing a model only unregisters it from the picker; model files on disk are left untouched.
* Add focused tests for removal behavior and picker eligibility.

## Removal eligibility
* Only persisted custom models that are no longer installed can be removed.
* Official catalog models are never removable.
* Installed custom models are not removable.
* The currently selected model is not removable.
* Removal only unregisters the model from the picker and does not delete files from disk.

## Testing
* Added unit tests covering stale, installed, official, and current models.
* Manually verified the model picker in the macOS app: a stale custom model exposes `Remove from picker` via the row context menu and requires confirmation before removal.
* `git diff --check` passes.

### DIFF
--- a/apps/MTPLXApp/Sources/MTPLXAppCore/Models/AppConfiguration.swift
+++ b/apps/MTPLXApp/Sources/MTPLXAppCore/Models/AppConfiguration.swift
@@ -556,6 +556,13 @@ public struct MTPLXAppConfiguration: Codable, Equatable, Sendable {
         customModels.append(option)
     }
 
+    @discardableResult
+    public mutating func removeCustomModel(id: String) -> Bool {
+        let originalCount = customModels.count
+        customModels.removeAll { $0.id == id }
+        return customModels.count != originalCount
+    }
+
     /// Persist a locally-forged model into the picker. Called by the
     /// Forge wizard's Registered stage when the build completes;
     /// dedup is by id (same branded name re-forged) AND by local
--- a/apps/MTPLXApp/Sources/MTPLXAppHost/Views/Forge/ForgeMineView.swift
+++ b/apps/MTPLXApp/Sources/MTPLXAppHost/Views/Forge/ForgeMineView.swift
@@ -323,7 +323,12 @@ struct ForgeMineView: View {
     private func removeFromPicker(entryID: String) {
         guard let entry = entries.first(where: { $0.id == entryID }) else { return }
         var config = backend.configuration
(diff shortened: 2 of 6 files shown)


youssofal:
Thanks, this reads like a finished change. A stale persisted custom model should be removable from the picker without touching files on disk, and the rules you list (official, installed and currently selected models stay put) match how the picker is meant to work. If you have the branch, open a pull request against main and we will review it for the next release. The model picker overlay changed in 2.11.2, so rebase on the tag once it is out.

nRanzo:
Rebased #471 onto v2.11.2 and re-tested the picker behavior. The stale custom model can still be removed as expected, with the confirmation flow working correctly.

youssofal:
Reviewed the rebased branch (3d6a436a) on top of v2.11.2 and the work for the next release: it merges clean, and the app suite passes with it (897 tests, 0 failures), including the eight tests you added and the localization table checks. The four strings the dialog and the context menu use already exist in every language table. The rule (persisted custom, not installed, not selected) and the files-untouched guarantee are exactly right. It is queued for the next release and will be merged under your authorship with it. Thank you for the rebase and the retest.

jsshapiro:
Does this resolve the issue where using the CLI to remove a model does not remove it from the app menu? I've had a couple of models from HF that either failed to download or crash immediately. Removing them is easy enough, but it would be nice to get them out of the menu.

I can open a new bug, but thought I'd ask if this resolved it first.

nRanzo:
> Does this resolve the issue where using the CLI to remove a model does not remove it from the app menu? I've had a couple of models from HF that either failed to download or crash immediately. Removing them is easy enough, but it would be nice to get them out of the menu.
> 
> I can open a new bug, but thought I'd ask if this resolved it first.

Kind of, this PR only adds a way to remove stale persisted custom models from the app menu (right-click, remove, and confirm). If the CLI removal doesn’t update the app menu automatically, that’s a separate issue and a new bug would make sense.

youssofal:
Merged for 2.11.3, your commit 3d6a436a with its authorship kept (merge a9932780), thank you. Release: https://github.com/youssofal/MTPLX/releases/tag/v2.11.3

PR #474, opened 2026-09-06
Fix Flash-Next FP8 n-gram scalar scales

GrahamJenkins:
## Summary

The stock converter fails on the official Qwen/Qwen3.8-Flash-Next-FP8 revision 236dfdf285828023ca3bcd3f37366c58a3469b13 because its 128 PLE n-gram FP8 shards share one BF16 scalar weight_scale rather than providing per-weight weight_scale_inv tensors.

This narrow fix falls back to that shared scalar only for n-gram shards, broadcasts it during FP8 dequantization, and accounts for the shared source tensor in fail-closed accounting. Ordinary block-scale weight_scale_inv conversion remains unchanged.

## Validation

- Added synthetic regression coverage for both ordinary 128x128 block weight_scale_inv and official n-gram shared scalar weight_scale layouts.
- Targeted regression tests pass on Linux with a NumPy-backed MLX stand-in; native MLX execution is not available on this host.
- Completed the full 185.6 GB official-source conversion, Forge verification, and model-load verification on Nexus. The verified patched converter SHA-256 is 4ecc6eb37366739ba214778adea7ba6aeacf133817d5162cba684abc8f1a8b73.

### DIFF
--- a/scripts/convert_qwen4_exp_fp8.py
+++ b/scripts/convert_qwen4_exp_fp8.py
@@ -163,11 +163,15 @@ class SourceReader:
         if dtype == "F8_E4M3":
             w = mx.array(fp8_e4m3_lut()[np.ascontiguousarray(arr)])
             scale_name = name.replace(".weight", ".weight_scale_inv")
+            if scale_name not in self.weight_map and ".ngram_embedding.shard_" in name:
+                scale_name = name.split(".ngram_embedding.shard_", 1)[0] + ".ngram_embedding.weight_scale"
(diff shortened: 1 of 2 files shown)


youssofal:
Merged for 2.11.3, your commit 136ca0dd with its authorship kept, thank you. The FP8 n-gram scalar scales are read correctly and the converter accepts the official FP8 checkpoint. Closing here because the merge went in by patch rather than through this button. Release: https://github.com/youssofal/MTPLX/releases/tag/v2.11.3

PR #391, opened 2026-08-29
perf(qwen4): restack Flash-Next optimizations on MTPLX 2.10.1

davidtai:
# Qwen3.8 Flash-Next serving stack

This pull request adds the Qwen3.8 Flash-Next serving stack to MTPLX: 24 runtime optimizations, the server code that arms them, and the tests. The stack decodes 1.40 times as fast as upstream 2.10.2 at 16,384 tokens of context.

Headline figures at 16,384 tokens of context, 3 seeds, cold prefill:

- Decode: 57.65 tok/s on upstream 2.10.2, **80.92 tok/s** on this branch.
- Prefill: 1,191 tok/s on upstream 2.10.2, **1,332 tok/s** on this branch.
- HumanEval pass@1: 0.933 with the optimizations off, **0.951** with them on.

Terms used in this document:

- MTP: multi-token prediction.
- TTFT: time to first token, measured by the client.
- KV: the key and value tensors of attention.
- GQA: grouped-query attention.
- QSA: Qwen Sparse Attention.
- GDN: Gated DeltaNet.
- PLE: per-layer embedding, fed from an n-gram sidecar table.
- FR-Spec: the frequency-ranked draft head over a pruned vocabulary.
- tok/s: tokens per second.

---

## 1. Benchmark results

Three engines ran in one benchmark run on one machine. Every engine received the same request body per cell. Section 2 gives the settings and the caveats.

### 1.1 Summary

Summary of decode, prefill, TTFT and peak memory by context size, one line per engine

The grid shows decode tok/s, prefill tok/s, TTFT on a logarithmic scale, and peak memory. The branch leads upstream 2.10.2 on decode at every context size and on TTFT from 16,384 tokens upward.

### 1.2 Decode tok/s

Decode tok/s by context size, one line per engine

| Cell | upstream 2.10.2 | branch | Δ% | mlx-serve 26.8.11 | Δ% |
| --- | ---: | ---: | ---: | ---: | ---: |
| vanity | 82.25 (79.55-85.82) _n=3_ | 100.49 (95.78-104.11) _n=3_ | +22.2% | 94.13 (89.85-99.88) _n=3_ | +14.4% |
| 1K | 67.95 (66.20-71.24) _n=3_ | 86.13 (82.27-93.65) _n=3_ | +26.8% | 67.43 (62.80-74.75) _n=3_ | -0.8% |
[...]

PR #475, opened 2026-09-06
[FEAT] 4 unimplemented optimizations from PR391 plus 2 new ones - six serving optimizations for Qwen3.8 Flash-Next on MTPLX 2.11.2 (four PR-391 remainder optimizations + two exact decode optimizations)

davidtai:
# Six serving optimizations for Qwen3.8 Flash-Next: the PR 391 remainder and two exact decode optimizations

> Interim update: the context ladders, the 16,384-token interleave, and the charts are final. Quality columns and the typical-threshold sweep fill in as those runs complete; cells that read "measuring" are still pending.

This pull request adds six serving optimizations to the Qwen3.8 Flash-Next path. The base is upstream MTPLX 2.11.2. Four are the PR 391 optimizations that 2.11.2 did not re-land: a hyper-connection read kernel, a prefill causal-mask fuse, a QSA prefill query tile, and a QSA split-K sparse decode kernel. Two are exact decode optimizations: cached async PLE rows and pooled QSA row selection. The four remainder optimizations are rounding-class, so the output differs from the stock path only by floating-point rounding. The two aux optimizations are exact, so their output is byte-for-byte the base. The base commits also re-engage four upstream verify optimizations that 2.11.2 shipped but left silently off when served (Section 3).

The comparison has two arms: release 2.11.2 alone (A) and release plus this pull request's six optimizations (C, this pull request). The headline is C against A.

Headline figures at 16,384 tokens of context, 3 seeds, cold prefill:

- Decode: 71.97 tok/s on release 2.11.2, **82.84 tok/s** with all six optimizations, +15.1%.
- Prefill: 1381.5 tok/s on release, **1397.9 tok/s** with the optimizations, +1.2%.
- Long context: both release and this pull request exceed the 100 GiB memory knob at 261,120 tokens on this machine; the prefill completes and the overflow is the speculative verify's KV-cache write, which #482 fixes so the cell decodes at a 100.82 GB peak (Section 1.5).
- Quality at the benchmark's own sampler (temperature 1, `xhigh`, one seed): HumanEval strict pass@1 0.9695 on release against 0.9634 with the six optimizations, completed-task 1.0000 against 0.9814; MBPP at that sampler ran for one arm only (Section 4).

Terms used in this document:

- decode: the token-by-token generation phase, after the prompt is read.
- prefill: reading the prompt into the KV cache before the first output token.
- optimization: one change an operator can switch off on its own.
- PLE: a per-layer embedding, a small extra lookup the model adds inside each layer.
- QSA: the sparse-attention path, which selects the earlier tokens each layer attends to.
- HC: the hyper-connection read inside the verify step.
- M4: the fixed four-row speculative verify width.
- rounding-class: the output differs from the stock path only by floating-point rounding.
- tok/s: tokens per second.

---

## 1. Benchmark results

Two arms ran in one benchmark run on one machine. Each arm received the same request body per cell. Section 2 gives the settings.

### 1.1 Summary

summary
*Summary grid: decode, prefill, time-to-first-token and peak memory by context size, one line per arm, each point the fastest of its seeds. Bars show the slowest-to-fastest spread.*

### 1.2 Decode tok/s

decode_tok_s

| Context | A release | C (this PR) | C vs A |
| --- | ---: | ---: | ---: |
| 1,024 | 87.32 (73.66-87.32) _n=3_ | 88.62 (85.88-88.62) _n=3_ | +1.5% |
| 8,192 | 72.29 (62.44-72.29) _n=3_ | 85.95 (75.93-85.95) _n=3_ | +18.9% |
| 16,384 | 71.97 _n=9_ | 82.84 _n=9_ | +15.1% |
| 32,768 | 76.47 (69.65-76.47) _n=3_ | 81.30 (71.72-81.30) _n=3_ | +6.3% |
| 65,536 | 69.26 (65.68-69.26) _n=3_ | 84.43 (70.82-84.43) _n=3_ | +21.9% |
| 131,072 | 64.26 (59.70-64.26) _n=3_ | 68.62 (65.67-68.62) _n=3_ | +6.8% |

The 261,120-token cell is in Section 1.5; both arms exceed the memory knob there.

### 1.3 Prefill tok/s

prefill_tok_s

The two exact decode optimizations do not change prefill; the candidate column is arm C prefill.

| Context | A release | C (this PR) | Delta |
| --- | ---: | ---: | ---: |
| 1,024 | 916.1 _n=3_ | 912.2 _n=3_ | -0.4% |
| 8,192 | 1359.3 _n=3_ | 1361.0 _n=3_ | +0.1% |
| 16,384 | 1381.5 _n=9_ | 1397.9 _n=9_ | +1.2% |
| 32,768 | 1212.5 _n=3_ | 1260.3 _n=3_ | +3.9% |
| 65,536 | 1150.7 _n=3_ | 1177.2 _n=3_ | +2.3% |
| 131,072 | 1116.7 _n=3_ | 1131.4 _n=3_ | +1.3% |

### 1.4 Time-to-first-token and peak memory

ttft_s
*Time-to-first-token by context size.*

peak_memory_gb
*Peak memory by context size. The 100 GiB cap holds on every arm that reaches the cell.*

The per-context tables for time-to-first-token and peak memory are in the perf report under `docs/perf/`.

### 1.5 261,120-token context

Both release 2.11.2 and this pull request exceed the 100 GiB memory knob at 261,120 tokens on this machine under the served launch. Metal returns `insufficient_memory` at the same point on both, at a sampled peak of 100.82 GB on release and 100.79 GB on this pull request.

The prefill is not what overflows. Every arm's receipt shows the full prompt read and tokens already emitted before the failure: `new_prefill_tokens` 261120, time to first token 243.5 s on release and 241.9 s on this pull request, then `completion_tokens` 12 on release and 3 on this pull request, with `finish_reason` `error` and `[METAL] Command buffer execution failed: Insufficient Memory`. The overflow is in the first decode step.

That decode step is the fixed four-row speculative verify, and the allocation that overflows is its **KV-cache write**. The verify holds its KV in `TensorOffsetKVCache`, whose `update_and_fetch` writes the new rows with the functional `mx.slice_update`. That op returns a new array, so it reallocates the whole `[1, 2, capacity, 256]` bf16 buffer for keys and again for values, about 0.27 GB each at the 262K capacity, in every one of the 12 full-attention layers: about 6.4 GB inside one verify command buffer, on a steady state already near 100.8 GB against a 107.374 GB limit. The stock `KVCache` used by plain autoregressive decode writes in place and allocates nothing, which is why the same prompt fits with native MTP off. A direct probe of that cache at this geometry measures 6.42 GB resident and, once the write is in place, a 0.00 GB update transient.

A second, smaller term sits in the same step: with a query length of 3 to 6 rows and 12 query heads per key/value head, query length times 12 is past the 32-row limit of MLX's fused vector attention kernel, so MLX declines the fused call and the unfused route materializes a `[24, S, 261120]` score tensor per layer. **Removing only that term does not make the cell fit.** A head-chunked build engaged and still failed at the same 100.82 GB peak. The prefill mask-fuse optimization declines in exactly this band, query length 3 to 8, so it does not help either; the wide 256-row prefill chunks still fuse. The analysis trail is in `over100-reports/oom255k/report.md`.

The earlier #391 tree only looked like it fit. That fit was one cold seed; its other two seeds also ran out of memory at the same cell.

This pull request does not change 261,120-token behaviour; the fix is #482, which writes the verify KV in place and keeps the verify attention on the fused kernel. Two windows confirm the cause and the fix:

| Window | Configuration | Result at 261,120 | Peak (GB) |
| --- | --- | --- | ---: |
| W5 | native MTP off (plain autoregressive decode, stock in-place `KVCache`) | fits, 547 tokens, normal `stop` | 94.19 |
| F2 | #482, MTP on, verify running | fits on all 3 cold seeds, 461 / 1024 / 845 tokens | 100.82 |

With the fix the cell decodes at 62.84 tok/s fastest (56.82-62.84), prefill 1082.9 tok/s, and sits about 6.5 GB under the knob, while 16,384-token output stays byte-identical to release.

### 1.6 Interleaved 16,384-token protocol

The arms ran interleaved at the 16,384 / 1,024 cell in the order A C D E, three windows per arm (D and E are the #478 typical-acceptance arms). The paired delta compares each arm-C window against its neighbouring arm-A windows, which cancels the slow warm-up drift across the run.

decode_16k_windows

| Window | Arm | Decode tok/s (fastest seed) |
| --- | --- | ---: |
| a1 | A (release) | 71.31 |
| c1 | C (this PR) | 76.94 |
| a2 | A (release) | 71.86 |
| c2 | C (this PR) | 77.11 |
| a3 | A (release) | 71.57 |
| c3 | C (this PR) | 76.66 |

Release 71.97 tok/s and candidate 82.84 tok/s are each the fastest window. The fastest-window gap is **+15.1%**. The paired per-seed 95% interval on the mean delta is [+1.76, +8.89] tok/s. Drift bracket (slowest-to-fastest spread between the release windows) +0.26 tok/s.

### 1.7 Supplementary measurements

These two sets are supplementary, not headline. Both use the fastest-of-seeds rule.

Greedy code-eval on the release arm, kept for reference only. It is not the quality gate: greedy is temperature 0, which is not the setting anything here was benchmarked at.

| Suite (release, greedy) | Strict pass@1 | Completed-task pass@1 | Truncation (rate; mean/max tokens) |
| --- | ---: | ---: | ---: |
| HumanEval / HumanEval+ / MBPP | 0.9756 / 0.9512 / 0.9040 | 0.9756 / 0.9512 / 0.9082 | 0.0% / 0.0% / 0.5% |

Three per-optimization knock-outs at 16,384 tokens, on this pull request's tree, each one window of three seeds. The plan was ten leave-one-out windows; the run was cancelled after three, which had already finished and passed their gates. The other seven were not measured, and each of these three is a single window, so its per-seed range is wide: read the change as indicative, not a paired result. This pull request's fastest ABAB decode is 82.84 tok/s.

| Optimization off | Decode with it off (fastest seed, range) | Change vs this PR (82.84 fastest) | Note |
| --- | --- | ---: | --- |
| HC mixer read | 79.45 (71.90-79.45) | -3.39 tok/s (-4.1%) | removed lane read off (armed:false) |
| QSA sparse decode | 82.00 (70.46-82.00) | -0.84 tok/s (-1.0%) | removed lane dropped from the install report |
| Prefill mask fuse | 81.94 (72.15-81.94) | -0.90 tok/s (-1.1%) | removed lane logged no engaged line |

The remainder-only arm: release plus the four ported optimizations and the four re-engaged keys, without the two exact decode optimizations. The battery calls it arm B. It was stopped after 32K, so it is not a headline. Decode is the fastest of its seeds with the slowest-to-fastest range.

| Context | Remainder-only decode (fastest, range) | vs release (A) | vs this PR (C) |
| --- | --- | ---: | ---: |
| 1K | 87.86 (83.75-87.86) _n=3_ | +0.6% | -0.9% |
| 8K | 84.71 (75.03-84.71) _n=3_ | +17.2% | -1.4% |
| 32K | 79.92 (70.29-79.92) _n=3_ | +4.5% | -1.7% |

---

## 2. Benchmark method

| Setting | Value |
| --- | --- |
| Model | `Youssofal/Qwen3.8-Flash-Next-MTPLX-Optimized-Speed`, revision `29ba90f82124961d0d902a9ea9bbb1034972af2f` |
| Arms | release MTPLX 2.11.2 (A); release plus this pull request's six optimizations (C, this pull request) |
| Engine build | MTPLX 2.11.2 on MLX 0.32.2 |
| Profile | cli-resolved (Turbo), the profile `mtplx serve` selects for this pack; a fresh SSD session cache directory per window |
| Prompt sizes | 1,024 / 8,192 / 16,384 / 32,768 / 65,536 / 131,072 / 261,120 tokens |
| Output length | 1,024 tokens maximum per request |
| Sampler | temperature 1, top-p 0.95, top-k 20 |
| Reasoning effort | `xhigh` |
| MTP depth | native, depth 3 |
| Seeds | 20260829, 20260830, 20260831 (3 seeds per load up to 32,768 tokens, 1 seed per load from 65,536 tokens) |
| Quality seeds | one seed, 20260829, at the same sampler as the speed cells (temperature 1, top-p 0.95, top-k 20, `xhigh`) |
| 16,384 protocol | interleaved A C D E, 3 windows per arm, 3 seeds each; each arm's 16K value is its fastest window; paired per-seed deltas of C against A with a 95% interval on the mean delta |
| Cell statistic | throughput cells report the fastest of the seeds with the slowest-to-fastest range in parentheses; TTFT and wall report the fastest (lowest); peak memory reports the highest |
| 261,120 note | both release and this pull request exceed the memory knob at this cell (Section 1.5) |
| n-gram table | pre-warmed before each cell |
| Memory cap | 100 GiB on every arm |
| Prefill state | cold; cross-request prefix restore off, so each seed prefills fresh; new prefill tokens equal prompt tokens; every cell cold by receipt |
| Thermal state | fans at maximum, a 40 degree Celsius gate before every cell |

Per-optimization attribution was not measured on this base; the numbers of record are the whole-PR deltas from the ladder (Section 1) and the 16K interleave (Section 1.6). A partial remainder-only arm (release plus the four remainder optimizations, at 1K, 8K and 32K) is kept as supplementary data in the docs appendix under `docs/perf/`; it is not one of the two arms in these tables.

---

## 3. What each optimization does

| Optimization | Phase | Exactness | Key |
| --- | --- | --- | --- |
| Hyper-connection read at M4 | decode | rounding | `MTPLX_QWEN4_HC_M4` |
| QSA split-K sparse decode | decode | rounding | `MTPLX_QSA_SPARSE_DECODE` |
| Cached async PLE rows | decode | exact | `MTPLX_QWEN4_PLE_CACHED_AUX` |
| Pooled QSA row selection | decode | exact | `MTPLX_QSA_POOLED_ROWSEL` |
| Prefill causal-mask fuse | prefill | rounding | `MTPLX_QWEN4_PREFILL_MASK_FUSE` |
| QSA prefill query tile | prefill | rounding | `MTPLX_QSA_PREFILL_QUERY_TILE` |

#### Hyper-connection read at M4 (`MTPLX_QWEN4_HC_M4`)
- **Problem.** The verify step reads the hyper-connection for a small block of rows as a chain of normalize, down-project and up-project steps.
- **Change.** It runs the whole read as one GPU kernel over a compact 8-bit copy of the weights. No native extension.
- **Effect.** Rounding-class output; decode gets faster.
- **Exactness.** Rounding: differs from stock only by floating-point rounding; covered by the quality gate.
- **Files.** `mtplx/kernels/qwen4_m4_hyper_read.py`, `mtplx/models/qwen4_exp.py`.
- **Switch.** Default on for a served Flash-Next pack. `MTPLX_QWEN4_HC_M4=0` turns it off.

#### QSA split-K sparse decode (`MTPLX_QSA_SPARSE_DECODE`)
- **Problem.** Each verify cycle the sparse attention gathers the selected KV rows into a large tensor per layer, then reads it.
- **Change.** A native split-K kernel reads the selected rows once in place and skips the gather and the extra copy. It needs the native extension `mtplx_native_qsa`; if the extension is not built the optimization declines to stock, prints the reason, and still serves.
- **Effect.** Rounding-class output; decode gets faster.
- **Exactness.** Rounding: differs from stock only by floating-point rounding; covered by the quality gate. A load-time parity probe on stock MLX gives worst relative L2 3.14e-05 against an fp32 reference with top-1 identical (1.0000), and relative L2 4.60e-03 against the stock-gather path with top-1 0.9896.
- **Files.** `mtplx/kernels/qsa_sparse_decode.py`, `native_extensions/qsa_sparse_gqa/` (`mtplx_native_qsa`).
- **Switch.** Default on for a served Flash-Next pack. `MTPLX_QSA_SPARSE_DECODE=0` turns it off.

#### Cached async PLE rows (`MTPLX_QWEN4_PLE_CACHED_AUX`)
- **Problem.** The model builds the extra per-layer embedding lookup inside the timed compute on every decode step, so the step waits for it.
- **Change.** A CPU-side native provider computes the fixed 64 row IDs and reads the cold rows, reusing the stock row cache, and the auxiliary embedding plane is built beside the compiled verify step with `mx.async_eval` so it overlaps the compiled replay instead of sitting on the critical path. It needs the native extension `mtplx_native_ple_cpu_rows`; if the extension is not built the optimization declines to stock and still serves, still exact.
- **Effect.** Byte-identical output; decode gets faster.
- **Exactness.** Exact: byte-identical output.
- **Files.** `mtplx/ple_cached_aux.py`, `mtplx/ple_cached_row_handoff.py`, `native_extensions/ple_cpu_rows/` (`mtplx_native_ple_cpu_rows`).
- **Switch.** Default on for a served Flash-Next pack. `MTPLX_QWEN4_PLE_CACHED_AUX=0` turns it off.

#### Pooled QSA row selection (`MTPLX_QSA_POOLED_ROWSEL`)
- **Problem.** The twelve QSA indexers each rebuild the same pool-kernel setup on every decode step, and each holds its own copy of the RoPE rotation table.
- **Change.** It binds the existing `qsa_indexer_pool_keys_metal` kernel metadata once per indexer at construction and shares one `inv_freq` object across all twelve indexers, so the repeated per-step setup leaves the decode step. Stock MLX, no native extension. The install validates the 48-layer QSA layout, the per-indexer geometry, the RMS-norm epsilon, the RoPE scale and the shared `inv_freq` identity, and a contract failure fails the model load.
- **Effect.** Byte-identical output; decode gets faster.
- **Exactness.** Exact: byte-identical output.
- **Files.** `mtplx/qsa_pooled_rowsel.py`.
- **Switch.** Default on for a served Flash-Next pack. `MTPLX_QSA_POOLED_ROWSEL=0` turns it off.

#### Prefill causal-mask fuse (`MTPLX_QWEN4_PREFILL_MASK_FUSE`)
- **Problem.** During prefill the dense attention builds a full score tensor and applies the causal mask by hand, because at head_dim 256 MLX declines its fused kernel and takes this unfused route.
- **Change.** It sends the same attention through MLX's fused SDPA with a causal flag, so it never builds the full score tensor. Stock MLX, no native extension. It declines only for the verify-width dead band (query length 3 to 8), falls back to stock there, and logs the refusal.
- **Effect.** Rounding-class output; prefill reads and writes less memory. It does not by itself make 261,120 tokens fit: the binding allocation there is the verify's KV-cache write, not attention, and this optimization declines in the verify's own shape band anyway (Section 1.5).
- **Exactness.** Rounding: differs from stock only by floating-point rounding; covered by the quality gate.
- **Files.** `mtplx/models/qwen4_exp.py`.
- **Switch.** Default on for a served Flash-Next pack. `MTPLX_QWEN4_PREFILL_MASK_FUSE=0` turns it off.

#### QSA prefill query tile (`MTPLX_QSA_PREFILL_QUERY_TILE`)
- **Problem.** The dense QSA attention in a wide prefill chunk builds a large score block.
- **Change.** It splits the query rows into tiles and runs attention one tile at a time, so a wide chunk keeps a narrow chunk's peak at the same total work. Stock MLX, no native extension. The value is rows per tile; the default 2,048 is inert at the production chunk width.
- **Effect.** Rounding-class output; prefill keeps a lower memory peak at the same work.
- **Exactness.** Rounding: differs from stock only by floating-point rounding; covered by the quality gate.
- **Files.** `mtplx/qwen4_prefill_chunk.py`, `mtplx/models/qwen4_exp.py`.
- **Switch.** Default on for a served Flash-Next pack. `MTPLX_QSA_PREFILL_QUERY_TILE=0` turns it off.

Every old `MTPLX_FABLE_*` name still works as an alias when the new key is unset.

### Four upstream optimizations that were silently off

These are not new optimizations. They are upstream 2.11.2's own re-landed verify optimizations, now actually engaged; an audit of the ~31 auto-armed keys found four whose readers were dead as served.

| Key | Reader | What it does in the verify step |
| --- | --- | --- |
| `MTPLX_QWEN4_DRAFT_K20_PRESCATTER` | `qwen4_draft_k20_prescatter.py:200` (re-cached `generation.py:323`) | pre-scattered K20 draft read on the device draft chain |
| `MTPLX_QWEN4_BLOCK_VERIFY` | `qwen4_block_verify.py:132` (re-cached `generation.py:337`) | block speculative verification in the accept loop |
| `MTPLX_QWEN4_OPDIET` | `runtime_options.py:61` | exact-preserving op diet in the compiled fixed-M4 verify graph |
| `MTPLX_QWEN4_VERIFY_GLUE` | `runtime_options.py:159` | fused QSA-rope glue inside the fixed-M4 verify body |

Cause: `mtplx serve` imports the generation and runtime modules before it stamps the auto-armed keys, so these four readers froze their default (off) at import and the auto-arm was a silent no-op; the arm A (release) numbers above were measured with the four off, which is the release as a user runs it. Fix: this pull request's base commits change the four readers to resolve the environment at use, change no defaults, and add `tests/test_qwen4_remainder_arming.py`, which replays the served import order. Arm C carries the fix, so it includes the four re-engaged optimizations.

### What is not ported

The fifth PR 391 optimization, verify graph-build overlap (`MTPLX_QWEN4_GRAPH_BUILD_OVERLAP`), is not in this pull request. Upstream 2.11.2 re-landed the fixed-M4 verify as a single compiled graph. The optimization rode a prefix and suffix split of that graph, and that split does not exist upstream. Porting it is a redesign of the compiled verify, and its whole claim is a submission-timing reorder that cannot be proven on the CPU path.

Feasibility, from the port report: on the PR 391 campaign the optimization targeted 1.934 ms per cycle of host-late GPU idle at 16,384 tokens; at the default one-layer split the binding term is about 0.53 ms per cycle, about +1.4 tok/s (about 1.4%); at a three-to-four-layer split it peaks near 1.7 to 1.8 ms per cycle, about +3.6 to +4.7 tok/s (about 4.5%). The optimization is exact by construction, so it is a speed-only optimization, not a quality risk. Recommendation: defer it to a dedicated split-compilation task with a bit-exact A/B gate. The one-page feasibility is in `over100-reports/remainder-port-report.md`.

---

## 4. Quality

The four remainder optimizations are rounding-class, so this pull request ships on the quality gate, not on bit-identity. The two aux optimizations are exact and do not affect it. The gate is the repo's own code-eval, HumanEval and MBPP, served, paired release against this pull request. Every cell runs at the benchmark's own sampler: temperature 1, top-p 0.95, top-k 20, reasoning effort `xhigh`, one seed (20260829). That is the setting the speed numbers were measured at, so the quality cells answer the question the speed cells raise.

Each cell reports three numbers per arm: strict pass@1 counts a task cut off at the output cap as a failure; completed-task pass@1 excludes the cut-off tasks; and the truncation rate carries the mean and max completion tokens. With reasoning `xhigh` at temperature 1 the release cut off about 10% of HumanEval tasks at an 8,192-token cap while still thinking, so the strict number measured verbosity, not correctness; the cap is now 32,768 tokens. Truncation is reported in its own column and never re-run, so a non-zero rate is data rather than a failed cell; the strict and completed-task columns bracket what it costs.

| Suite | Arm | Strict pass@1 | Completed-task pass@1 | Truncation (rate; mean/max tokens) |
| --- | --- | ---: | ---: | ---: |
| HumanEval | release (A) | 0.9695 | 1.0000 | 3.0% (mean 3058 tok) |
| HumanEval | this PR (C) | 0.9634 | 0.9814 | 1.8% (mean 2734 tok) |
| MBPP | release (A) | not run | not run | not run |
| MBPP | this PR (C) | 0.9461 | 0.9830 | 3.7% (mean 3367 tok) |

_Provenance of the candidate rows: this pull request's own tree was not re-run for quality. The candidate cells are the typical-acceptance sweep's **exact** cell, served from `264e0835` with the typical rule switched off. That tree is this pull request's code plus the typical rule, so with the rule off it executes this pull request's optimizations and nothing else. Release MBPP at this sampler was never run, and the programme was cut after the exact arm, so that cell reads "not run"; the greedy release MBPP figure is in the supplementary row of Section 1.7._

_Scoring: prompt + solution + tests, solution = last fenced block defining the entry point (evalplus sanitize). Re-scored offline from saved completions after the live gate was found to drop the prompt's helper definitions (HumanEval/38, /50)._

Each optimization carries a `=0` opt-out, the kill switch if an optimization moves quality.

---

## 5. Changes that did not work

These candidates were measured and rejected on the earlier #391 stack. Their code is not in this pull request. The numbers are from that stack, at the 17,408-token shape.

| Candidate | Measured effect | Reason |
| --- | --- | --- |
| Native sidecar sync raw PLE | about -8.5% | The synchronous read sits on the critical path; the cached optimization supersedes it. |
| Native sidecar async raw PLE | about -1.9% | The uncached predecessor of the cached optimization. |
| GPU-stream PLE transport | -1.88% | The GPU-stream factory did not beat the CPU queue transport. |
| Queued sampled-D3 selector | -4.49% | The extra device-to-host boundary adds draft cost. |
| MTP depth 4 | about -30% | The deeper draft costs more than it accepts. |
| Draft temperature 0.85 | about -2.5% | The benchmark contract is temperature 1. |

The full rejected inventory with per-row numbers is in the perf report under `docs/perf/`.

---

## 6. How to run and how to disable

Build the venv and the native extensions with the setup script, then serve the pack. The server arms all six optimizations by default:

```bash
scripts/fable/setup_over100_venv.sh
mtplx serve \
  --model ~/.mtplx/models/Youssofal--Qwen3.8-Flash-Next-MTPLX-Optimized-Speed \
  --model-id mtplx-flash-next-optimized-speed
```

Read `GET /health` and confirm the install verdicts name the optimizations. For the QSA sparse decode optimization, confirm the `[mtplx] qsa_sparse_decode armed:` line; a serve that printed `declined to stock` ran the stock path.

Turn one optimization off with its key set to `0`:

```bash
MTPLX_QWEN4_HC_M4=0 mtplx serve --model ...
MTPLX_QSA_SPARSE_DECODE=0 mtplx serve --model ...
MTPLX_QWEN4_PLE_CACHED_AUX=0 mtplx serve --model ...
MTPLX_QSA_POOLED_ROWSEL=0 mtplx serve --model ...
MTPLX_QWEN4_PREFILL_MASK_FUSE=0 mtplx serve --model ...
MTPLX_QSA_PREFILL_QUERY_TILE=0 mtplx serve --model ...
```

The two native optimizations need their extensions built in the serve environment: `mtplx_native_qsa` for QSA sparse decode and `ple_cpu_rows` for cached PLE. The setup script builds both. Without an extension its optimization declines to the stock path and still serves.

---

## 7. File map and provenance

| Area | Files |
| --- | --- |
| Runtime | `mtplx/models/qwen4_exp.py`, `mtplx/runtime_options.py`, `mtplx/qwen4_prefill_chunk.py`, `mtplx/runtime.py`, `mtplx/ple_cached_aux.py`, `mtplx/ple_cached_row_handoff.py`, `mtplx/qsa_pooled_rowsel.py`, `mtplx/qwen4_aux_lanes.py` |
| Kernels | `mtplx/kernels/qwen4_m4_hyper_read.py`, `mtplx/kernels/qsa_sparse_decode.py`, `mtplx/native/__init__.py` |
| Native extensions | `native_extensions/qsa_sparse_gqa/` (`mtplx_native_qsa`), `native_extensions/ple_cpu_rows/` (`mtplx_native_ple_cpu_rows`) |
| Server | `mtplx/server/openai.py`, `mtplx/profiles.py` |
| Wheel signer | `scripts/bundle_native_runtime_wheel.py` |
| Tests | `tests/test_qwen4_hc_m4.py`, `tests/test_qwen4_prefill_mask_fuse.py`, `tests/test_qsa_sparse_decode.py`, `tests/test_qsa_sparse_decode_wiring.py`, `tests/test_qsa_sparse_gqa_native.py`, `tests/test_pr391_ple_cached_aux_cpu.py`, `tests/test_pr391_ple_cached_row_handoff_cpu.py`, `tests/test_pr391_fixed_m4_pool_install_cpu.py`, `tests/test_qwen4_aux_lanes.py`, `tests/test_bundle_native_runtime_wheel.py` |
| Documentation | the perf report and charts under `docs/perf/` |

- Base: upstream main `21be78b3` (MTPLX v2.11.2).
- Branch: `perf/qwen38-aux-lanes-main` @ `72f58d41` = upstream main, then four remainder commits (hyper-connection read, prefill mask fuse, QSA prefill query tile, QSA split-K sparse decode), then one aux commit (cached async PLE rows, pooled QSA row selection).
- Optimizations source, read-only: PR 391 head `a5e38bb7`.

History. On the PR 391 campaign at 16,384 tokens the release default reached 71.17 tok/s, the full PR 391 stack reached 80.92 tok/s, and upstream 2.10.2 reached 57.65 tok/s on the same instrument. The nine keys 2.11.2 re-landed recovered about 60% of the full-stack gain; these six optimizations carry the rest and the two exact decode optimizations. On the earlier #391 stack the two aux optimizations together gave +1.59% decode over the same build with both off, byte-identical, in a same-build pair at 16,384 tokens. Those numbers are superseded by the two-arm battery above, which is measured on 2.11.2.


youssofal:
David, thank you for this. I read the full diff, ran your test files (562 passed, 23 skipped on CPU), built both native extensions in a scratch directory, and checked the C++ n-gram row hash against the numpy one on 3,690 cases (0 mismatches). Here is where I landed.

Things I want to take:

- The cached async PLE rows path. I verified it produces the same bytes as the current path at every stage except the GPU op itself (same rows, same dequantize call, same LRU order), so I consider it exact.
- Pooled QSA row selection, after one change (below). It reuses the existing fused pool kernel, which the current tests already pin as bit-exact against the eager chain.
- The query-tile plumbing, which is inert at the shipped 2048-token prefill chunk.

Things that need fixing before they can merge:

1. Block verification. Your change reads `MTPLX_QWEN4_BLOCK_VERIFY` at use, and the server already stamps it on for every served Flash-Next pack, so merging as-is would switch the accept loop to block verification in production. The law that was in the tree was not exact at depth 3 (I have an enumeration oracle showing the sampled distribution drifts from the target). I have rewritten it on our side so it is provably exact, with the oracle as a test, and I am measuring its speed tonight. Please drop the block-verify part of the last two commits and I will handle that lane separately.
2. The prescatter re-engagement does not work. `claim_draft_route` still checks the raw module global, which is now `None`, so it returns early and the lane never runs while `/health` reports it as armed. One-line fix: check `is_enabled()` there. This also means your arm C ran without prescatter.
3. Setting `MTPLX_QWEN4_OPDIET=0` now makes the server fail to boot, because pooled row selection is stamped on unconditionally and its installer requires op-diet. It should only be stamped when op-diet resolves on, or the installer should decline instead of raising.
4. The native wheels cannot be bundled as written. Neither `setup.py` declares an exact `mlx==0.32.2` requirement, your own bundler rejects a wheel without one, and the sparse-GQA `pyproject.toml` pins nanobind 2.12.0, which its own CMake guard refuses against mlx 0.32.2. The existing `qsa_kernels` extension shows the convention. I would also rather see the decode kernel added to `qsa_kernels` than a second package that duplicates the prefill kernel.

Two smaller notes: the HC-M4 kernel reads bf16 weights, not an 8-bit copy as the description says, and the `4096`-token prefill docstring is stale.

On the speed claim: the +15% at 16K is the whole stack including the frozen-key re-engagement, and your own knock-out table attributes roughly 6 to 8 points to the six lanes themselves. I have already landed a general fix for the frozen keys on our side (one refresh after the profile environment is applied), so that part of the gain will ship regardless. For the four rounding-class lanes (HC read, mask fuse, sparse decode, and the query tile when it is active) I want the HumanEval/MBPP cells in the table filled in before they go on by default, and I will run an A/B on this machine with each lane switched individually.

If you can rebase with the four fixes above and split the docs artifacts into their own commit, I can take the exact lanes right away and evaluate the rest with numbers.

youssofal:
A short update on the exact lanes. The pooled row selection lane is ported on our side with your authorship, gated so that a daemon with op-diet off still boots on the stock path, with your two test files and our environment and health tests green. On this M5 Max with the Flash-Next Speed pack, an 8.8k-token prompt decoded greedily to 1,024 tokens is byte-identical with the lane on and off, with the same acceptance ladder and 71.8 against 72.5 tok/s, so at that length it changes nothing either way. It needs the 16K and 100K cells before it goes on by default. The cached PLE rows lane waits on the native pins from the review.

youssofal:
2.11.3 ships the fix for the frozen setting readers this PR found, with credit in the release notes. The pooled row-selection kernel measured a tie on the 9k-token prompt (79.4 to 79.6 tok/s with it, 79.6 to 79.7 without), so it is not in this release and the PR stays open. Release: https://github.com/youssofal/MTPLX/releases/tag/v2.11.3

PR #489, opened 2026-09-09
Fix remote Qwen detection and keep AR-only sources out of speculative Forge

PhilipJohnBasile:
Remote Qwen checkpoints without MTP weights, including `nex-agi/Nex-N2.5-mini`, were reported as `missing-model-weights` even when the Hugging Face listing contained all trunk shards. Use the existing local/remote trunk check so runtime inspection recognizes these as autoregressive checkpoints with MTP disabled. Separately, require MTP weight evidence for Forge: `can_run` also includes AR-only models and must not admit those into speculative conversion.

The guidance now explains that Forge cannot create missing trained heads. AR-only Forge conversion remains unsupported; this change fixes detection and guidance, not conversion support or model qualification.

Validation:
- New regressions reproduced both defects before the fix; config-only sources remain refused.
- All 180 artifact and Forge tests passed natively on macOS; Ruff and `git diff --check` passed.
- Live metadata-only probe of `nex-agi/Nex-N2.5-mini` at `87420286149d9cce9bd46cd335ef9bda33c37c1b` sees 16 shards, `can_run=true`, `mtp_supported=no`, and `native-ar-only-missing-mtp`; Forge still refuses speculative conversion.
- No full model download, load, inference, or performance qualification was performed.

### DIFF
--- a/mtplx/backends/registry.py
+++ b/mtplx/backends/registry.py
@@ -1766,14 +1766,7 @@ def compatibility_for_inspection(inspection: Any) -> CompatibilityVerdict:
             # A dir with no TRUNK weights at all is a different failure — no
             # model, not a missing head — and keeps a clean human refusal
             # instead of a FileNotFoundError deep in the loader.
-            try:
-                trunk_weights_exist = any(
-                    path.name != "mtp.safetensors"
-                    for path in Path(model_dir).glob("*.safetensors")
-                )
-            except OSError:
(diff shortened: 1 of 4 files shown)


youssofal:
Merged for 2.11.3, your commit ef3919ba with its authorship kept (merge e062a465), thank you. Release: https://github.com/youssofal/MTPLX/releases/tag/v2.11.3

Issue #401, opened 2026-08-30
Dashboard request timestamps and aggregate MTP acceptance do not render


johncarterlives:
## Summary

Two dashboard values cannot render from the current request metrics contract:

1. Every Recent requests `WHEN` cell displays `—`.
2. The Speculative panel displays `— accepted of — drafted` even when its per-depth acceptance bars contain valid values.

Both behaviors were reproduced with a synthetic request on MTPLX 2.10.0 and remain present in the current `main` source. They do not appear model- or hardware-specific, so no environment dump or identifying machine information is included.

## Reproduction

1. Start MTPLX with the packaged dashboard enabled.
2. Complete any synthetic chat request.
3. Open `/dashboard/` and inspect Recent requests and Speculative.

## Expected behavior

- Recent requests renders a relative completion time.
- Speculative renders aggregate accepted and drafted token totals consistent with the per-depth arrays.

## Actual behavior

- Recent requests renders `—` for `WHEN`.
- Per-depth acceptance renders, but the aggregate subtitle renders `— accepted of — drafted`.

## Source evidence

- `dashboard/src/components/RequestLogTable.tsx` hard-codes the `WHEN` cell to `—` and leaves `relativeTime` unused. `CompletedEvent.when_s` is transient, while `MetricsLatest` has no persisted completion timestamp for rows returned by `/metrics`.
- `dashboard/src/components/DepthAcceptanceBars.tsx` reads `latest.accepted_drafts` and `latest.drafted_tokens` for its subtitle.
- `mtplx/server/openai.py::_metrics_envelope` emits `accepted_by_depth` and `drafted_by_depth`, but not those aggregate fields.

## Suggested direction

- Add a persisted completion timestamp to the metrics envelope/type and render it with the existing `relativeTime` utility.
- Either emit `accepted_drafts` and `drafted_tokens` in the envelope or derive both totals from the per-depth arrays in the dashboard.

## Data handling

This report contains only synthetic observations and public upstream source paths. It intentionally excludes prompts, responses, model identifiers, session identifiers, local filesystem paths, hardware identifiers, logs, and private repository details.

---
[aidevops.sh](https://aidevops.sh) v3.32.294 plugin for [OpenCode](https://opencode.ai) v1.18.25 with gpt-5.6-sol

youssofal:
[...]

PR #490, opened 2026-09-10
fix(dashboard): derive accepted/drafted totals from the per-depth arrays

dajiaohuang:
## Summary

The Per-depth acceptance subtitle and the Drafted / verify call tile read `accepted_drafts` and `drafted_tokens`, which `_metrics_envelope` does not emit — it emits `accepted_by_depth` and `drafted_by_depth`. Both surfaces therefore rendered `— accepted of — drafted` and `0.00 tok/call` while the bars beside them drew real per-depth numbers.

Both now total the per-depth arrays they already receive, through a shared `depthTotal` helper. The helper still prefers a flat total, so nothing changes if the envelope later carries one.

This is the aggregate half of #401. The Recent requests `WHEN` column is deliberately untouched: no metrics row persists a completion timestamp (the `last_metrics` entries carry durations only), so a relative time needs a new field on the envelope and on its request-log row — a server-side contract change rather than a dashboard one.

## Verification

- `bun run typecheck` (`tsc --noEmit`): clean.
- `bun run build` (`tsc -b && vite build`): success. The rebuilt `mtplx/dashboard/_static` is included here; the CSS asset hash is unchanged (`index-ZaQvA7xU.css`), so only the JS chunk moved.
- The rebuilt bundle was confirmed inside the wheel from `python -m build`: it carries `assets/index-CUmjZF08.js` and no longer carries the stale `index-CjL8TS30.js`.
- The derivation was exercised directly against real envelope shapes with `bun`: per-depth arrays only gives `13 verify calls · 23 accepted of 36 drafted` and `2.77 tok/call`; an emitted flat total still wins; empty or absent arrays give `—` and `0 drafted · 0 verifies`.
- `python -m build`: success (`mtplx-2.11.2`).
- `python -m pytest tests/test_no_mlx_imports.py tests/test_runtime_kpis.py`: 13 passed, 8 failed, and all 8 failures are Windows-only artifacts (no `mlx`; `mtplx/thermal.py` imports `pwd`, which does not exist on Windows). The diff changes no `.py` file. `tests/test_public_cli.py` cannot be collected on Windows at all (`os.geteuid`), and CONTRIBUTING's `scripts/fresh_venv_smoke.sh` expects a POSIX venv layout (`$VENV/bin/python`), so neither ran here.
- `git diff --check`: clean.

## Benchmark Evidence

Not applicable — this changes dashboard rendering only, not the runtime path.


youssofal:
Thank you, this is in 2.12.0. Your commit is in main under your name, 292dd38b. The dashboard now computes its draft totals from the per-depth counts when a report has no totals of its own. The commit was applied directly, so GitHub will show this PR as closed rather than merged.

Issue #456, opened 2026-09-04
Flash-Next: active_bytes creeps past the Metal limit under deep-session churn (103.1 GB > 96 GiB with bank=0, cache=0, live sessions 7.8 GB) — nothing sheddable, only restart recovers


nomishbhardwaj:
Config: MTPLX 2.11.1, M5 Max 128 GB, `Youssofal--Qwen3.8-Flash-Next-MTPLX-Optimized-Speed`, `--mtp --depth 3 --draft-temperature 0.3 --profile turbo --context-window 262144`, temp 0.3 / top-p 0.9 / top-k 40, reasoning high, SSD session cache on, `MTPLX_SUSTAINED_DENSE_DECODE_MAX_CONTEXT=262144`, marathon-protect envs. (This particular server ran the #453 branch, which is why the receipt below shows the `live_session` estimate mode — but the defect reproduces on stock 2.11.1 and 2.10.2, receipts below, and the shed's own numbers show it is not the guard's doing: there was nothing left to shed.)

**After 62 minutes of real agent use, `active_bytes` exceeded the Metal limit itself and every request 507'd regardless of size.** Fresh server, one coding-agent session grown to ~116K (plus its post-compaction lineage), several client-side retry/abort cycles along the way (the client's no-data watchdog re-sends while a deep prefill is still grinding). End state, from the serve log:

```
prefill_admission_shed prompt_tokens=123724 reusable_prefix_tokens=116033
  reusable_prefix_mode=live_session miss_tokens=7691
  active_bytes=103136170688 cache_bytes=0
  projected_bytes=106606953728 threshold_bytes=99986838650 limit_bytes=103079215104
  bank_bytes_before=0 bank_bytes_after=0 cache_cleared=true
→ allocation_failure_shed, HTTP 507 — repeatedly, including for a 7,691-token miss
```

`active_bytes` 103.1 GB is over the 96 GiB engine limit on its own, with the RAM bank at 0 and the MLX cache at 0. There is no eviction policy that can help this state; only a restart recovers it.

**The live memory does not account for it.** `/admin/sessions` at that moment:

```
anon-768016d2… prefix=116033 bytes=4.05G
anon-40dc6e37… prefix=106066 bytes=3.77G
(+ two ~0-byte sessions)
TOTAL live session bytes: 7.8G across 4 sessions
```

[...]

Issue #450, opened 2026-09-04
qwen4 fixed-M4 lane gate counts idle session-bank bytes as live and never asks the bank to yield (128 GB Mac, 2.11.1)


reports-svg:
**Version:** 2.11.1 (PyPI). Machine: M5 Max 128 GB, `iogpu.wired_limit_mb=112000`. Model: `Youssofal--Qwen3.8-Flash-Next-MTPLX-Optimized-Speed`. Serve flags: `--profile turbo --depth 3 --generation-mode mtp --scheduler-mode cooperative --batching-preset agent --max-active-requests 4 --ssd-session-cache on`.

## What we saw

At ~136k context, every request after the first fell from ~37 tok/s (2.10.1, same machine, same session) to ~13 tok/s. Receipt, verbatim:

```
[qwen4-fixed-M4] lane skipped for this request, plain eager batched verify:
  prompt 136549 tokens: live 98.8 GB + promotion 3.9 GB over the 100.0 GB line
  (allocator cache released: 4.2 GB)
```

Immediately before it, the prefill-admission shed had run and **admitted** the request without evicting anything:

```
[mtplx] memory guard {"action": "prefill_admission_shed", "prompt_tokens": 136549,
  "active_bytes": 89201410050, "cache_bytes": 8521077009, "projected_bytes": 105374454483,
  "threshold_bytes": 99986838650, "limit_bytes": 103079215104,
  "bank_bytes_before": 7761837632, "superseded_session_entries_evicted": 0,
  "lru_entries_evicted": 0, "bank_bytes_after": 7761837632, "cache_cleared": true}
```

So: 7.76 GB of session bank was live, none of it belonged to the incoming request, and the lane gate failed by 2.7 GB with that bank still counted.

## Why (from source, 2.11.1)

- `generation._qwen4_fixed_m4_lane_fits` computes `live = get_active_memory() + get_cache_memory()` and requires `live + (prompt + growth_reserve) × 28,416 B ≤ 0.97 × memory_limit`. On failure it releases the **allocator cache** and re-checks — but it never asks `SessionBank` to yield or evict, even though the bank's own docs say its ceiling "yields to live KV at runtime."
- `_qwen4_fixed_m4_compiled_verify_requested` calls the gate directly (it even `del cached_tokens`); there is no bank yield/evict on that path.
[...]

PR #500, opened 2026-09-15
fix(memory): float admission-shed and pressure guard on real phys_footprint, not just MLX allocator stats

ArctifoxNL:
## Summary

Two independently reported kernel panics (watchdog timeout, near-zero free pages) on an M5 Max / 128 GB Mac running MTPLX 2.11.2, both traced to a single MTPLX worker process resident *above* its own configured `MTPLX_MEMORY_LIMIT_BYTES` by roughly 1-1.25 GB. Related: #456, which shows the same accounting gap from a different angle (`active_bytes` itself drifting ~17 GB past a fresh floor with the bank and MLX cache both at zero).

Every memory guard in `server/openai.py` -- the prefill admission shed (including the #450 refusal check), the background pressure loop, the qwen4 fixed-M4 lane gate -- compares against `mx.get_active_memory() + mx.get_cache_memory()`: MLX's own account of what it allocated through Metal. That account can drift below what the kernel actually holds resident for the process (touched-then-kept weight pages, non-Metal Python/NumPy heap, thread stacks, the session bank's host-side buffers), so a request can be admitted -- or a generation left running -- at a footprint the allocator calls safe while the OS is already out of free pages.

## Fix

New module `mtplx/os_memory.py`: `phys_footprint_bytes()` reads `RUSAGE_INFO_V4.ri_phys_footprint` via `proc_pid_rusage` -- the same counter macOS's own jetsam/memory-pressure subsystem uses to make this exact call. Stable Darwin API since macOS 10.9, no new dependency (stdlib `ctypes` only), no version gating needed for MTPLX's Apple-Silicon-only, macOS 14+ target.

Wired in as an independent floor, `max(allocator_bytes, phys_footprint_bytes())`, at three points:
- `_prefill_admission_shed`'s cheap worst-case gate and initial projection
- the post-cache-clear recompute
- the final post-reclamation refusal check (the exact check #450 added)
- `_allocator_pressure_level` (feeds the background pressure loop and the sustained-CRITICAL prefill abort)

Because it's a `max()`, a healthy process -- or a platform where the probe fails (non-Darwin, missing libproc) -- sees byte-identical behavior to today. It can only make a guard trip *earlier or more accurately*, never later.

## A note on how this was built

An earlier draft of the `rusage_info_v4` struct (transcribed from memory, not the SDK header) was 232 bytes instead of the real 296 -- `proc_pid_rusage` writes a full reply regardless of the caller's buffer size, so that silently overran the ctypes buffer by 64 bytes on every call. It didn't crash immediately; it corrupted heap metadata that crashed ~2000 calls later at an unrelated allocation, reproduced with a tight Python loop calling the probe in isolation. Fixed by checking the real struct against `<sys/resource.h>` in the SDK (confirmed `sizeof(struct rusage_info_v4) == 296` via a compiled C snippet) and adding a size assertion in `_load_libproc` that disables the probe entirely if a future SDK ever changes this -- RUSAGE_INFO_V4 is a frozen ABI, so this should never trigger, but the assertion converts "silent heap corruption" into "feature quietly off" if it ever does.

## Testing

- New `tests/test_phys_footprint_guard.py`: the probe itself (plausible value, fails closed on a bad pid), the admission-shed floor (inert when the probe fails or reads lower than the allocator; triggers the cheap gate and forces the final refusal when the real footprint alone crosses the line), and `_allocator_pressure_level` (CRITICAL/WARNING from real footprint alone, unaffected when the probe fails or reads lower).
- Full existing suite: `python -m pytest tests/` -- all green, no regressions.
- `python -m pytest tests/test_no_mlx_imports.py tests/test_public_cli.py tests/test_runtime_kpis.py`, `python -m build`, `scripts/fresh_venv_smoke.sh` -- all pass per CONTRIBUTING.md.
- `ruff check` on the diff: zero new findings versus the pre-patch baseline.

Hardware used for the originating incidents: M5 Max, 128 GB, macOS 26.5.1, MTPLX 2.11.2, `Youssofal--Qwen3.8-Flash-Next-MTPLX-Optimized-Speed`. Happy to share the two panic-log-derived receipts (memoryStatus, processByPid entries) if useful.


ArctifoxNL:
My mac kept crashing due to memory overload, this seems to fix it. I am still stress testing it, if I find anything I will let you know.

ArctifoxNL:
the cache shedding seems more reliable now.

youssofal:
Thank you, this is in 2.12.0, and it fixes #516. Your commit is in main under your name, d2041193. I added one follow-up: memory a daemon holds outside Metal now counts only above an allowance, set by `MTPLX_HOST_MEMORY_ALLOWANCE_BYTES` (03ece458). The guards also read how much memory macOS can still hand out. The commit was applied directly, so GitHub will show this PR as closed rather than merged.

Issue #493, opened 2026-09-11
Session Bank lacks automatic cleanup mechanism, causing OOM in production


taozhiyuai:
# Session Bank lacks automatic cleanup mechanism, causing OOM in production

## Summary
The MTPLX session bank accumulates orphaned blob files without automatic cleanup, leading to excessive disk/memory usage and potential OOM conditions in long-running production environments.

## Environment Details
- **MTPLX Version**: 2.11.2
- **Model**: Qwen3.8-Flash-Next-MTPLX-Optimized-Speed (107GB weights)
- **Hardware**: M3 Max MacBook Pro, 128GB RAM
- **OS**: macOS 15.7.9 (Sequoia)
- **Test Date**: 2026-09-11

### Observed Issue
After running MTPLX server for ~3 weeks in a production-like environment, the session bank accumulated **394,155 blob files totaling 44.1 GB** while only maintaining **17 active entries**.

**Key Metrics:**
- Total blobs: 394,155 files (44.1 GB)
- Active entries: 17 entries (11 total hits)
- Blob/Entry ratio: 23,185:1

**Memory impact:**
- Model baseline: 77.9 GB
- Session bank overhead: ~15 GB (resident portion)
- 79K tokens request: 102.8 GB peak (triggered Jetsam)
- 131K context: OOM risk

**Jetsam Event Log:**
- Timestamp: 2026-09-11 17:34:12 GMT+8
- Bug Type: 298 (Jetsam)
- Memory pressure: 96.8% (123.95 GB used of 128 GB)
- Free memory: 2.76 GB
- Largest process: ds4-server (concurrent workload)

### Root Cause Analysis
1. **Orphan Blobs**: The session bank creates new blob files for each session but does not automatically prune orphaned blobs when entries expire or are deleted.
2. **No GC Mechanism**: There is no built-in garbage collection or cleanup command.
3. **Manifest vs Disk Mismatch**: The manifest.sqlite tracks 17 active entries, but the blobs directory contains 394K+ orphaned files from previous sessions.

### Impact
- **Memory Pressure**: Session bank's resident memory (~15 GB) reduces available KV cache budget
- **OOM Risk**: 131K context configuration triggered macOS Jetsam memory protection
- **Disk Bloat**: 44 GB of stale data accumulated over 3 weeks
- **Performance Degradation**: Lower context windows must be used to avoid OOM

## Cleanup Results
After manual cleanup:
- Session bank size: 4.1 MB (was 45 GB)
[...]

PR #502, opened 2026-09-15
feat(cli): add `mtplx gc` to reclaim orphaned SessionBank SSD cache files

ArctifoxNL:
## Summary

`SessionBankColdTier._cleanup_untracked_cache_once` (`cache_bank/cold_tier.py`) already reconciles the on-disk store against `manifest.sqlite` and deletes whatever the manifest no longer references. It works; the gap is that it only ever runs as a side effect of a write that finds the store close to its configured size cap. On a Mac with generous free disk that cap is rarely approached, so orphans a user never triggers the cap against just accumulate forever: #493 measured 394,155 orphaned blobs (44.1 GB) against 17 live manifest entries after three weeks of normal use, with no CLI command to reclaim them short of deleting the whole session bank by hand.

## What this adds

`mtplx gc [--dir PATH] [--apply] [--force] [--json]` -- reports orphaned entry directories, orphaned blob files, and `evicted_entries/` leftovers by default; `--apply` deletes them.

New module `mtplx/session_bank_gc.py` reimplements the same reconciliation as a small, standalone utility rather than importing `SessionBankColdTier` directly: `cache_bank/cold_tier.py` imports `cache_bank/codec.py`, which imports `mlx.core` at module scope purely for the tensor encode/decode path this reconciliation never touches (it only reads `manifest.sqlite` and `payload.json` files, then deletes what nothing references). Routing a disk-cleanup command through that import chain would mean `mtplx gc` stops working exactly when it might matter most -- a broken or absent MLX install -- which would contradict the CLI's own "doctor and inspect run on any machine" contract (`test_no_mlx_imports.py`). Verified: `mtplx.session_bank_gc` and `mtplx.commands.public` both import cleanly with `mlx`/`mlx_lm` blocked.

Read-only by default is deliberate given the file counts involved. Because a concurrently running server writes new entries under a lock this reconciliation does not hold across processes, `--apply` checks for a running server on the probed ports (reusing `daemon_client.probe_running_daemons`, the same helper `mtplx stop` uses) and requires `--force` to proceed if one is found -- a session committed between the scan and the delete could otherwise lose its blobs.

## Testing

- New `tests/test_session_bank_gc.py`: reconciliation correctness (referenced entries/blobs survive, orphans are found and only removed under `--apply`), the CLI handler (dry-run report shape, the running-server block, `--force` override).
- Full existing suite: `python -m pytest tests/` -- all green, no regressions.
- `python -m pytest tests/test_no_mlx_imports.py tests/test_public_cli.py tests/test_runtime_kpis.py`, `python -m build`, `scripts/fresh_venv_smoke.sh` -- all pass per CONTRIBUTING.md.
- `ruff check` on the diff: zero new findings versus the pre-patch baseline.
- Manual end-to-end smoke test of both the dry-run and `--apply --force` paths against a hand-built session-bank fixture.

Related: #493.


youssofal:
Merged for 2.11.3, your commit 1d264e61 with its authorship kept (merge 94f493eb), thank you. `mtplx gc` is in, and the daemon runs the same reconciliation every time it opens the store. Release: https://github.com/youssofal/MTPLX/releases/tag/v2.11.3

PR #507, opened 2026-09-18
fix(qwen4): MTPLX_AR_PIPELINE must reproduce the classic AR sampler (byte-identical output)

davidtai:
## The bug

The pipelined-AR decode lane (`MTPLX_AR_PIPELINE`) draws its output tokens from a different RNG than the classic AR loop, so the same request and seed produce a different continuation depending on whether the lane engages.

- The classic loop calls `_sample_from_logits`, applies MTPLX's filtering order, and draws from the request's `numpy.random.Generator`.
- The lane draws output token 0 the same way, but every later token from an `mx.random` stream via `_mx_lazy_sample` (whose own docstring notes "the streams differ").

Both streams are individually deterministic, but they are not the same stream, so token 0 matches the classic loop and token 1 onward diverges. This is the serial-decoding divergence seen when comparing MTPLX against another engine: the fast path does not reproduce the engine's own reference decode. It is latent on the shipped 128 GB config (the lane declines when the n-gram table is streamed rather than resident), but real wherever the lane engages.

## The fix

Route `MTPLX_AR_PIPELINE` through the classic loop's per-token `_sample_from_logits(row, sampler, rng)` draw, so the pipelined path is byte-identical to classic AR for a given seed. Byte-identity requires the NumPy draw for every token, and that draw needs the materialized logits row, which forecloses the lane's build-ahead overlap — there is no overlap left to keep, so the flag now shares the classic path. The model's pipeline forward-mode is still offered via `set_ar_pipeline_mode` (gating stays observable) but is released before decode, so decode runs in the model's default state.

`_mx_lazy_sample` / `_mx_lazy_shape` are kept as device-sampler primitives (still exercised by the lazy-sampler-law test); the non-finite sentinel in `_sample_from_logits` is untouched.

The build-ahead overlap that preserves correctness *and* speed is a separate, larger change tracked in #501; this PR is the minimal correctness fix that can land on `main` on its own.

## Test

`tests/test_async_decode.py::test_ar_pipeline_matches_classic_token_for_token`: for one seed at temperature 1.0 / top-p 0.95 / top-k 20, the classic and `MTPLX_AR_PIPELINE=1` paths emit an identical token sequence. It fails on the current code (divergence at token 1) and passes with this change.


youssofal:
Thank you, this is in 2.12.0. Your commit is in main under your name, db6fe0fb. You found that the pipelined plain decoder drew its first token from a different random stream than every later token. Later in this release, the n-gram change left no model that uses the pipelined mode, so that path and `MTPLX_AR_PIPELINE` were removed (5a9fe002), and plain decoding always uses the standard loop. The commit was applied directly, so GitHub will show this PR as closed rather than merged.

One more thing about credit. Your earlier commits in main, from August and early September, were recorded under david@tea.xyz and david.tai@reg.finance. Neither address is linked to your GitHub account, so GitHub does not count those commits as yours. If you still have those addresses, adding them to your account under Settings, Emails makes GitHub credit them.

davidtai:
Lol thanks, I keep forgetting what accounts are on different machines.

Issue #390, opened 2026-08-29
Qwen3.8-Flash-Next: qwen4_exp lane requires ngram-manifest.json + mtplx_runtime.json that no public artifact provides


jbbottoms:
Hi — great work on the MoE lane. I'm running Qwen3.8-Flash-Next on an M5 Max 128GB and wanted to benchmark mtplx serving it, but the qwen4_exp runtime hard-requires `ngram-manifest.json` (`qwen4_runtime.py` loads it unconditionally) and `mtplx_runtime.json`, and neither ships in `Vontra/Qwen3.8-Flash-Next-MLX-oQ4-MTP@43a82b3`. `forge build` can't generate the manifest either (no `save_ngram_manifest` caller in the shipped package), and `forge discover` only lists 27B-class branded artifacts.

Could you publish those two metadata files (or the forge step that emits them)? Happy to run comparative numbers on M5 Max and share them back.

youssofal:
Fixed by the published packs. Both official Flash-Next artifacts on Hugging Face, Qwen3.8-Flash-Next-MTPLX-Bare-Speed and Qwen3.8-Flash-Next-MTPLX-Optimized-Speed, ship the ngram manifest and mtplx_runtime.json, resolve from the catalog by id, and are what the 2.10.0 release notes point at. Third-party conversions without the sidecar still classify honestly instead of loading broken.

PR #508, opened 2026-09-18
forge: convert Qwen3.8-Flash-Next (qwen4_exp) sources in tree (n-gram sidecar, streamed body, MTP head)

bpmforge:
## Summary

**Bottom line:** `mtplx forge build` could not convert any Qwen3.8-Flash-Next model. It handed the model to mlx-lm, which does not know this architecture, and stopped with `Model type qwen4_exp not supported` (#390). This PR adds a build path for that family, so anyone can turn a Flash-Next fine-tune into an MTPLX pack with one command.

What the new path does, in plain terms:

- **Memory table.** The 51B n-gram table is converted one shard at a time straight into the sidecar file the runtime reads. It never sits in RAM. About 3 minutes.
- **Main model.** Converted with MTPLX's own Flash-Next code, one tensor at a time. Converting a whole 5 GB shard at once trips the Mac GPU watchdog, so this avoids that.
- **Draft head.** Written in the exact layout the runtime loads, so speculative decoding works out of the box.
- **Build check.** A Flash-Next build no longer fails the "required depths D1, D2" check, because this family measures plain decode plus its own speculative default.

Everything after conversion (vision graft, verify, stamp) is the existing `forge build` flow. Nothing changes for other model families.

## Verification

Rebased onto v2.11.3 on 2026-09-18. On that base:

- 350 tests pass (the new lane tests plus `test_no_mlx_imports`, `test_public_cli`, `test_runtime_kpis`).
- `uv build --wheel` ships the new module; `scripts/fresh_venv_smoke.sh` passes against that wheel.
- Full end-to-end build on 2.11.2 (2026-09-16): `mtplx forge build --repo orcarouter/Qwen3.8-Flash-Next-Uncensored --recipe '{"body_bits":4,"body_group_size":32,"body_mode":"affine","qwen4_qsa_8bit":true}' --max` finished, verified and stamped. The pack has the same tensor count and sizes as the official Optimized Speed pack.

## Benchmark Evidence

| | |
|---|---|
| Mac | M5 Max, 128 GB, macOS 27.0 (26A428) |
| Runtime | MTPLX 2.11.2 (mlx 0.32.2, mlx-lm 0.31.3) |
| Model | orcarouter/Qwen3.8-Flash-Next-Uncensored (draft head identical to base) |
| Quantization | body 4-bit/g32 plus the Optimized Speed 8-bit set, QSA q/k/v/o 8-bit/g64, n-gram 4-bit/g32, draft experts 4-bit/g32 |
| Sampler | temperature 1.0, top-p 0.95, top-k 20 (the pack's default) |
| Tokens | 512 per response, 2 runs, served through `mtplx serve` |
| Fans | `--max` (diagnostic run, not a headline claim) |
| Date, commit | 2026-09-16, 9d30660 |

| Mode | tok/s |
|---|---|
| Plain decode (depth 0) | 44.7 |
| Speculative, depth 3 | 87.0 (1.95x) |

Build time about 11 minutes end to end from a Thunderbolt NVMe.


youssofal:
Thank you, this is in 2.12.0. Your commit is in main under your name, 3ce745f8, and Forge now converts Flash-Next source models directly. I added one follow-up that reads the converted files back through the loaders the runtime uses (2980f443). The commit was applied directly, so GitHub will show this PR as closed rather than merged.

Issue #301, opened 2026-08-20
Forge writes three drafter norm tensors shifted by exactly +1.0 when extracting an embedded head (draft acceptance collapses to 0-2%)


Blakeolson21:
## Summary

Running Forge against a checkpoint that embeds its MTP (multi-token
prediction) tensors in the main shards produces an `mtp.safetensors` sidecar
in which exactly three of the seven F32 RMSNorm gain tensors are wrong:

- `mtp.layers.0.self_attn.q_norm.weight` (256,)
- `mtp.layers.0.self_attn.k_norm.weight` (256,)
- `mtp.norm.weight` (5120,)

Each of the three comes out as the source value plus exactly 1.0, elementwise
and bit-exact in F32. The other 26 tensors in the sidecar, including the
remaining four F32 norm gains, are byte-identical to the values stored in the
source shards. The trunk loads correctly, so the drafter runs against a trunk
it no longer matches. Base decode is healthy (12.67 to 12.89 tok/s on this
machine) while draft acceptance collapses to 0 to 2 percent. That collapse
makes every draft depth slower than plain decoding.

The extraction is deterministic: two independent Forge runs on 2026-08-19
produced tensor-identical corrupted sidecars.

## Environment

- mtplx 2.8.1, installed binary at `~/.mtplx/bin/mtplx`. Per the changelog,
  2.8.2 and 2.8.3 (current main) change nothing in the forge extraction or
  sanitize path, so this applies to current main.
- macOS 26.3, Apple M3 Max, 64 GB
- Source model: `chimingw/Qwen3.8-27B-Uncensored-OrcaRouter-MLX-6bit`
  (Qwen3.8-27B family, 6-bit MLX quantization, group size 64). The
  checkpoint's index lists 29 `language_model.mtp.*` tensors, and its norm
  gains are stored in the absolute convention, the same one the official
  sidecar-layout builds use.

## Repro

1. Download `chimingw/Qwen3.8-27B-Uncensored-OrcaRouter-MLX-6bit` (the
   embedded-MTP layout: 29 `language_model.mtp.*` keys inside
   `model-0000x-of-00006.safetensors`).
2. Run `mtplx forge build` on it to produce an MTPLX artifact with an
   `mtp.safetensors` sidecar.
3. Compare the sidecar tensors against the same tensors read directly from the
   source shards (method below). Three norm gains differ by exactly +1.0; the
   other 26 tensors match byte for byte.
[...]

PR #511, opened 2026-09-18
qwen3.5 MTP: restore the +1.0 norm convention on sidecars forge did not build

stooit:
mlx-lm's `qwen3_5` sanitize shifts trunk RMSNorm gains by +1.0 when it detects a raw zero-centred checkpoint — but it drops every `mtp.*` key *before* that loop, so an MTP head never receives the shift.

Forge shifts at build time (#301), which covers every artifact it produces. A sidecar that arrives any other way — copied in by hand, or extracted from the base checkpoint a pack was derived from — reaches `_load_mtp_weights` still delta-encoded.

## Why this is worth catching

Nothing currently detects it:

- `inject_qwen3_5_mtp_support` logs `native head bound (depth 1, N tensors)`
- `validate_qwen3_5_mtp_support` passes
- decode runs at **~0% acceptance**, i.e. measurably **slower than plain AR**
- no error, no warning

The failure is an inversion rather than degradation. `pre_fc_norm_embedding` is uniformly negative in the delta convention, so applying it as `x * w` instead of `x * (1 + w)` flips the sign of the head's output. The correct token then lands near the *bottom* of the distribution: measured median rank 247,513 of 248,320, where random would be ~124,160.

## The change

Call `shift_delta_mtp_norms` in `_load_mtp_weights` as well as in forge.

The helper is detector-gated on the two separable norm families (q/k means, low-set means), so an absolute-convention sidecar passes through byte-identical and nothing is ever double-shifted. That idempotence is covered by a test.

## Verification

On a head extracted from a raw HF checkpoint and bound to a matching trunk:

| | held-out top-1 agreement | median rank of correct token |
|---|---|---|
| before | 0.0% | 247,513 |
| after | 75.2% | 0 |

Applying the same shift to an already-absolute head degrades it (79.6% → 58.6%), which is the falsification the detector gate exists to prevent — and it stays at 79.6% through this path, because the gate declines to shift it.

## Tests

`tests/test_qwen3_5_mtp_sidecar_norms.py`, 7 cases, no model loads: detector on both conventions, idempotence, loader restores delta, loader leaves absolute alone, non-norm tensors untouched, prefix stripping.

Existing suites pass unchanged: `test_qwen3_5_mtp_backend`, `test_qwen3_5_mtp_object_level`, `test_mtp_patch`, `test_mtp_payload_guards`, `test_artifacts`, `test_compressed_tensors` (158 tests).


youssofal:
Thank you, this is in 2.12.0. Your commit is in main under your name, 555a0692. A draft head added from a raw checkpoint now loads with the +1.0 norm convention, and a head that is already correct is left exactly as it is. The commit was applied directly, so GitHub will show this PR as closed rather than merged.

PR #514, opened 2026-09-20
Splash engine: serve the MTPLX API over Inco's Apple-silicon runtime

ksg98:
Adds [Inco's Splash](https://inco.ai/blog/splash/) as a second inference engine behind the existing API:

```bash
brew install incoai/tap/splash
mtplx serve --engine splash --model incoai/Qwen3.8-27B-Splash
```

Same endpoints, same dashboard, same native app — only the kernels change. `--engine` defaults to `mlx`, so nothing changes unless you ask for it.

Splash is built the opposite way round from a general engine: it supports a small set of models and rebuilds itself around each one, with fused Metal kernels compiled for that model's exact shapes and a DFlash 2 draft trained for it. It's faster on the two packages it supports, and it loads only those packages.

## Measured on a 64 GB M3 Max

macOS 26.6.2, Splash 1.0, `incoai/Qwen3.8-27B-Splash`, reasoning off, through the bridge over HTTP.

| | decode | TTFT | draft acceptance |
| --- | ---: | ---: | ---: |
| Short code prompt, 300 tok out | **66.5 tok/s** | 432 ms | 77% |
| Short prose prompt, 136 tok out | 30.7 tok/s | 252 ms | 28% |

Decode tracks draft acceptance closely, and acceptance depends on how predictable the text is. Code runs about twice as fast as prose on the same machine. Over a five-request mix: min 30.7, mean 45.4, p95 62.7 tok/s.

**Prefill and prefix reuse**, on a 25,812-token prompt:

| | TTFT | prefill |
| --- | ---: | ---: |
| Cold | 129.7 s | 199 tok/s |
| Same prompt again | **275 ms** | 25,792 of 25,812 tokens reused |

That's the part worth calling out: replaying a 25K-token context returns the first token in **275 ms instead of 129.7 s**, a ~470× improvement, because Splash reuses the prefix it already computed. It makes multi-turn work on a long document feel immediate after the first turn.

Memory: 21.1 GB active and peak, on a 64 GB machine.

For reference, Inco publishes 74 tok/s decode and 363 tok/s prefill at 32K for this model on a 48 GB M5 Pro. The M3 Max numbers above are lower, as expected on older silicon.

## How it fits together

```
mtplx serve --engine {mlx|splash}
   |                      |
   v                      v
MLX in-process      splash serve on a private loopback port
```

Splash has no runtime model-swap API, so loading a model is process lifecycle: the bridge supervises one engine process and proxies to it. Splash 1.0's CLI also hardcodes port 8000 behind an exclusive lock, so the bridge runs its inner server on a private port and keeps the public one for MTPLX.

`--engine splash` imports no MLX at all, so a Mac that only ever runs Splash doesn't need it installed.

## The dashboard

The bridge drives `mtplx.server.dashboard_state` — the MLX server's own module, which has no MLX dependency — rather than imitating its output. The Live tab therefore works unchanged: `progress` events move the decode gauge, `completed` carries each request's envelope, `new_max_tps` fires the record toast, and the min/max/mean/p95 row comes from the same `RollingMetrics`.

A finished request's speeds are the engine's own: Splash's `/status` counters are cumulative, so the bridge reads them either side of a request and reports the difference. Only the live gauge mid-stream is timed by the bridge, because Splash publishes counters rather than a live rate. Memory is `memory_actual`, the process footprint, not the KV pages alone.

## What Splash can't do, and how that's reported

| | MLX engine | Splash engine |
| --- | --- | --- |
| Models | many architectures | Splash packages only |
| Speculation | native MTP, tunable depth | DFlash 2, fixed |
| **KV cache** | 4-bit / 8-bit / unquantized | **8-bit, fixed** |
| Runtime settings | scheduler, batching, adaptive depth | none |

Splash's Metal kernels read 8-bit KV directly, so the width is a property of the compiled kernels, not a setting — there is no flag, env var or engine argument for it. The bridge publishes `kv_quant_policy: {supported: false, modes: ["q8"], disabled_reason: …}`, and both KV controls in the app (the Settings card and the top-bar inference-params panel) lock to q8 and explain why rather than offering a choice the engine can't honor. A test walks every view file and fails if a new surface offers a narrower width without consulting a policy.

MTP depth reads absent. Splash's draft acceptance is real and is reported per request, labelled `draft: dflash2` so it is never confused with MTPLX's MTP.

## In the app

- **Engine dropdown in the top bar**, before the model control, since the engine decides which models are loadable.
- **The model picker lists Splash packages** when Splash is selected, in the same rows as the MLX list. Splash keeps its own model setting, so switching engines never overwrites the other side's choice.
- **Settings → Engine** shows each package with Download / Update / Cancel and live installer progress, via a new `mtplx splash list|install|verify`. Downloads and manifest verification stay with Splash's own installer.

## Tests

44 Python and 23 Swift, all passing. `tests/test_public_cli.py` fails 21 tests both with and without this branch on my machine (missing `[server]` extras), so no regression there.

Two things worth noting about how they're written:

- The Swift decoding tests run against **real payloads captured from a live Splash daemon**, not hand-written approximations, so a missing non-optional field can't silently blank the dashboard.
- The Python contract tests **parse the app's own Swift DTOs** to derive the required fields, so the server and the app can't drift apart.

## Requirements

Splash needs an M3 or newer Mac on macOS 26.4+ with at least 36 GB of unified memory (48 GB recommended). Packages are 17.4 GB (27B) and 20.9 GB (35B-A3B), downloaded and verified on first use.

See [`docs/splash-engine.md`](docs/splash-engine.md).


youssofal:
Thanks for building this and for the numbers. A second engine behind the API is a large change, and it can't go into 2.12.0. I'll look at it after the release.

PR #518, opened 2026-09-21
forge(qwen4_exp): optional expert pruning from routing statistics (stacked on #508)

bpmforge:
## Summary

**Stacked on #508** (the qwen4_exp forge lane). Only the top two commits are new; this is a draft until #508 lands.

Adds an optional recipe key, `qwen4_expert_keep`, to the Flash-Next build lane. It takes a JSON keep-set produced from router statistics (`k`, `layers`, per-layer `keep`, optional `mtp_keep`) and prunes the resident expert bank at conversion time:

- Sets `text_config.num_experts = k` before the backend is constructed.
- Slices the sanitized `switch_mlp.{gate,up,down}_proj` stacks and the router rows to the kept experts.
- Leaves the MTP head at full width by default. `qwen4_prune_mtp_head: true` opts in to slicing it with `mtp_keep`.
- `attach_mtp` now reads the head's expert count off its own router rows in `mtp.safetensors` (`mtp_head_num_experts`), so a pack may carry a 512-expert head on a 256-expert trunk.

Active compute (top-k) is unchanged. Only the resident expert bank shrinks. `k` is uniform per layer because the runtime config carries one `num_experts`. Without the key the lane behaves exactly as in #508.

## Why the head stays full width

Measured on orcarouter/Qwen3.8-Flash-Next-Uncensored, body 4-bit/g32 with 3-bit/g32 experts, K=256 of 512, M5 Max 128 GB, greedy, 512 tokens, 3 prompts x 2:

| pack | body on disk | AR tok/s | depth-3 tok/s | gain |
|---|---|---|---|---|
| full, 512 experts | 60 GiB | 50.9 | 87.6 | 1.72x |
| K=256, head pruned by proxy | 33 GiB | 48.9 | 71.7 | 1.47x |
| K=256, head full width | 33 GiB | 49.0 | 75.9 | 1.55x |

Held-out code perplexity: 1.9226 (K=256) vs 1.9318 (full), same 200K-token slice. Pruning the head from trunk routing statistics costs draft acceptance for ~0.7 GiB saved, so the default keeps it. The remaining gap is the pruned trunk's distribution drifting from what the head was trained against.

## Verification

- `tests/test_forge_qwen4_expert_prune.py` (11 tests): keep-set parsing and validation, stack and router slicing, head-width inference from bf16 and packed u32 routers.
- Lane, forge CLI, public CLI and `test_no_mlx_imports` pass on the v2.11.3 base. Full suite: 6968 passed.
- End to end: `mtplx forge build` with the keep-set produced a 65 GiB pack (9 body shards + 1.58 GiB head + 29.8 GiB table), verified and stamped; `mtplx serve` loads it and reports the rows above.


bpmforge:
Closing: the expert-pruning experiment this lane enabled is a dead end. K=256 packs fit 64 GB and keep perplexity flat, but task accuracy collapses (HumanEval 61.6 vs 95.1, GSM8K 6.5 vs 95.0, MMLU-STEM 33.5 vs 82.5). The dense 27B is the better 64 GB choice. The head-width runtime change is available on the branch if anyone needs a trunk and head with different expert counts.

youssofal:
Thanks for coming back with the task accuracy numbers. Perplexity staying flat while HumanEval and GSM8K fell sharply is a useful result for anyone considering expert pruning.

